PR #167, opened 2012-08-22
Change identification and resolution of PA Films.

fredvdd:
Create canonical URIs for Films based on their prodId rather than RT
film number. Resolving existing content uses both URI styles to avoid
creating duplicate content. Aliases of the alternate style is added to
the content.


src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -320,22 +323,42 @@ private String getClosedPostfix(Channel channel) {
...
+                film = new Film();
+                Item.copyTo((Episode) previous, film);
+            }
+            film.addAlias(rtFilmAlias);
         } else {
             film = getBasicFilm(progData);
         }
+
+        String certificateValue = progData.getCertificate();
+        if (!Strings.isNullOrEmpty(certificateValue)) {
```
tfm: Dias, please check why this extra functionality for certs is in this pull request; was it just moving some existing code around?

PR #613, opened 2015-01-01
refactoring one ListenableFuture instance to Java 8 CompletableFuture

yulin2:
Hi, I'm doing research on new concurrent constructs in Java 8. I found CompletableFuture in Java 8 has the same functionality as Guava ListenableFuture. But CompletableFuture is much nicer because it comes together with Lambda expression in Java 8, and it is monadic, which makes the code more readable and cleaner. Also, it provides more ways to compose different tasks. Therefore, using CompletableFuture instead of ListenableFuture is better for future extension and maintenance of the code.

I tried to refactoring one ListenableFuture instance to Java 8 CompletableFuture in `atlas`, so you can see the difference. You don't have to merge this pull request. I'm just wondering your opinion on this kind of refactoring (or migrating the code from Java 6, 7 to Java 8). Do you think the refactoring is useful? Do you have any plan to use Java 8? Thanks.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bbc/atoz/BbcSlashProgrammesAtoZUpdater.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/atoz/BbcSlashProgrammesAtoZUpdater.java
@@ -5,6 +5,7 @@ import static com.metabroadcast.common.scheduling.UpdateProgress.SUCCESS;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


Dalamar42:
Sorry it took us so long to reply. Unfortunately your PR slipped through the cracks and I only noticed it now that we are cleaning up old PRs.

That's an interesting idea on the basis that it allows us to shorten the callback to a Lambda function rather than having to use the Java anonymous class syntax which is quite clunky. That said with this approach we do make the code less explicit by mixing the success and error handling paths and null checking the exception to detect a failure is a bit clunky in its own right. I don't think making the code less descriptive is worth the benefit of using Lambdas.

You did give us an interesting idea though. We could make something along the lines of `MoreFutures.getCallback(o -> success(o), e -> failure(e))` that returns a FutureCallback.
That could give us brevity without losing self documentation.

PR #614, opened 2015-01-06
Filter broadcasts without crids from controller output

tfm:
These would never be uploaded to YouView anyway, and break
since later code relies on them being present

### DIFF
--- a/src/main/java/org/atlasapi/query/v2/ContentFeedController.java
+++ b/src/main/java/org/atlasapi/query/v2/ContentFeedController.java
@@ -18,6 +18,7 @@ import org.atlasapi.feeds.youview.hierarchy.ContentHierarchyExpander;
 import org.atlasapi.feeds.youview.hierarchy.ItemAndVersion;
 import org.atlasapi.feeds.youview.hierarchy.ItemBroadcastHierarchy;
 import org.atlasapi.feeds.youview.hierarchy.ItemOnDemandHierarchy;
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/query/v2/ContentFeedController.java
```diff
@@ -119,8 +123,10 @@ public void generateFeed(HttpServletRequest request, HttpServletResponse respons
             if (content.get() instanceof Item) {
                 Item item = (Item) content.get();
                 
+                Predicate<ItemBroadcastHierarchy> broadcastFilter = FilterFactory.broadcastFilter(Optional.<DateTime>absent());
```
oli-hall: you might want to check what that does when the DateTime not present...

oli-hall:
back to you, @tfm

src/main/java/org/atlasapi/query/v2/ContentFeedController.java
tfm: I did, I think it does the right thing. Are you saying you think it doesn't?

oli-hall: If no DateTime, it returns ``Predicates.alwaysTrue()``. In that case, there's no point in having it, as it gives nothing extra. Afaik, the only point in running the extra filtering is to get the Pid filtering, which this won't do.

tfm: Fixed the predicate :)

PR #617, opened 2015-01-12
Increase full fetch window for Nitro TV content to 7 days ahead

oli-hall:


src/main/java/org/atlasapi/remotesite/bbc/nitro/LocalOrRemoteNitroFetcher.java
```diff
@@ -63,7 +63,7 @@ public boolean apply(Item input) {
                         final Interval fetchForBroadcastsWithin = 
                                 MediaType.AUDIO.equals(input.getMediaType()) 
                                     ? broadcastInterval(today.minusDays(5), today.plusDays(1)) 
-                                    : broadcastInterval(today.minusDays(3), today.plusDays(3));
+                                    : broadcastInterval(today.minusDays(3), today.plusDays(7));
```
tfm: Would be more inclined to put in an explicit `or broadcast is in the future` condition. Then the intention is really clear.

oli-hall: Not sure what you mean... wrap the entire predicate? surely we only care about TV content in the future?

tfm: something like 

```
public boolean apply(org.atlasapi.media.entity.Broadcast input) {
                                return 
input.getTransmissionTime().isAfterNow() ||
fetchForBroadcastsWithin.contains(input.getTransmissionTime());
                            }});
```

oli-hall: OK, will go with that. I'm sensing I've not quite got the full context around this, but will go with your suggestion for now.

PR #621, opened 2015-01-13
Opta events refactor

oli-hall:
- combine data-models for soccer/non-soccer to reduce code duplication
 - remove duplicated data-handlers
 - refactor module to simplify code paths and reduce duplication


src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
```diff
@@ -57,26 +54,47 @@
     
     @PostConstruct
     public void startBackgroundTasks() {
-        scheduler.schedule(soccerIngestTask().withName("Opta Soccer Events Updater"), RepetitionRules.NEVER);
-        scheduler.schedule(sportsIngestTask().withName("Opta Sports Events Updater"), RepetitionRules.NEVER);
+        scheduler.schedule(ingestTask(soccerFetcher()).withName("Opta Football Events Updater"), RepetitionRules.NEVER);
```
tsprlng: would suggest Opta Events (Football) and Opta Events (Non-football) to make the class easier to find

src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
```diff
@@ -57,26 +54,47 @@
...
+    private OptaEventsFetcher<SportsTeam, SportsMatchData> httpEventsFetcher(String sportPrefix, OptaDataTransformer<SportsTeam, SportsMatchData> dataTransformer) {
+        return new HttpOptaEventsFetcher<>(sportConfig(sportPrefix), HttpClients.webserviceClient(), dataTransformer, new UsernameAndPassword(username, password), baseUrl);
+    }
+
+    private OptaSportsDataHandler dataHandler() {
+        return new OptaSportsDataHandler(organisationStore, eventStore, utility());
+    }
+
+    /**
+     * Opta Sports are configured through three parameters: feed type, a competition id, and a season. Each sport is
```
tsprlng: nice

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsMatchInfo.java
```diff
@@ -28,17 +30,27 @@ public MatchInfoAttributes attributes() {
...
         
+        @SerializedName("GroupName")
+        private String groupName;
         @SerializedName("MatchDay")
         private String matchDay;
         @SerializedName("MatchType")
         private String matchType;
         @SerializedName("Period")
         private String period;
+        @SerializedName("RoundNumber")
```
tsprlng: maybe comment which are football-specific?

src/main/java/org/atlasapi/remotesite/opta/events/sports/OptaSportsDataHandler.java
```diff
@@ -117,7 +113,7 @@ private String parseTeamIdFrom(String uId) {
     }
     
     private Optional<DateTime> parseStartTime(SportsMatchData match, OptaSportType sport) {
-        String dateStr = match.matchInformation().date();
+        String dateStr = match.matchInformation().date().date();
```
tsprlng: hmm

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsTeamData.java
```diff
@@ -16,6 +32,8 @@ public TeamDataAttributes attributes() {
     
     public static class TeamDataAttributes {
         
+        @SerializedName("HalfScore")
```
tsprlng: scoreAfterFirstHalf? theirHalfOfTheScore?

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/InnerTimingType.java
```diff
@@ -1,4 +1,4 @@
-package org.atlasapi.remotesite.opta.events.soccer.model;
+package org.atlasapi.remotesite.opta.events.sports.model;
 
 import com.google.gson.annotations.SerializedName;
```
tsprlng: This timing stuff is definitely the kind of classes where a 'wtf is this' sort of class-level comment would be really good.

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsMatchInfo.java
```diff
@@ -51,8 +63,29 @@ public String period() {
...
+    
+    public static class MatchDate {
+        
+        private String date;
+        
+        public MatchDate(String date) { 
+            this.date = checkNotNull(date);
+        }
+        
+        public String date() {
```
tsprlng: 'dateString' maybe? / why is this wrapped at all? just their format?

src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
oli-hall: gp, updated

src/main/java/org/atlasapi/remotesite/opta/events/sports/OptaSportsDataHandler.java
oli-hall: Ah yes, the date date. Welcome to the bizarre and wonderful world of Opta JSON event models... where everything is not what it seems (except when it is)

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/InnerTimingType.java
oli-hall: Hmm, not sure about this. This is basically just a representation of the Opta soccer feed, verbatim. There's no particular need for these classes, other than to stop default gson deserialization complaining that the model is missing classes.

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsMatchInfo.java
oli-hall: i'm not sure it's that simple. It's a combination of the two, and i think some *may* be football specific, and i think some are merely omitted from feeds when they don't have enough information.
tl;dr not 100% certain which are football-specific

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsMatchInfo.java
oli-hall: fun with wierd date formats, which change depending upon the phase of the moon when the match is broadcast, or some such metric. I'll add a comment in to explain it though, 'tis rather wierd.

src/main/java/org/atlasapi/remotesite/opta/events/sports/model/SportsTeamData.java
oli-hall: Are you bitching about Opta's field naming?

tsprlng: we have control over the names in our objects. the serializedname can remain unhelpful.

oli-hall: That does however assume we know that what the intended meaning of this field is, which we don't. Either of your proposals assume further knowledge about the context, which we lack currently. I agree that where we know the usage and the field is ambiguously named, then yes, we should change things on our side to be more user-friendly, but I don't think that that's the case here.

PR #622, opened 2015-01-13
Computing media type for nitro brands and series

maxtomassi:


src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroBrandExtractor.java
```diff
@@ -49,4 +53,21 @@ protected MasterBrand extractMasterBrand(Brand source) {
         return source.getMasterBrand();
     }
 
+    @Override
+    protected void extractAdditionalFields(Brand source, org.atlasapi.media.entity.Brand content, DateTime now) {
+        content.setMediaType(computeMediaType(source));
+    }
+
+    private MediaType computeMediaType(Brand source) {
```
oli-hall: worth moving this up to the ``NitroContentExtractor`` and then overriding the method for the special case of episode/clip/film? (might also be worth checking what happens for non-film items, as they may not be covered by the episode case...

oli-hall:
:+1: apart from that one comment

PR #623, opened 2015-01-16
Treat a 404 as no content

tfm:
The BT API cannot return an emptyset for a valid,
but empty, group. Therefore we will catch that case
and interpret any 404 as an empty group.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/btvod/portal/XmlPortalClient.java
+++ b/src/main/java/org/atlasapi/remotesite/btvod/portal/XmlPortalClient.java
@@ -15,11 +15,14 @@ import nu.xom.Elements;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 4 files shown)


src/main/java/org/atlasapi/remotesite/btvod/portal/XmlPortalClient.java
```diff
@@ -39,13 +42,18 @@ public XmlPortalClient(String uriBase, SimpleHttpClient httpClient) {
...
+    public Optional<Set<String>> getProductIdsForGroup(String groupId) {
         int pageNumber = 1;
         PageData pageData = null;
         ImmutableSet.Builder<String> ids = ImmutableSet.builder();
         do {
             try {
                 pageData = getProductIdsForGroupPage(groupId, pageNumber);
+            } catch (HttpStatusCodeException e) {
+                if (HttpStatusCode.NOT_FOUND.is(e.getStatusCode())) {
+                    return Optional.absent();
```
oli-hall: I would argue the comment about returning ``Optional`` in case of 404 belongs in the implementation - it's an implementation-specific detail, no?

oli-hall:
:+1:

PR #624, opened 2015-01-16
Ensure that merging Item with Episode and vice versa converts existing type correctly

oli-hall:


src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -64,16 +64,21 @@ public Item merge(Item current, Item extracted) {
...
+                Item newItem = new Item();
+                Item.copyTo(current, newItem);
+                current = newItem;
+            }
+        } else if (current instanceof Item) {
+            if (extracted instanceof Episode) {
+                Episode newEp = new Episode();
+                Item.copyTo(current, newEp);
+                
+                current = mergeEpisodeSpecificFields(newEp, (Episode) extracted);
```
tfm: Aren't we missing the line that sets `current = newEp`

tfm:
:+1:

PR #626, opened 2015-01-19
Store youview items by pcrid, where available

tfm:
This reduces, greatly, the number of items where they're
repeated on regional variants.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/youview/YouViewContentExtractor.java
+++ b/src/main/java/org/atlasapi/remotesite/youview/YouViewContentExtractor.java
@@ -18,7 +18,6 @@ import org.atlasapi.media.entity.Policy.Platform;
 import org.atlasapi.media.entity.Publisher;
 import org.atlasapi.media.entity.Version;
 import org.atlasapi.remotesite.AttributeNotFoundException;
-import org.atlasapi.remotesite.ContentExtractor;
 import org.atlasapi.remotesite.netflix.ElementNotFoundException;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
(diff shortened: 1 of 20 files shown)


src/main/java/org/atlasapi/remotesite/youview/YouViewContentExtractor.java
```diff
@@ -97,6 +96,31 @@ public Item extract(Publisher targetPublisher, Element source) {
...
+            DateTime transmissionTime = getTransmissionTime(source);
+            
+            return String.format("http://%s/programmecrid/%d%02d%02d/%s", 
+                    publisher.key(), 
+                    transmissionTime.getYear(),
+                    transmissionTime.getMonthOfYear(),
+                    transmissionTime.getDayOfMonth(),
+                    programmeCrid.get());
+        } else {
+            return String.format("http://%s/scheduleevent/%s", publisher.key(), getId(source));
```
oli-hall: erm, this duplicates the existing ``scheduleEventUriFor`` method, and is less clear than using the method directly on this line. Also, superfluous ``else``

src/test/java/org/atlasapi/remotesite/ContentMergerTest.java
```diff
@@ -176,6 +183,75 @@ public void testEpisodeEpisodeMergingProducesEpisode() {
...
+        extracted.setAliases(ImmutableSet.of(new Alias("3", "4")));
+        
+        current.setAliasUrls(ImmutableSet.of("http://a.com/b", "http://b.com/c"));
+        extracted.setAliasUrls(ImmutableSet.of("http://c.com/d"));
+        
+        Item merged = contentMerger.merge(current, extracted);
+        
+        assertEquals(2, merged.getAliases().size());
+        assertEquals(2, merged.getAliasUrls().size());
+        assertEquals("1", Iterables.getFirst(merged.getAliases(), null).getNamespace());
```
oli-hall: This seems potentially flakey... it's a set, ordering is not guaranteed. You could set both the current aliases to have the same NS

oli-hall:
back to you @tfm

oli-hall:
:+1:

PR #627, opened 2015-01-20
Actually use the returned value from the ContentMerger

oli-hall:
(rather than assuming that the existing Content is being modified (which isn't always true).)

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/LocalOrRemoteNitroFetcher.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/LocalOrRemoteNitroFetcher.java
@@ -6,6 +6,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.atlasapi.media.entity.Brand;
+import org.atlasapi.media.entity.Container;
 import org.atlasapi.media.entity.Content;
 import org.atlasapi.media.entity.Episode;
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/remotesite/bbc/nitro/LocalOrRemoteNitroFetcher.java
```diff
@@ -146,17 +146,17 @@ public LocalOrRemoteNitroFetcher(ContentResolver resolver, NitroContentAdapter c
             T fetched = fetchedIndex.remove(existing.getCanonicalUri());
             if (fetched != null) {
                 if (fetched instanceof Brand) {
-                    contentMerger.merge((Brand) existing, (Brand) fetched);
+                    resolved.add((T) contentMerger.merge((Brand) existing, (Brand) fetched));
```
oli-hall: Unforunately, because we've gone from values of type ``T`` to return types of ``Container`` or ``Item``, we have to do an unchecked cast to ``T``, which isn't ideal. Any ideas how to fix this?

oli-hall: I'm going to have to change this method i think, this ``T`` typing is not helpful, and prevents this method working properly.

maxtomassi:
:+1:

PR #628, opened 2015-01-20
Do not ingest actual availability start if revoked

tfm:


src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroAvailabilityExtractor.java
```diff
@@ -200,6 +201,11 @@ private boolean shouldIngestActualAvailabilityStart(Availability source, String
             return true;
         }
         
+        if (REVOKED.equals(source.getStatus())) {
```
oli-hall: ``source.getStatus()`` or ``source.getRevocationStatus()``?

oli-hall:
back to you, @tfm

PR #630, opened 2015-01-22
Orphan old, and no longer needed, schedule events

tfm:


src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalanceBreakerController.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+    
+    @RequestMapping(value="/system/youview/orphan/{date}", method=RequestMethod.POST)
+    public void breakEquivalence(HttpServletResponse response, 
+            @PathVariable("date") String date) throws IOException {
+        
+        LocalDate day = null;
+        try {
+            day = dateFormat.parseLocalDate(date);
+        } catch (IllegalArgumentException iae) {
```
oli-hall: That's a bad request rather than a 404 to my thinking. If date is outside a given range, then yes, 404, but invalid date == 400

src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalenceBreaker.java
```diff
@@ -0,0 +1,121 @@
...
+        this.lookupEntryStore = checkNotNull(lookupEntryStore);
+        this.contentResolver = checkNotNull(contentResolver);
+        this.lookupWriter = checkNotNull(lookupWriter);
+        this.referenceSchedulePublisher = checkNotNull(referenceSchedulePublisher);
+        this.publishersToOrphan = ImmutableSet.copyOf(publishersToOrphan);
+    }
+    
+    @Override
+    protected void runTask() {
+        DateTime from = DateTime.now();
```
oli-hall: would a ``Clock`` be useful for simulating the date ranges in testing?

src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalenceBreaker.java
```diff
@@ -0,0 +1,121 @@
...
+                }
+            }
+        }
+        
+    }
+
+    private void orphan(Item equiv) {
+        lookupWriter.writeLookup(ContentRef.valueOf(equiv), 
+                Iterables.transform(ImmutableSet.of(equiv), ContentRef.FROM_CONTENT), Publisher.all());
+
```
oli-hall: erroneous blank line

src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalenceBreaker.java
```diff
@@ -0,0 +1,121 @@
...
+                    } else {
+                        log.trace("Not orphaning item {}", equiv.getCanonicalUri());
+                    }
+                }
+            }
+        }
+        
+    }
+
+    private void orphan(Item equiv) {
```
oli-hall: a comment around how things are being orphaned would be good, it's non-obvious to someone not extremely well-versed in how lookup works as to what we're doing here

src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalenceBreaker.java
```diff
@@ -0,0 +1,121 @@
...
+        lookupWriter.writeLookup(ContentRef.valueOf(equiv), 
+                Iterables.transform(ImmutableSet.of(equiv), ContentRef.FROM_CONTENT), Publisher.all());
+        
+    }
+
+    private boolean shouldOrphan(Identified identified) {
```
oli-hall: again, javadoc on the overall orphaning strategy would be good here

tfm: See the class-level javadoc; thought that was the right place.

oli-hall: Ah ok, missed that. 'tis clearer now

oli-hall:
:+1:

src/main/java/org/atlasapi/remotesite/youview/YouViewEquivalanceBreakerController.java
tfm: except that it's part of the resource name, and if it's not parsable as a date, then that resource is not found.

PR #632, opened 2015-01-22
Create Publisher rather that use the field from result which is not yet ...

wwajerowicz:
At the moment when result.getPublisher() is called, publisher has not been yet set.

This is a fix which doens't solve the underlying problem, which is the more specific fields should be written after more general fields and possibly override them, rather than the other way round as it is now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -543,7 +543,7 @@
        <dependency>
           <groupId>com.metabroadcast.atlas.glycerin</groupId>
           <artifactId>glycerin</artifactId>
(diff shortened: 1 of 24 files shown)


tfm:
Having reviewed the code, given the few number of subclasses, I think the fix for the underlying problem is safe to do. We should separate out into 2 abstract methods in `ContentModelTransformer` the creation of the object, vs setting the fields. Therefore in the superclass we can do

```
   create();
   setContentFields();
   setTypeSpecificFields();
```

PR #634, opened 2015-01-23
Use LookupEntryStore directly to orphan items

tfm:
This should see a much better throughput, as we'll reduce the number of writes greatly. We're now
writing one entry per item in the equivalent set, rather than detaching each separately, which each would cause one write per item in the set.


oli-hall:
:+1:

PR #638, opened 2015-01-27
Improved content creation for Nitro

maxtomassi:


tfm:
So the functional change here is change the order of detection of an `Episode` to take priority over detection of a `Film`?

maxtomassi:
@tfm the method extractAdditionalItemFields in the same class assumes that if the field episodeOf is populated then the item is an episode and it does the explicit cast. That was the real problem (ClassCastException). With this fix we ensure that when the field episodeOf is populated, an Episode is created: this should make sense in functional terms too. What's your view?

tfm:
It sounds like we were previously mis-identifying some things as films; is that the case? So in some instances, according to the old logic `item instanceof film && item.episodeOf() != null`. Is that right?

PR #639, opened 2015-01-29
Mbst 9956 tasks refactor

oli-hall:
This is the long awaited Tasks refactor. This pull request/branch SHOULD NOT BE MERGED into either stage or master until a good way to handle the inevitable errors and bugs with YouView has been agreed upon. While I'd hope my code works, I would be surprised if it works perfectly first time.


src/main/java/org/atlasapi/query/v2/ContentFeedController.java
```diff
@@ -122,8 +138,8 @@ public void generateFeed(HttpServletRequest request, HttpServletResponse respons
             }
             if (content.get() instanceof Item) {
                 Item item = (Item) content.get();
-                
-                Predicate<ItemBroadcastHierarchy> broadcastFilter = FilterFactory.broadcastFilter(Optional.<DateTime>absent());
+                // TODO what is the default here?
```
tfm: Still a `TODO`?

PR #647, opened 2015-03-20
Make Opta ID parsing sport-specific

tfm:
The original Opta feed that was used was had IDs that
differ from the new one we're using. We therefore had
had to remove the prefix 't'. However, it transpired
that we should only do this removal for rugby, since
for other sports the IDs were consistent between the
two feeds.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
+++ b/src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
@@ -54,43 +54,57 @@ public class OptaEventsModule {
     
     @PostConstruct
     public void startBackgroundTasks() {
-        scheduler.schedule(ingestTask(soccerFetcher()).withName("Opta Events (Football) Updater"), RepetitionRules.NEVER);
-        scheduler.schedule(ingestTask(sportsFetcher()).withName("Opta Events (Non Football) Updater"), RepetitionRules.NEVER);
-    }
-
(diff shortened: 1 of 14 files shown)


src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
```diff
@@ -120,15 +134,21 @@ private OptaSportsDataHandler dataHandler() {
...
-                .withSeasonId(Iterables.get(configItems, 2))
-                .build();
+                .withSeasonId(Iterables.get(configItems, 2));
+        
+        if (Iterables.size(configItems) == 4) {
+            config.withPrefixToStripFromId(Iterables.get(configItems, 3));
+        }
+        
+        return config.build();
     }
```
wwajerowicz: Wouldn't it be better to have separate properties for each config value? e.g. [prefix].[sportName].feedType, [prefix].[sportName].competitionId etc?

wwajerowicz:
Generally looks good.

I guess some of the config from OptaEventsUtility could be put somewhere in a config/db, but I guess that's for another time.

src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java
tfm: Yeah, agree. Also agree with the other comment that this should ultimately live in the db. Wasn't going to change that for the one extra optional parameter, though :)

PR #654, opened 2015-04-08
Content writers return written item

mbst-jamie:


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -116,7 +120,8 @@ private Void deserializeAndUpdateContent(HttpServletRequest req, HttpServletResp
             log.error("Error reading input for request " + req.getRequestURL(), e);
             return error(resp, HttpStatusCode.SERVER_ERROR.code());
         }
-        
+
+        resp.setHeader("X-Content-ID", codec.encode(BigInteger.valueOf(content.getId())));
```
tfm: Return a `Location` header, not some custom thing.

PR #656, opened 2015-04-16
Knowledgemotion add getty publisher

williamberg:


src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionFileProcessor.java
```diff
@@ -26,8 +26,9 @@
         KnowledgeMotionSourceConfig.from("BBC Worldwide", Publisher.KM_BBC_WORLDWIDE, "km-bbcWorldwide:%s", "http://bbc.knowledgemotion.com/%s"),
         KnowledgeMotionSourceConfig.from("British Movietone", Publisher.KM_MOVIETONE, "km-movietone:%s", "http://movietone.knowledgemotion.com/%s"),
         KnowledgeMotionSourceConfig.from("Bloomberg", Publisher.KM_BLOOMBERG, "bloomberg:%s", "http://bloomberg.com/%s"),
-        KnowledgeMotionSourceConfig.from("AP", Publisher.KM_AP, "ap:%s", "http://ap.knowledgemotion.com/%s")
-    );
+        KnowledgeMotionSourceConfig.from("AP", Publisher.KM_AP, "ap:%s", "http://ap.knowledgemotion.com/%s"),
+        KnowledgeMotionSourceConfig.from("Getty", Publisher.GETTY, "getty:%s", "http://gettyimages.co.uk/%s")
```
tsprlng: Basically you might want to prefix this like KM_GETTY and make up a URL pattern similar to the others – to reflect that this isn't actually Getty's full, own set of data, it's a subset that's specific to KM.

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionSpreadsheetColumn.java
```diff
@@ -16,7 +16,7 @@
     DATE("Date"),
     DURATION("Duration"),
     KEYWORDS("Keywords"),
-    PRICE_CATEGORY("Price category (1= stock, 2=news, 3=brand)"),
```
tsprlng: they actually named a column this way?!

tsprlng:
:+1: but with comments

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionFileProcessor.java
```diff
@@ -26,8 +26,9 @@
...
+        );
```
wwajerowicz: @williamberg  Shouldn't the Getty publisher be named KM_GETTY to be consistent with names of other KnowledgeMotion publisher?

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionSpreadsheetColumn.java
williamberg: lol bbq

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionFileProcessor.java
williamberg: We do actually want to add this to the existing publisher, that's been pretty well discussed.

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionFileProcessor.java
williamberg: https://github.com/atlasapi/atlas/pull/656#discussion_r28515041

src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionFileProcessor.java
tsprlng: There is no existing publisher.

williamberg: Ah. Then I agree with you.

PR #660, opened 2015-04-30
Take first spotlight result if there are many.

tfm:


tsprlng:
:+1:, but there really ought not to be multiple results anyway

PR #661, opened 2015-05-06
Bt vod mpx

wwajerowicz:


src/main/java/org/atlasapi/remotesite/btvod/BtVodMpxImageExtractor.java
```diff
@@ -0,0 +1,56 @@
+package org.atlasapi.remotesite.btvod;
```
mbst-jamie: I got to here

pom.xml
```diff
@@ -335,6 +335,11 @@
     <dependencies>
         <dependency>
             <groupId>org.atlasapi</groupId>
+            <artifactId>atlas-model</artifactId>
```
tfm: My earlier comment about this has been lost. Why have we added this dependency?

src/main/java/org/atlasapi/remotesite/btvod/model/BtVodPlproduct$pricingTier.java
```diff
@@ -0,0 +1,35 @@
+package org.atlasapi.remotesite.btvod.model;
+
+public class BtVodPlproduct$pricingTier {
+
+    private Long plproduct$absoluteStart;
+    private Long plproduct$absoluteEnd;
+    private BtVodPlproduct$amounts plproduct$amounts;
+
+    public BtVodPlproduct$pricingTier() {
```
tfm: We were going to rename these hideous fields/classes with $ in them

src/main/resources/config/environments/stage.properties
```diff
@@ -59,6 +59,7 @@ messaging.system=AtlasOwlStage
 
 
 bt.events.s3.bucket=stage-events-data
+bt.vod.mpx.feed.baseUrl=http://feed.product.theplatform.eu/f/CiIRPC/btv-prd-search
```
tfm: Remove this, it should be in puppet

src/main/resources/config/environments/prod.properties
```diff
@@ -58,6 +58,7 @@ messaging.system=AtlasOwlProd
 
 
 bt.events.s3.bucket=prod-events-data
+bt.vod.mpx.feed.baseUrl=http://feed.product.theplatform.eu/f/CiIRPC/btv-prd-search
```
tfm: Remove this, it should be in puppet

tfm:
@wwajerowicz back to you

pom.xml
wwajerowicz: I think I had some problems updating it getting updated from my local maven. I think having an explicit dependency also makes it a bit clearer that it acutally depends on model.

PR #662, opened 2015-05-22
Maintain schedules for POSTed items

tfm:
This will also ensure that schedule update events are created,
which means that deer schedule stores will be updated

### DIFF
--- a/src/main/java/org/atlasapi/query/v2/ContentWriteController.java
+++ b/src/main/java/org/atlasapi/query/v2/ContentWriteController.java
@@ -1,5 +1,7 @@
 package org.atlasapi.query.v2;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -133,6 +150,18 @@ private Void deserializeAndUpdateContent(HttpServletRequest req, HttpServletResp
...
+            Maybe<Channel> channel = channelResolver.fromUri(broadcast.getBroadcastOn());
+            if (channel.hasValue()) {
+                scheduleWriter.replaceScheduleBlock(item.getPublisher(), 
+                                                    channel.requireValue(), 
+                                                    ImmutableSet.of(new ItemRefAndBroadcast(item, broadcast)));
+            }
+        }        
+    }
+
     private Content merge(Maybe<Identified> possibleExisting, Content update, boolean merge) {
```
wwajerowicz: Couldn't this logic be moved to ContentWriter? This way we would ensure that the schedules get updated the same way whether the update is coming through POST or directly from Java API.

tfm: Quite deliberately limiting the scope to content POSTs, else we'd duplicate updates where content is updated through a schedule updater which writes content and then writes a schedule.

wwajerowicz:
:+1:

PR #663, opened 2015-05-27
Switch to new RTE search feed

tfm:
This feed has complete VOD listings, which lets us
remove related links for content not in the feed.


src/main/java/org/atlasapi/remotesite/rte/RteHttpFeedSupplier.java
```diff
@@ -2,30 +2,62 @@
...
+    public Document get() {
         try {
-            return atomClient.get(feedUrl);
+            return httpClient.get(new SimpleHttpRequest<Document>(feedUrl, TRANSFORMER));
         } catch (Exception e) {
             throw Throwables.propagate(e);
         }
     }
 
+    private SimpleHttpClient buildFetcher() {
```
mbst-jamie: UGH

mbst-jamie:
:+1:

src/main/java/org/atlasapi/remotesite/rte/RteHttpFeedSupplier.java
williamberg: I assume the HTTP client for this feed needs to be configured specially. Couldn't it be constructed and supplied from the module, though?

PR #665, opened 2015-05-29
Deactive old content in BT VoD catalogue

tfm:


src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
```diff
@@ -102,6 +112,7 @@ public void runTask() {
                     itemExtractor.getResult().getFailures()));
             
             contentGroupUpdater.finish();
+            oldContentDeactivator.finish();
```
wwajerowicz: I think it would be better to add `.finish()` method to MultiplexingVodContentListener which would call `.finish()` on all it's delegates. Otherwise, it will be easy to forget to do that if you add a new `Listener`

tfm: Yeh, good point. I'm going to come up with a better name for the content listener, then will make this change. It's no longer really a content listener if we add a `finish` method

PR #674, opened 2015-06-16
Batch and parallelise requestes to Nitro

tfm:
Instead of requesting all programmes/availabilities/versions
for a schedule day in one go, batch them to improve throughput


mbst-jamie:
:+1:

PR #680, opened 2015-07-02
Ingest images from MPX

tfm:
- Add new source, to ingest low-res images into for testing
  without affecting users of the existing source
- Derive brand image from the image of the first episode
  in the feed, according to editorial ordering


mbst-jamie:
:+1:

PR #682, opened 2015-07-07
Create tag for content in the "new" feed

tfm:


src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
```diff
@@ -57,16 +70,22 @@ public BtVodUpdater(ContentResolver resolver, ContentWriter contentWriter,
...
-        oldContentDeactivator.start();
+        
+        newFeedContentMatchingPredicate.init();
+        
+        BtVodDescribedFieldsExtractor describedFieldsExtractor = new BtVodDescribedFieldsExtractor(topicResolver, topicWriter, publisher, newFeedContentMatchingPredicate, newTopic);
+        
+        staleTopicContentRemover.beforeContent();
+        contentGroupUpdater.beforeContent();
+        oldContentDeactivator.beforeContent();
         brandImageExtractor.start();
```
wwajerowicz: If `MultiplexingVodContentListener` calls `beforeContent()` on all it's delegates, it's not necessary to call the method on delegates directly

src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
```diff
@@ -131,8 +150,9 @@ public void runTask() {
                     itemExtractor.getResult().getProcessed(),
                     itemExtractor.getResult().getFailures()));
             
-            contentGroupUpdater.finish();
-            oldContentDeactivator.finish();
+            contentGroupUpdater.afterContent();
+            staleTopicContentRemover.afterContent();
+            oldContentDeactivator.afterContent();
```
wwajerowicz: same with `afterContent()`

src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
```diff
@@ -28,21 +32,30 @@
...
+    private final TopicWriter topicWriter;
+    private final BtVodContentMatchingPredicate newFeedContentMatchingPredicate;
+    private final Topic newTopic;
+    private final BtVodStaleTopicContentRemover staleTopicContentRemover;
     
     public BtVodUpdater(ContentResolver resolver, ContentWriter contentWriter,
             BtVodData vodData, String uriPrefix,
             BtVodContentGroupUpdater contentGroupUpdater,
-            BtVodDescribedFieldsExtractor describedFieldsExtractor,
             Publisher publisher, BtVodOldContentDeactivator oldContentDeactivator,
```
wwajerowicz: Instead of injecting `BtVodContentGroupUpdater`,`BtVodOldContentDeactivator `,`BtVodOldContentDeactivator` we could just inject a single `MultiplexingVodContentListener`

tfm: Good idea. Will do as part of the refactor.

PR #684, opened 2015-07-14
Ingest explicit series from BT VoD feed.

wwajerowicz:


src/test/java/org/atlasapi/remotesite/btvod/BtVodExplicitSeriesWriterTest.java
```diff
@@ -0,0 +1,127 @@
...
+                Sets.<String>newHashSet(),
+                seriesUriExtractor
+        );
+    }
+
+
+    @Test
+    public void testDoesntExtractsSeriesFromEntryWhichIsNotSeason() {
+        BtVodEntry entry = row();
+        entry.setProductType("episode");// "Apr 30 2014 12:00AM"
```
tfm: Why this datetime?

src/main/java/org/atlasapi/remotesite/btvod/BtVodSeriesProvider.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
+import org.atlasapi.media.entity.Series;
+import org.atlasapi.remotesite.btvod.model.BtVodEntry;
+
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class BtVodSeriesProvider {
```
tfm: Javadoc would be useful here.

PR #713, opened 2015-10-01
BT ingest: write everything at the end of ingest.

wwajerowicz:
Plus some refactor.


tfm:
Do we need to modify the `MergingContentWriter`'s `TopicMergeStrategy` to be `REPLACE`?

PR #720, opened 2015-10-19
fixed channel 5 ingesting default images

Spartakovic:


src/main/java/org/atlasapi/remotesite/five/FiveBrandProcessor.java
```diff
@@ -211,7 +211,9 @@ private String childValue(Element element, String childName) {
     private Maybe<String> getImage(Element element) {
         Elements imageElements = element.getFirstChildElement("images").getChildElements("image");
         if (imageElements.size() > 0) {
-            return Maybe.just(imageElements.get(0).getValue());
+            if (!imageElements.get(0).getValue().contains("api-images.channel5.com/images/default")) {
```
Dalamar42: Since you have the same logic in two places you could extract the if branch condition to a method to avoid duplication.

PR #729, opened 2015-10-27
Ingest release date from Nitro

Spartakovic:
This is ingested into a release date, of type first broadcast. We will use this as a sort order for upload to YouView where there is no position provided by Nitro.

`ContentMerger` now merges release dates, too.


Dalamar42:
LGTM

PR #733, opened 2015-10-29
Add back-to-backs title santisation

tfm:


src/main/java/org/atlasapi/remotesite/btvod/TitleSanitiser.java
```diff
@@ -20,11 +20,17 @@
 
     private static final Map<Pattern, String> PATTERNS_TO_REMOVE = ImmutableMap.<Pattern, String>builder()
             .put(Pattern.compile("ZQ[A-Z]{1}"), "")
-            .put(Pattern.compile("_"), " ")
-            .put(Pattern.compile("^Back to Backs - "), "")
+            .put(Pattern.compile("_"), " ")            
             .put(Pattern.compile("\\(Curzon\\).*(-\\sHD)?$"), "")
             .put(Pattern.compile(" \\(Before DVD\\)$"), "")
             .put(Pattern.compile(" : Coming Soon$"), "")
+            .put(Pattern.compile(" - Volume \\d+ - "), "")
```
Dalamar42: We don't have tests for the Vol and Volume scenarios do we?

src/main/java/org/atlasapi/remotesite/btvod/TitleSanitiser.java
```diff
@@ -20,11 +20,17 @@
...
+            .put(Pattern.compile(" - Vol\\.? \\d+ - "), "")
+            .put(Pattern.compile(" - Vol\\.? \\d+ - "), "")
+            .put(Pattern.compile("^Back to Backs - .* - Back to Back \\d+ - "), "")
+            .put(Pattern.compile("^Back to Backs - Back to Back \\d+ - "), "")
+            .put(Pattern.compile("^Back to Backs - .* - Back to Back - "), "")
```
Dalamar42: Why do the first two back to back have \\d+, but the last two don't.

PR #734, opened 2015-10-29
Add collection title format

tfm:
Tidy up existing collection rules pulling them into
extraction patterns rather than replacement patterns, too.


Dalamar42:
LGTM

PR #735, opened 2015-10-29
Event Refs - not ready for merge yet, for comments please!

ljeyaratnam:
- ModelContentTransformer - changes to handle event refs

### DIFF
--- a/src/main/java/org/atlasapi/query/v2/ContentWriteController.java
+++ b/src/main/java/org/atlasapi/query/v2/ContentWriteController.java
@@ -188,7 +188,7 @@ public class ContentWriteController {
         existing.setKeyPhrases(update.getKeyPhrases());
         existing.setClips(merge ? merge(existing.getClips(), update.getClips()) : update.getClips());
         existing.setPriority(update.getPriority());
-
+        existing.setEventRefs(merge ? merge(existing.events(), update.events()) : update.events());
         if (existing instanceof Item && update instanceof Item) {
             return mergeItems((Item)existing, (Item) update);
         }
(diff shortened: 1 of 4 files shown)


Dalamar42:
LGTM (looks good to me)

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -124,7 +125,9 @@ private Void deserializeAndUpdateContent(HttpServletRequest req, HttpServletResp
         if (Strings.isNullOrEmpty(content.getCanonicalUri())) {
             return error(resp, HttpStatusCode.BAD_REQUEST.code());
         }
-        
+
+        content = updateEventPublisher(content);
```
Dalamar42: Better move this inside the complexify method as this is part of that process

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -250,5 +253,15 @@ private Void error(HttpServletResponse response, int code) {
         response.setContentLength(0);
         return null;
     }
-    
+
+    private Content updateEventPublisher(Content content) {
+        for(EventRef eventRef: content.events()) {
```
Dalamar42: You shouldn't be modifying EventRefs in place even the model allows for it. We want our models to be immutable and someone could change content.events() to return a copy of the internal collection. Instead get the events collection, modify it and set the value back in Content

Dalamar42: Then if someone wants to make Content immutable they will see the setter being used and will know to account for it.

PR #739, opened 2015-10-30
Add scheduled task that ingests off-schedule content from Nitro

emilssolmanis:
**NOTE**: it's based against the YV-force-update branch because we need the refactoring that's done as part of that feature.


src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -0,0 +1,47 @@
...
+
+        ImmutableSet<Item> items;
+        try {
+            items = contentAdapter.fetchEpisodes(query);
+        } catch (NitroException e) {
+            throw new RuntimeException(e);
+        }
+
+        for (Item item : items) {
+            contentWriter.createOrUpdate(item);
```
tfm: We're missing the `LocalOrRemoteNitroFetcher` functionality here, to fetch the hierarchies, aren't we?

Dalamar42:
LGTM

JamieAP:
Defo Java :+1:

src/main/java/org/atlasapi/remotesite/bbc/nitro/LocalOrRemoteNitroFetcher.java
```diff
@@ -115,8 +115,22 @@ public LocalOrRemoteNitroFetcher(ContentResolver resolver, NitroContentAdapter c
         this.fullFetchPermitted = checkNotNull(fullFetchPermitted);
         this.contentMerger = checkNotNull(contentMerger);
     }
-    
-    public ResolveOrFetchResult<Item> resolveOrFetchItem(Iterable<Broadcast> broadcasts) throws NitroException {
+
+    public ResolveOrFetchResult<Item> resolveItems(Iterable<Item> items)
```
tfm: I wonder if this signature should be `resolveOrFetch(ProgrammesQuery query)` which then does the query using the `NitroContentAdapter`, which is already a collaborator.

It's odd to have the items passed to it, and have a `ResolveOrFetchResult`, since it's not fetching; it's actually resolving and merging.

PR #741, opened 2015-11-02
Update brand description from oldest series description

Dalamar42:


JamieAP:
Yeah this looks like Java :+1:

PR #746, opened 2015-11-04
Feature/tve certificates

Dalamar42:


src/test/java/org/atlasapi/remotesite/btvod/CertificateUpdaterTest.java
```diff
@@ -0,0 +1,69 @@
...
+        certificateUpdater.updateCertificates(target, source);
+
+        verifyExpectedCertificatesInContent(target, certificateA, certificateB, certificateC);
+    }
+
+    private void verifyExpectedCertificatesInContent(Content content, Certificate... certificates) {
+        assertThat(content.getCertificates().size(), is(certificates.length));
+
+        assertThat(content.getCertificates().containsAll(ImmutableList.copyOf(certificates)),
+                is(true));
```
emilssolmanis: Arguably this would read better as just an `assertTrue`.

PR #754, opened 2015-11-09
Change series provider to attempt to resolve explicit series by paren…

Dalamar42:
…t ref and series number if parent guid is not present


tfm:
:+1: although can you include a comment as to why we do this?

Dalamar42:
ack

PR #756, opened 2015-11-11
Ingest service ID and advertise from dates from BT

tfm:
…MpxChannelDataIngester, added logic to save aliases and availableDates in BtChannelDataUpdater and BtMpxChannelDataIngester, Added additional fields to extract from Json in GsonBtMpxClient, Added Spring wiring and bean in BtChannelsModule class


pom.xml
```diff
@@ -228,6 +228,14 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
tfm: Why was this added?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private final SubstitutionTableNumberCodec codec = new SubstitutionTableNumberCodec();
+
+    public void addAliasesAndAvailableDateToChannel(PaginatedEntries paginatedEntries, ChannelResolver channelResolver) {
+
+        List<Entry> entries = paginatedEntries.getEntries();
+
+        Set<Long> channelIdsThatHaveAliasesAndAvailableDateAdded = Sets.newHashSet();
+
+        for(int i = 0; i < entries.size(); i++) {
```
tfm: Can't we use a for (Entry entry : paginatedEntries.getEntries() ) loop here?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+            Entry currentEntry = entries.get(i);
+            String linearEpgChannelId = currentEntry.getLinearEpgChannelId();
+            DateTime advertiseAvailableDate = currentEntry.getAvailableDate();
+            long currentGuid = codec.decode(currentEntry.getGuid()).longValue();
+
+            Maybe<Channel> channelMaybe = channelResolver.fromId(currentGuid);
+
+            if(channelMaybe.hasValue()) {
```
tfm: What if there's no value?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+                Channel channel = channelMaybe.requireValue();
+                Set<Alias> aliases = channel.getAliases();
+
+                aliases.removeIf(alias -> {
```
tfm: Don't use Java 8

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+
+                    if (alias.getNamespace().equals(linearEpgChannelId)) {
+                        return true;
+                    }
+                    return false;
+                });
+
+                //Using "namespace" as namespace for now.
+                channel.addAlias(new Alias("namespace", linearEpgChannelId));
```
tfm: Pass in the alias namespace as a constructor arg

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+
+                //Turned off for now until it is added to atlas model and atlas deer. Did a git pull but couldn't get dias' changes probably because it is only on stage. (my branch is based off master)
```
tfm: Remember whether we ingest the available from date needs to be configurable based on the environment. We should similarly only wipe it, if we're ingesting it.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.StreamSupport;
+
+public class BtChannelDataUpdater {
+
+    private final SubstitutionTableNumberCodec codec = new SubstitutionTableNumberCodec();
+
+    public void addAliasesAndAvailableDateToChannel(PaginatedEntries paginatedEntries, ChannelResolver channelResolver) {
```
tfm: The resolver should be a field, not passed in to this method. Passing in here breaks encapsulation, since someone using this class shouldn't need to know which `ChannelResolver` a `BtChannelUpdater` needs.

pom.xml
timothy22000: I needed this to write lambda expressions (probably auto-generated by IntelliJ when I used the little light bulb button to fix the problem). If I remove it, it would give me "Lambda expressions are not supported at this language level".

tfm: Unfortunately you can't just upgrade to Java 8, since it's not running on a Java 8 VM yet.

timothy22000: I removed this and tweak the Module settings in Project Structure and it seems to be ok now and won't give the error above.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: This explains the pom.xml comment above now.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: Yes, we can. I was using it out of habit. :)

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: Why are we not using Java 8? Is it because it would break the codebase?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: Not too sure what you mean. The only two things that I can think of relating to this comment is:

plproductmetadata$linearEpgChannelId: "urn:BT:linear:service:751764"

i) use alias.getNamespace() to replace "namespace"? but alias.getNamespace() is "urn:BT:linear:service:751764" which is being used as the value. Another is that alias.getNamespace() would be removed in the step before which would make alias.getNamespace() empty.

ii) Do you mean use plproductmetadata$linearEpgChannelId as namespace?

pom.xml
timothy22000: Ignore the above. Saw your comment below about not using Java 8.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: My thought process was do nothing (Would not add that channel's channelId to the channelIdsThatHaveAliasesAndAvailableDate Set). Thus, it will not be filtered out by the Iterable.filter method below when removing aliases from channelIdsThatDidNotHaveAliasesAndAvailableDateSet.

How would you like me to handle the case with no value?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        removeStaleAliasesAndAvailableDateFromChannel(channelIdsThatHaveAliasesAndAvailableDateAdded, channelResolver.all());
+    }
+
+    private void removeStaleAliasesAndAvailableDateFromChannel(Set<Long> channelIdsThatHaveAliasesAndAvailableDateAdded, Iterable<Channel> channels) {
+        Iterable<Channel> channelsThatHasStaleAliasesAndAvailableDates = Iterables.filter(channels,
+                                                                            shouldRemoveValuesFromChannel(channelIdsThatHaveAliasesAndAvailableDateAdded));
+
+        StreamSupport.stream(channelsThatHasStaleAliasesAndAvailableDates.spliterator(), false).forEach(channel -> {
```
timothy22000: I won't be able to use StreamSupport as well since it is Java 8. Will move to good ol loops or Guava's FluentIterable.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: This was the last task on the ticket and I wasn't sure how to do it yet. Wanted to ask you about it in the morning :) I need to investigate this a bit more.

Idea so far:
Each environment creates a BtAllChannelsChannelGroupUpdater with different publisher. (Publisher.BT_TV_CHANNELS_TEST2 and others) which seems to be set in ChannelGroup under the createOrQueryBtAllChannelsChannelGroup().

Approach: Get publisher from ChannelGroup to be able to tell which environment it is coming from.

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
tfm: Perhaps a log entry? Think about someone trying to debug a problem and what they'd like to see in logs when doing so.

timothy22000: Ok!

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: I am currently only ingesting availableFrom date if it is only from prod. (latest commit)

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Set;
+
+import static com.google.common.base.Predicates.not;
+import static com.google.gson.internal.$Gson$Preconditions.checkNotNull;
+
+public class BtChannelDataUpdater {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(BtChannelDataUpdater.class);
+
+    private final SubstitutionTableNumberCodec codec = new SubstitutionTableNumberCodec();
```
tfm: this should be lowercaseonly codec

MartinsIrbe:
LGTM

PR #763, opened 2015-11-11
added query value for advertise_from

Spartakovic:


src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -169,16 +186,33 @@ public boolean apply(ChannelNumbering input) {
...
+        filteredGroup.setChannelNumberings(filtered);
+        return filteredGroup;
+    }
+
     private boolean hasMatchingGenre(Channel channel, Set<String> genres) {
         return !Sets.intersection(channel.getGenres(), genres).isEmpty();
     }
 
+    private boolean isAdvertised(Channel channel) {
+        return channel.getAdvertiseFrom().isBeforeNow() || channel.getAdvertiseFrom().isEqualNow();
```
tfm: This isn't null safe.

tfm:
This needs refactoring. The controller isn't the right place to do this filtering. However, this needs to go out immediately, so the refactor will be done after the merge to `master`

PR #777, opened 2015-11-12
Events endpoint basic data

ljeyaratnam:


src/test/java/org/atlasapi/input/ItemModelTransformerTest.java
```diff
@@ -72,6 +72,8 @@ public void testTransformItemWithBroadcastVersionsTransformsAllVersionFields()
 
     @Test
     public void testTransformItemWithEventRefs() {
+        given(idCodec.decode("12345")).willReturn(BigInteger.valueOf(12345));
```
emilssolmanis: Don't do BDD methods in tests that aren't BDDs, this should be `when(mock.thing()).thenReturn(null)`

PR #778, opened 2015-11-12
Feature/tve channel

Dalamar42:


src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
```diff
@@ -60,21 +58,18 @@ public BtVodUpdater(
             ImageExtractor imageExtractor,
```
emilssolmanis: Not that you made it this way, but that is one **long** list of collabs, which kind of smells.

emilssolmanis:
Nothing super obviously wrong from what I can see. :+1: :shipit:

src/main/java/org/atlasapi/remotesite/btvod/BtVodUpdater.java
Dalamar42: Yeah that's very smelly. The whole wiring of this ingester is worth a rethink.

PR #788, opened 2015-11-30
Disable hierarchy scoring for subscription catchup brands

tfm:
These brands only have a few episodes available, so hierarchy
matching will score badly, since hierarchies will differ
greatly from equivalence candidates with fuller catalogues
(such as broadcast data sources)

### DIFF
--- a/src/test/java/org/atlasapi/equiv/scorers/SubscriptionCatchupBrandDetectorTest.java
+++ b/src/test/java/org/atlasapi/equiv/scorers/SubscriptionCatchupBrandDetectorTest.java
@@ -1,0 +1,155 @@
+package org.atlasapi.equiv.scorers;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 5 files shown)


src/test/java/org/atlasapi/equiv/scorers/SubscriptionCatchupBrandDetectorTest.java
```diff
@@ -0,0 +1,156 @@
...
+        Brand brand = new Brand();
+        Series s1 = series("http://example.org/s1", 1);
+        Series s2 = series("http://example.org/s2", 2);
+        Series s3 = series("http://example.org/s3", 3);
+        
+        assertFalse(detector.couldBeSubscriptionCatchup(brand, ImmutableList.of(s1, s2, s3)));
+    }
+
+    @Test
+    // Three seasons is considered too many to be possibly subscription catchup
```
Dalamar42: This comment doesn't match the test

Dalamar42:
One comment. Otherwise LGTM

PR #794, opened 2015-12-03
Fix bug where images/descriptions only came from first row

Dalamar42:
- The existing logic mistakenly selectes images and descriptions from
the first MPX row to be processed and ignored all those that were
deduped into it


emilssolmanis:
lgtm :shipit:

PR #801, opened 2015-12-08
Query content with event refs

ljeyaratnam:
(cherry picked from commit 6254a11)


src/main/java/org/atlasapi/query/content/LookupResolvingQueryExecutor.java
```diff
@@ -34,76 +34,40 @@
 public class LookupResolvingQueryExecutor implements KnownTypeQueryExecutor {
 
     private final Logger log = LoggerFactory.getLogger(LookupResolvingQueryExecutor.class);
-    private final KnownTypeContentResolver cassandraContentResolver;
-    private final boolean cassandraEnabled;
-    private final EquivalentContentResolver mongoEquivsResolver;
-    private final LookupEntryStore lookupEntryStore;
 
-    public LookupResolvingQueryExecutor(KnownTypeContentResolver cassandraContentResolver, KnownTypeContentResolver mongoContentResolver, LookupEntryStore mongoLookupResolver, boolean cassandraEnabled) {
```
tfm: Why did this class need changing so extensively? Was the logic moved elsewhere?

PR #803, opened 2015-12-10
Content event ref

ljeyaratnam:
For panels, we need to be able to query and get list of contents that has a specific event ids. When event and content is matched, we write the new item and it is saved in topLevelItems collection. I have added a functionality to query the topLevelItems collection to get the content uri and query the lookup collection to get list of contents from other collection.


Dalamar42:
LGTM

PR #807, opened 2015-12-17
Updated the Nitro extractor to use Contributions instead of People. A…

MartinsIrbe:
Upated:
- Nitro extractor to use Contributions instead of People.
- the logic for the VersionTypes to use Types from Glycerine model.
- the GlycerineNitroContentAdapter to make use of Contributions instead of People.
- using Images.Image from the Glycerine.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
@@ -1,5 +1,30 @@
 package org.atlasapi.remotesite.bbc.nitro;
 
+import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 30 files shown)


src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -1,30 +1,5 @@
 package org.atlasapi.remotesite.bbc.nitro;
 
-import static com.google.common.base.Preconditions.checkArgument;
```
emilssolmanis: Fix your IDE settings, these imports shouldn't get changed just for the lulz.

emilssolmanis:
LGTM

ljeyaratnam:
LGTM but missing changes to OffScheduleContentIngestTask

PR #808, opened 2015-12-17
Nitro changes

ljeyaratnam:


src/main/java/org/atlasapi/system/PaContentDeactivationPredicate.java
```diff
@@ -0,0 +1,70 @@
+package org.atlasapi.system;
+
+import com.google.api.client.repackaged.com.google.common.base.Strings;
```
emilssolmanis: Wrong import.

src/main/java/org/atlasapi/system/PaContentDeactivationPredicate.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class PaContentDeactivationPredicate implements Predicate<Content> {
+
+    private static final Predicate<String> HAS_FILM_URI = new Predicate<String>() {
+        @Override
+        public boolean apply(String s) {
+            return !Strings.isNullOrEmpty(s) && s.startsWith("http://pressassociation.com/films/");
```
emilssolmanis: Somewhat on the fence about this, the **correct** way would be parsing this as a URL and comparing host + path, since we don't care about protocol. This might just work for the foreseeable future though. @Dalamar42 @tfm ?

ljeyaratnam: ah, Jamie made those changes. when I pulled from master, those changes got merged.

emilssolmanis: So should you undo them or what? Not sure I understand, if Jamie made them on master, why are the in the diff now?

src/main/java/org/atlasapi/system/PaContentDeactivationPredicate.java
ljeyaratnam: again Jamie made those changes

src/main/java/org/atlasapi/system/PaContentDeactivationPredicate.java
Dalamar42: I'd agree with @emilssolmanis about parsing this as a URL and comparing like that. It's more self-documenting I think.

Also yeah is the change showing up in the diff? I wouldn't merge this until we understand what exactly this diff is actually changing on master

emilssolmanis:
LGTM

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -147,7 +148,7 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
         for (List<PidReference> ref : Iterables.partition(refs, NITRO_BATCH_SIZE)) {
             ProgrammesQuery query = ProgrammesQuery.builder()
                     .withPid(toStrings(ref))
-                    .withMixins(TITLES, PEOPLE)
+                    .withMixins(TITLES, PEOPLE, GENRE_GROUPINGS)
```
MartinsIrbe: I don't think you need to add this as this is already in my pull-request with all the other mixins.
https://github.com/atlasapi/atlas/blob/MBST-13615-changing-extractor-for-people-to-contributions/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -58,7 +58,7 @@ public OffScheduleContentIngestTask(
     protected void runTask() {
         ProgrammesQuery query = ProgrammesQuery
                 .builder()
-                .withMixins(ProgrammesMixin.TITLES, ProgrammesMixin.PEOPLE)
+                .withMixins(ProgrammesMixin.TITLES, ProgrammesMixin.PEOPLE, ProgrammesMixin.GENRE_GROUPINGS)
```
MartinsIrbe: This also is included in my pull-request.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PidUpdateController.java
```diff
@@ -44,7 +45,7 @@ public void updatePidFromNitro(HttpServletResponse response, @PathVariable("pid"
             items = contentAdapter
                     .fetchEpisodes(ProgrammesQuery.builder()
                             .withPid(pid)
-                            .withMixins(TITLES, PEOPLE)
+                            .withMixins(TITLES, PEOPLE, GENRE_GROUPINGS)
```
MartinsIrbe: And this too has already been included in my pull-request.

src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/BaseNitroItemExtractor.java
```diff
@@ -155,7 +153,7 @@ private Restriction generateRestriction(Optional<WarningText> warningText) {
     }
 
     private boolean isVersionOfType(com.metabroadcast.atlas.glycerin.model.Version nitroVersion, String versionType) {
-        VersionTypes versionTypes = nitroVersion.getVersionTypes();
+        com.metabroadcast.atlas.glycerin.model.Version.VersionTypes versionTypes = nitroVersion.getVersionTypes();
```
MartinsIrbe: We should be using com.metabroadcast.atlas.glycerin.model.Version.Types
This is also in my pull-request.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
ljeyaratnam: the reason it is in the pull request is that I committed these changes to this branch and cherry picked my commits to merge into stage. I am assuming you committed to stage and cherry picked from stage? I will remove these.

MartinsIrbe:
I think that you need only following:
NitroEpisodeExtractor.class
NitroGenresExtractor.class
NitroItemSource.class
NitroEpisodeExtractorTest.java
NitroGenresExtractorTest.java
Everything else is already in my pull-request - https://github.com/atlasapi/atlas/pull/807

otherwise lgtm :+1:

MartinsIrbe:
LGTM :+1:

PR #812, opened 2016-01-04
Set partition key for equivalence assertion messages

Dalamar42:
- Equivalence assertion messages are not idempotent so we need to
guarantee message ordering by ensuring messages about the same
resource are on the same partition


emilssolmanis:
:shipit:

PR #813, opened 2016-01-06
Copy topics from variants of VoD items

tfm:
Only the merge victim's topics are currently stored; this change
ensure that topics across all variants of an item are stored.


Dalamar42:
LGTM

PR #815, opened 2016-01-06
Mbst 13621 available mixin

ljeyaratnam:


MartinsIrbe:
LGTM

tfm:
Not possible, given the semantic change in meaning between versionAvailabilites and availableVersions. The former returned all versions and avialabilities, the latter only currently-available versions. That causes a problem with updating YouView when revoked content.

PR #817, opened 2016-01-07
Add config for jacoco and sonar

emilssolmanis:


Dalamar42:
I think we should keep dep version numbers in the properties section to make them more visible. Otherwise LGTM

PR #822, opened 2016-01-12
added rules for VF equivalence

Spartakovic:


JamieAP:
lgtm

tfm:
We need to configure these as VOD data sources.

tfm:
:+1:

PR #828, opened 2016-01-20
update to output advertiseFrom field in endpoint and tests for different env

timothy22000:
…ent env for thenw fields

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
+++ b/src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
@@ -109,7 +109,7 @@ public class BtChannelDataUpdater {
 
         Channel channel = channelMaybe.requireValue();
 
-        if (advertiseAvailableDate != null && advertiseAvailableDate.getMillis() > 0) {
+        if (advertiseAvailableDate != null && advertiseAvailableDate.getMillis() >= 0) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -109,7 +109,7 @@ public void addAvailableDateToChannel(PaginatedEntries paginatedEntries) {
 
         Channel channel = channelMaybe.requireValue();
 
-        if (advertiseAvailableDate != null && advertiseAvailableDate.getMillis() > 0) {
+        if (advertiseAvailableDate != null && advertiseAvailableDate.getMillis() >= 0) {
```
tfm: We never expect `advertiseAvailableDate` to be non-null and less than 0 do we? If so the second condition here is redundant

tfm:
Fine, bar one comment

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
timothy22000: For the prod env, the availableDate field is equal to zero which is why I made it advertiseAvailableDate.getMillis() >= 0 so it would be set. But I don't expect it to be less than zero yes. 

http://bt.feed.theplatform.eu/f/wzIRPC/btv-med-feed-linear?form=cjson

PR #831, opened 2016-01-25
 Update so that advertiseFrom field is set even if null

timothy22000:


tfm:
This isn't required. The field will be `null` if not set to anything.

PR #832, opened 2016-01-25
Updating ingesters to set generic image

timothy22000:
- Update ImageSimplifier to not output generic images.
- Updated ingesters for Channel 4 (PMLSD), Channel 5, BBC Nitro with placeholder regex that will be changed once there is actual URLs.

Spreadsheet for candidate generic images:
https://docs.google.com/spreadsheets/d/1zdVekSKB8PjgAVWmlCOyIBgnUioZygmg4pzXZUEXYBQ/edit#gid=0

Updated requirements (4/2/2016):
to, so to confirm: channel 5 generic images should have the image type generic_image_content_player, and bbc should have  generic_image_content_originator

[1:17] 
the output side of things should filter generic_image_content_player

Update (9/2/2016):
- Improved logging for FiveModule so that NoHttpResponseException can be caught and logged properly.
- Added time elapsed for FiveModule in the case it fails.
- Added merging logic in FiveBrandProcessor to update existing brands


src/main/java/org/atlasapi/output/simple/ImageSimplifier.java
```diff
@@ -17,8 +18,11 @@
         }
         org.atlasapi.media.entity.simple.Image simpleImage = new org.atlasapi.media.entity.simple.Image(image.getCanonicalUri());
         if (image.getType() != null) {
-            simpleImage.setImageType(image.getType().getName());
+            if(image.getType() != ImageType.GENERIC) {
```
tfm: Why is this special-cased?

src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroContentExtractor.java
```diff
@@ -38,6 +42,8 @@
 
     private static final Logger log = LoggerFactory.getLogger(NitroContentExtractor.class);
 
+    private static final Pattern GENERIC_PATTERN = Pattern.compile("((http://|https://)?ichef.bbci.co.uk/images/ic/1024x576/).*\\.jpg");
```
tfm: Isn't this far, far, too broad?

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
```diff
@@ -89,6 +95,17 @@ public static void addImages(Described content, String anImage) {
 		if (! Strings.isNullOrEmpty(anImage)) {
 			Matcher matcher = IMAGE_PATTERN.matcher(anImage);
 			if (matcher.matches()) {
+
+				Matcher genericMatcher = GENERIC_PATTERN.matcher(anImage);
+
+				if(genericMatcher.matches()){
```
tfm: Shouldn't the only difference be between generic and non-generic images that we set the type to `ImageType.GENERIC` in the generic case? Why, in the `genericMatcher.matches()` case, are we setting any other fields?

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
```diff
@@ -269,13 +279,22 @@ private String getLocationUri(String originalUri) throws Exception {
...
         if (imageElements.size() > 0) {
             String image = imageElements.get(0).getValue();
+
+            Matcher genericImageUrlMatcher = GENERIC_PATTERN.matcher(image);
+            if(genericImageUrlMatcher.matches()) {
+                Image imageObj = new Image("http://" + image);
+                imageObj.setType(ImageType.GENERIC);
+                return Maybe.just(imageObj);
+            }
             if (!image.contains("api-images.channel5.com/images/default")) {
```
tfm: So we are already doing something to suppress default images for c5? can this be incorporated into the new set of changes?

src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroContentExtractor.java
timothy22000: It was based on the results of the mongo queries. For BBC Nitro, they all had similar patterns so I thought that was the generic pattern.

I replied to your comment in the jira ticket with the results:
https://jira.metabroadcast.com/browse/MBST-13977

src/main/java/org/atlasapi/output/simple/ImageSimplifier.java
timothy22000: It was for this ticket. https://jira.metabroadcast.com/browse/MBST-13789

The special case is to only set images if it is not generic.

Did I misinterpret the ticket?

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
timothy22000: oh default = generic? I can probably extend the regex to cover the pattern here `api-images.channel5.com/images/default`.

timothy22000: As it suppresses the image completed, we decided not to incorporate default images into the new set of changes.

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
timothy22000: Updated.

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
timothy22000: The support issues url should been handled by the second check for `api-images.channel5.com/images/default`

src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroContentExtractor.java
```diff
@@ -179,4 +195,11 @@ private MediaType computeMediaType(MasterBrand masterBrand) {
                MediaType.VIDEO;
     }
 
+    private List<String> createGenericUrlList() {
+        ArrayList<String> list = Lists.newArrayList();
```
Dalamar42: Nitpick: Declare this variable using the interface List<String>

src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroContentExtractor.java
```diff
@@ -179,4 +195,11 @@ private MediaType computeMediaType(MasterBrand masterBrand) {
...
+
+        list.add("http://ichef.bbci.co.uk/images/ic/1024x576/p028s846.png");
+
+        return ImmutableList.copyOf(list);
```
Dalamar42: Why not just return ImmutableList.of("...");

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
```diff
@@ -63,7 +66,10 @@
 
 
 	private static final Pattern IMAGE_PATTERN = Pattern.compile("(http.+?)\\d+x\\d+(\\.[a-zA-Z]+)");
-	
+
+	//TODO update with actual generic pattern once tom confirms
```
Dalamar42: Please don't merge this with the TODO still pending :)
Those things have a tendency to still be around years after they were originally written

Dalamar42: Also GENERIC_IMAGE_PATTERN would make it more obvious what this is for

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
Dalamar42: I'm a bit confused about this. If I understand this correctly this will not create an image at all if it's not generic. Is that intended?

Shouldn't this be:
`
Image image = new Image(anImage);
if (genericMatcher.matches() {
  image.setType(ImageType.GENERIC);
}
content.setImages(ImmutableSet.of(image));
`

?

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
```diff
@@ -46,6 +50,7 @@
 public class FiveEpisodeProcessor {
 
     private static final Logger log = LoggerFactory.getLogger(FiveEpisodeProcessor.class);
+    private static final Pattern GENERIC_PATTERN = Pattern.compile("((https://|http://)?i1.ytimg.com/vi/).*\\.jpg");
```
Dalamar42: Similar: GENERIC_IMAGE_PATTER is more descriptive

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
```diff
@@ -269,13 +279,22 @@ private String getLocationUri(String originalUri) throws Exception {
...
 
-    private Maybe<String> getImage(Element element) {
+    private Maybe<Image> getImage(Element element) {
         
         Elements imageElements = element.getFirstChildElement("images").getChildElements("image");
+
         if (imageElements.size() > 0) {
             String image = imageElements.get(0).getValue();
+
+            Matcher genericImageUrlMatcher = GENERIC_PATTERN.matcher(image);
```
Dalamar42: Same comment as before. Why is this if statement doing any else other than setting the type? We seem to be duplicating the logic:

`Image imageObj = new Image("http://" + image);`
and
`return Maybe.just(imageObj);`

timothy22000: it is setting the type here so that we can check whether it is Generic in the FiveEpisodeProcessor which this method is being called at. 

I need to add a check there as I notice it is missing.
 :)

Dalamar42: That's not my question. Why are we duplicating:
```
Image imageObj = new Image("http://" + image);
return Maybe.just(imageObj);
```
?

This could be rewritten to not create and return the image in multiple places if you define some guards to return Maybe.nothing() early in the method if appropriate.

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
timothy22000: This is an interesting one. From what I understand, setImage will set the image that will be shown/chosen and setImages will have all the possible images linked to that content. 

So, actually an image is being created in the next line (will leave a comment there) even though I am not setting one here inside setImages. 

That behaviour is not what I want though. Thanks to your suggestion (which is right btw). I want to set all the possible images first in setImages and label the generic ones, then loop through them to set a non-generic one as the chosen image with setImage.

Here is the updated logic:
`if (! Strings.isNullOrEmpty(anImage)) {
			Matcher matcher = IMAGE_PATTERN.matcher(anImage);
			if (matcher.matches()) {

				Matcher genericMatcher = GENERIC_PATTERN.matcher(anImage);

				Image image = new Image((matcher.group(1) + IMAGE_SIZE + matcher.group(2));
				if(genericMatcher.matches()){
					image.setType(ImageType.GENERIC);
				}
				content.setThumbnail(matcher.group(1) + THUMBNAIL_SIZE + matcher.group(2));
				content.setImages(ImmutableSet.of(image));
			}
		}
		
		if(content.getImages().size() > 0) {
			for(Image image : content.getImages()) {
				if(image.getType() != ImageType.GENERIC) {
					content.setImage(image.getCanonicalUri());
				}
			}
		}`

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
```diff
@@ -89,6 +95,15 @@ public static void addImages(Described content, String anImage) {
...
+				Matcher genericMatcher = GENERIC_PATTERN.matcher(anImage);
+
+				if(genericMatcher.matches()){
+
+					Image image = new Image(anImage);
+					image.setType(ImageType.GENERIC);
+					content.setImages(ImmutableSet.of(image));
+				}
 				content.setThumbnail(matcher.group(1) + THUMBNAIL_SIZE + matcher.group(2));
 				content.setImage((matcher.group(1) + IMAGE_SIZE + matcher.group(2)));
```
timothy22000: Image still being set here. This will be replaced in the new logic.

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
timothy22000: ah ok I get what you mean now. I will rewrite the logic.

src/main/java/org/atlasapi/output/simple/DescribedModelSimplifier.java
```diff
@@ -1,30 +1,35 @@
...
+import com.metabroadcast.common.ids.NumberToShortStringCodec;
 import org.atlasapi.feeds.utils.DescriptionWatermarker;
 import org.atlasapi.media.entity.Broadcast;
 import org.atlasapi.media.entity.Described;
 import org.atlasapi.media.entity.Item;
 import org.atlasapi.media.entity.LookupRef;
 import org.atlasapi.media.entity.MediaType;
-import org.atlasapi.media.entity.Publisher;
 import org.atlasapi.media.entity.Specialization;
-import org.atlasapi.media.entity.simple.*;
```
timothy22000: Our style guide and intelliJ's optimize import doing its work.

src/main/java/org/atlasapi/remotesite/five/FiveBrandProcessor.java
```diff
@@ -127,14 +129,35 @@ private Brand extractBrand(Element element) {
...
         } else {
-            return createBrand(element);
+            return brand;
         }
 
     }
 
+    private Brand mergeBrand(Brand current, Brand extracted) {
+
+        current.setCanonicalUri(extracted.getCanonicalUri());
```
tfm: Never change the canonical URI or publisher based on inbound data; it would be a bug if it changed.

timothy22000: got it. will update this. :+1:

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
```diff
@@ -89,8 +92,25 @@ public static void addImages(Described content, String anImage) {
...
+				if(genericImageUrlList.contains(url)){
+					image.setType(ImageType.GENERIC_IMAGE_CONTENT_PLAYER);
+				}
 				content.setThumbnail(matcher.group(1) + THUMBNAIL_SIZE + matcher.group(2));
-				content.setImage((matcher.group(1) + IMAGE_SIZE + matcher.group(2)));
+				content.setImages(ImmutableSet.of(image));
+			}
+		}
+
+		if(content.getImages().size() > 0) {
```
tfm: Why does this check need to exist?

src/main/java/org/atlasapi/remotesite/five/FiveBrandProcessor.java
```diff
@@ -127,14 +129,34 @@ private Brand extractBrand(Element element) {
...
         } else {
-            return createBrand(element);
+            return brand;
         }
 
     }
 
+    private Brand mergeBrand(Brand current, Brand extracted) {
+
+        current.setPublisher(extracted.getPublisher());
```
tfm: We shouldn't set the publisher; we'd never want to change publisher.

src/main/java/org/atlasapi/remotesite/five/FiveBrandProcessor.java
```diff
@@ -208,13 +240,22 @@ private String childValue(Element element, String childName) {
...
 
-    private Maybe<String> getImage(Element element) {
+    private Maybe<Image> getImage(Element element) {
         Elements imageElements = element.getFirstChildElement("images").getChildElements("image");
         if (imageElements.size() > 0) {
             String image = imageElements.get(0).getValue();
-            if (!image.contains("api-images.channel5.com/images/default")) {
-                return Maybe.just(image);
+
+            if(!image.contains("http")) {
```
tfm: Why this check?

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
```diff
@@ -269,14 +274,23 @@ private String getLocationUri(String originalUri) throws Exception {
...
+    private Maybe<Image> getImage(Element element) {
         
         Elements imageElements = element.getFirstChildElement("images").getChildElements("image");
+
         if (imageElements.size() > 0) {
             String image = imageElements.get(0).getValue();
-            if (!image.contains("api-images.channel5.com/images/default")) {
-                return Maybe.just(image);
+
+            if(!image.contains("http")) {
```
tfm: Why this check?

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
```diff
@@ -269,14 +274,23 @@ private String getLocationUri(String originalUri) throws Exception {
         return genreMap.mapRecognised(ImmutableSet.of("http://www.five.tv/genres/" + element.getFirstChildElement("genre").getValue()));
     }
 
-    private Maybe<String> getImage(Element element) {
+    private Maybe<Image> getImage(Element element) {
         
         Elements imageElements = element.getFirstChildElement("images").getChildElements("image");
+
         if (imageElements.size() > 0) {
```
tfm: if (!imageElements.isEmpty())

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroSeriesExtractorTest.java
```diff
@@ -21,6 +24,15 @@ public void testMediaTypeIsSetCorrectly() {
         Assert.assertEquals(MediaType.VIDEO, extractedVideo.getMediaType());
     }
 
+    @Test
+    public void testGenericImageExtraction() {
+        org.atlasapi.media.entity.Series extractedGenericImage = extractor.extract(seriesWithGenericImage());
+        Image genericImage = Iterables.getOnlyElement(extractedGenericImage.getImages());
+
+        Assert.assertEquals("http://ichef.bbci.co.uk/images/ic/1024x576/p028s846.png", extractedGenericImage.getImage());
```
tfm: import static Assert.assertEquals()

src/main/java/org/atlasapi/remotesite/five/FiveBrandProcessor.java
timothy22000: I wanted to cover the case where uri does not have "http://" before it so nothing slips through the filter. similar to NitroImageExtractor.

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
timothy22000: I wanted to cover the case where uri does not have "http://" before it so nothing slips through the filter. similar to NitroImageExtractor.

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
timothy22000: I made this check so that only non-generic_image_content_player image is set at the `image` field. I suppose this is not required if the OutputContentMerger handles it?

tfm: My point was why do you need to check if size > 0, you're then iterating over the image array anyway, and if size is 0, it'll do nothing

src/main/java/org/atlasapi/remotesite/five/FiveEpisodeProcessor.java
timothy22000: unfortunately nu.xom.Elements doesn't have isEmpty(). :(

src/main/java/org/atlasapi/remotesite/channel4/C4AtomApi.java
timothy22000: ah! because forEach implicitly guards against empty list so I can skip that check and just do 
for(Image image : content.getImages()) straightaway

PR #833, opened 2016-01-26
Extracting programmeId as an Aliases on channel4

Spartakovic:
Extracting programmeId as an Aliases


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4EpgEpisodeExtractor.java
```diff
@@ -51,6 +51,11 @@ protected Episode setAdditionalEpisodeFields(Entry entry, Map<String, String> lo
         String startTime = lookup.get(DC_START_TIME);
         Duration duration = C4AtomApi.durationFrom(lookup);
 
+        String programmeId = lookup.get(C4AtomApi.DC_PROGRAMME_ID);
```
tfm: Let's add an `Alias` rather than `AliasUrl`. That way we don't need to match based on a prefix, but can instead just look for an `Alias` with a particular namespace. The namespace should be `gb:channel4:prod:pmlsd:programmeId`

src/test/java/org/atlasapi/remotesite/channel4/pmlsd/C4SeriesAndEpisodesExtractorTest.java
```diff
@@ -66,6 +69,7 @@ public void testParsingASeries() throws Exception {
 		
 		assertThat(firstEpisode.getSeriesNumber(), is(3));
 		assertThat(firstEpisode.getEpisodeNumber(), is(1));
+		assertThat(firstEpisode.getAliases(), is((Set<Alias>) ImmutableSet.of(new Alias("gb:channel4:programmeId", "41337/001"))));
```
tfm: See previous comment; this is pretty much what we should be doing, but I don't see any code in this commit to set `Alias`?

tfm:
Back to you, @Spartakovic

src/test/java/org/atlasapi/remotesite/channel4/pmlsd/C4SeriesAndEpisodesExtractorTest.java
Spartakovic: I was actually setting Aliases at first, but because I screwed up releasing process I thought setting Aliases doesn't work, so I changed implementation by setting aliasUrls instead, will change that @tfm

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4EpisodeGuideEpisodeExtractor.java
```diff
@@ -24,6 +25,10 @@ protected Episode setAdditionalEpisodeFields(Entry entry, Map<String, String> lo
             Episode episode) {
         episode.addAliasUrl(C4AtomApi.canonicalizeEpisodeFeedId(entry));
         episode.addAliasUrl(entry.getId());
+        String programmeId = lookup.get(C4AtomApi.DC_PROGRAMME_ID);
+        if (programmeId != null) {
+            episode.addAlias(new Alias("gb:channel4:prod:pmlsd:programmeId", programmeId));
```
tfm: Move the `namespace` to a common place and a constant

src/test/resources/org/atlasapi/remotesite/channel4/pmlsd/glee-epg2.atom
```diff
@@ -0,0 +1,29 @@
...
+    <dc:relation.SeriesNumber>1</dc:relation.SeriesNumber>
+    <dc:relation.EpisodeNumber>5</dc:relation.EpisodeNumber>
+    <dc:relation.TXChannel>E4</dc:relation.TXChannel>
+    <dc:relation.TXStartTime>Sunday 15 August 2010 05:55 PM</dc:relation.TXStartTime>
+    <dc:relation.Duration>55:00</dc:relation.Duration>
+    <dc:relation.Repeat>true</dc:relation.Repeat>
+    <dc:relation.Subtitles>true</dc:relation.Subtitles>
+    <dc:relation.AudioDescription>true</dc:relation.AudioDescription>
+    <link href="http://api.channel4.com/programmes/glee.atom?apikey=mbq3hr7z9dvqwucjne7t79s6" rel="up" type="application/atom+xml" title="Up"/>
+    <link href="http://www.channel4.com/programmes/glee/episode-guide/series-1/episode-5?client=xbox" rel="EpisodeUrl" type="text/html" title="Episode Url"/>
```
tfm: We need a test without client=xbox

Spartakovic: There is an existing test for that

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4AtomApi.java
```diff
@@ -47,9 +47,11 @@
 	
 	public static final String DC_EPISODE_NUMBER = "dc:relation.EpisodeNumber";
 	public static final String DC_SERIES_NUMBER = "dc:relation.SeriesNumber";
+    public static final String DC_PROGRAMME_ID = "dc:relation.programmeId";
```
tfm: When this is merged to master, please do a formatting commit to fix the tabs in this file.

tfm:
Good to go to production, @Spartakovic

PR #834, opened 2016-01-28
- merge broadcasts if and only if simple item type is broadcast

ljeyaratnam:


tfm:
There are many whitespace changes here. What happened?

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -256,16 +278,36 @@ private Content complexify(Description inputContent) {
...
+                if (Iterables.isEmpty(item.getVersions())) {
+                    item.addVersion(new Version());
+                }
+                Version existingVersion = item.getVersions().iterator().next();
+                Version postedVersion = Iterables.getOnlyElement(update.getVersions());
+                mergeVersions(existingVersion, postedVersion);
+            }
+            return (Content) existing;
+        }
+        throw new IllegalStateException("Entity for " + update.getCanonicalUri() + " not Content");
```
tfm: Turn this into a guard, as it leads to less nesting.

if (!(existing instanceof Item) {
  throw new ...
}

// rest of class

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -256,16 +278,36 @@ private Content complexify(Description inputContent) {
...
     }
 
+    private Content mergeBroadcasts(Maybe<Identified> possibleExisting, Content update) {
+        if (possibleExisting.isNothing()) {
+            throw new IllegalStateException("Entity for " + update.getCanonicalUri() + " does not exists");
+        }
+        Identified existing = possibleExisting.requireValue();
+        if (existing instanceof Item) {
+            Item item = (Item) existing;
+            if (!update.getVersions().isEmpty()) {
```
tfm: Why do need this check? Can't we just have a `for` loop over all `Version`s?

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -256,16 +278,36 @@ private Content complexify(Description inputContent) {
...
+                if (Iterables.isEmpty(item.getVersions())) {
+                    item.addVersion(new Version());
+                }
+                Version existingVersion = item.getVersions().iterator().next();
+                Version postedVersion = Iterables.getOnlyElement(update.getVersions());
```
tfm: What if there's more than one `Version`

ljeyaratnam:
white space - reformatted the code

tfm:
Always separate reformatting commits from functional changes; otherwise it's difficult to see what's changed. Don't worry about it this time round, but for future reference.

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
ljeyaratnam: I followed the existing code pattern to merge just only one versions. Also I am not sure how we can be sure if we were to loop over the update versions and update version and existing version are same

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -216,7 +238,7 @@ private Item mergeItems(Item existing, Item update) {
             mergeVersions(existingVersion, postedVersion);
```
ljeyaratnam: As you can see existing code only merges one version

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -256,16 +278,41 @@ private Content complexify(Description inputContent) {
...
-    
+
     private Void error(HttpServletResponse response, int code) {
         response.setStatus(code);
         response.setContentLength(0);
         return null;
     }
 
+    private Content mergeBroadcasts(Maybe<Identified> possibleExisting, Content update) {
+        if (possibleExisting.isNothing()) {
```
tfm: If `possibleExisting` is nothing, we should not throw. We should save a new item. See previous behaviour with the `merge` method.

PR #835, opened 2016-01-28
Require exact title match on vod content brands

tfm:
Where a dataset may be incomplete, we shouldn't fuzzy
match on titles, since there may not be a match to win
out against a partial title match.


Dalamar42:
LGTM

PR #836, opened 2016-01-28
merging aliases in c4

Spartakovic:


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4AtomBackedBrandUpdater.java
```diff
@@ -401,10 +402,17 @@ private boolean equivalentRestrictions(Restriction existing, Restriction fetched
 
     private <T extends Described> T updateDescribed(T existing, T fetched) {
         
-        SetView<String> mergedAliases = Sets.union(existing.getAliasUrls(), fetched.getAliasUrls());
+        SetView<String> mergedAliasUrls = Sets.union(existing.getAliasUrls(), fetched.getAliasUrls());
```
tfm: Let's not merge aliases, let's overwrite them. Otherwise we'll not remove if they change.

tfm:
@Spartakovic back to you

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4AtomBackedBrandUpdater.java
```diff
@@ -401,10 +402,15 @@ private boolean equivalentRestrictions(Restriction existing, Restriction fetched
...
         
-        if (!mergedAliases.equals(existing.getAliasUrls())) {
-            existing.setAliasUrls(mergedAliases);
+        if (!mergedAliasUrls.equals(existing.getAliasUrls())) {
+            existing.setAliasUrls(mergedAliasUrls);
+            copyLastUpdated(fetched, existing);
+        }
+
+        if (fetched.getAliases() != null) {
```
tfm: You need to check lastUpdated, per the same check for `mergedAliasUrls`

PR #838, opened 2016-01-29
MBST-13163: Added PA genre mapping to MetaBroadcast tags.

MartinsIrbe:
Added PA genre mapping to MetaBroadcast tags for brand, series and episode.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
@@ -1,8 +1,5 @@
 package org.atlasapi.remotesite.pa;
 
-import static org.atlasapi.persistence.logging.AdapterLogEntry.errorEntry;
-import static org.atlasapi.persistence.logging.AdapterLogEntry.warnEntry;
-
 import java.util.List;
 import java.util.Set;
 
@@ -43,12 +40,12 @@ import org.atlasapi.remotesite.pa.listings.bindings.PictureUsage;
 import org.atlasapi.remotesite.pa.listings.bindings.Pictures;
 import org.atlasapi.remotesite.pa.listings.bindings.ProgData;
 import org.atlasapi.remotesite.pa.listings.bindings.StaffMember;
-import org.joda.time.DateTime;
-import org.joda.time.DateTimeZone;
-import org.joda.time.Duration;
-import org.joda.time.Interval;
-import org.joda.time.format.DateTimeFormat;
-import org.joda.time.format.DateTimeFormatter;
+
+import com.metabroadcast.common.base.Maybe;
+import com.metabroadcast.common.intl.Countries;
+import com.metabroadcast.common.media.MimeType;
+import com.metabroadcast.common.text.MoreStrings;
+import com.metabroadcast.common.time.Timestamp;
 
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
@@ -58,11 +55,15 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
-import com.metabroadcast.common.base.Maybe;
-import com.metabroadcast.common.intl.Countries;
-import com.metabroadcast.common.media.MimeType;
-import com.metabroadcast.common.text.MoreStrings;
-import com.metabroadcast.common.time.Timestamp;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.Duration;
+import org.joda.time.Interval;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+import static org.atlasapi.persistence.logging.AdapterLogEntry.errorEntry;
+import static org.atlasapi.persistence.logging.AdapterLogEntry.warnEntry;
 
 public class PaProgrammeProcessor implements PaProgDataProcessor {
     
@@ -87,6 +88,8 @@ public class PaProgrammeProcessor implements PaProgDataProcessor {
     private final PaCountryMap countryMap = new PaCountryMap();
     
     private final GenreMap genreMap = new PaGenreMap();
+
+    private final PaTagMap paTagMap = new PaTagMap();
 
     public PaProgrammeProcessor(ContentWriter contentWriter, ContentResolver contentResolver, AdapterLog log) {
         this.contentResolver = contentResolver;
@@ -242,6 +245,7 @@ public class PaProgrammeProcessor implements PaProgDataProcessor {
         brand.setMediaType(channel.getMediaType());
         setCertificate(progData, brand);
         setGenres(progData, brand);
+        setTags(progData, brand);
 
         selectImages(progData.getPictures(), brand, PA_PICTURE_TYPE_BRAND, PA_PICTURE_TYPE_SERIES, Maybe.<String>nothing());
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -498,6 +504,13 @@ private void setGenres(ProgData progData, Content content) {
         }
     }
 
+    private void setTags(ProgData progData, Content content) {
+        Set<String> tagsFromGenres = paTagMap.map(content.getGenres());
+        if (!tagsFromGenres.isEmpty()) {
```
emilssolmanis: Why do we need this guard?

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private void mapNews() {
+        paTagMap.put("http://pressassociation.com/genres/2F02", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F03", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F04", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F05", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F06", "news-weather");
+    }
+
+    public Set<String> map(Set<String> genres) {
```
emilssolmanis: This needs documentation.

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,168 @@
+package org.atlasapi.remotesite.pa;
+
+import java.util.Map;
+import java.util.Set;
+
+import com.google.api.client.util.Maps;
+import com.google.api.client.util.Sets;
+
+public class PaTagMap {
```
emilssolmanis: This needs documentation.

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public Set<String> map(Set<String> genres) {
+        Set<String> tags = Sets.newHashSet();
+        for (String genre : genres) {
+            if (genre.contains("http://pressassociation.com/genres/")) {
+                tags.add(paTagMap.get(genre));
+            }
+        }
+
+        // Checking if the tags set has only a one tag with the value - film, this means that the
```
emilssolmanis: this means that the... what?

src/test/java/org/atlasapi/remotesite/pa/PaTagMapTest.java
```diff
@@ -0,0 +1,311 @@
...
+
+    private PaTagMap paTagMap;
+
+    @Before
+    public void setUp() {
+        paTagMap = new PaTagMap();
+    }
+
+    @Test
+    public void testTagMappingForFilms() {
```
emilssolmanis: Most of these tests can be refactored and made simpler using [parameterized tests](https://github.com/junit-team/junit/wiki/Parameterized-tests).

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
MartinsIrbe: To check if the mapped tags set isn't empty before adding tags to the content.

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: :+1:

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: :+1:

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: the film is action by default, therefore we are adding `action` tag.

src/test/java/org/atlasapi/remotesite/pa/PaTagMapTest.java
MartinsIrbe: Good idea, however I haven't used parameterised tests before, will look into it when I got more time :+1:

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,176 @@
...
+    private void mapNews() {
+        paTagMap.put("http://pressassociation.com/genres/2F02", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F03", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F04", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F05", "news-weather");
+        paTagMap.put("http://pressassociation.com/genres/2F06", "news-weather");
+    }
+
+    /**
+     * This methods maps PA content genres with MetaBroadcast tags.
```
emilssolmanis: The JavaDoc syntax mandates you to have the summary on the first line, this should complain if you try genning the docs. I.e., it should be 
```
/** Summary
  *
  * Optional extended description
  * 
  * @param ...
  */
```

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,176 @@
...
+
+import com.google.api.client.util.Maps;
+import com.google.api.client.util.Sets;
+
+/**
+ * This class is used for mapping PA content genres with MetaBroadcast tags.
+ */
+public class PaTagMap {
+
+    private final Map<String, String> paTagMap = Maps.newHashMap();
```
emilssolmanis: This should also be immutable, it isn't right now.

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,176 @@
+package org.atlasapi.remotesite.pa;
+
+import java.util.Map;
+import java.util.Set;
+
+import com.google.api.client.util.Maps;
+import com.google.api.client.util.Sets;
+
+/**
+ * This class is used for mapping PA content genres with MetaBroadcast tags.
```
emilssolmanis: This is not proper documentation. Everything you've written here can be deduced from the class name. You should document the parts that can't, like *why* this exists and where / how it's used and link to those classes.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
emilssolmanis: @MartinsIrbe Yeah, but presumably it's an empty set by default anyway, so if you just reset it to empty, that's a noop. Which would make the guard redundant. Or am I wrong?

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: didn't know, will add it :)

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: :+1: will include that

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
MartinsIrbe: yes you are right, will remove it :+1:

emilssolmanis:
lgtm. rebase before merging. :shipit:

PR #842, opened 2016-02-03
Converted using pricing from flat feed to relational feed

Spartakovic:
Itunes was using flat itunes feed for ingesting Location. This PR changes it to use relational feed only

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfPricingSource.java
+++ b/src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfPricingSource.java
@@ -7,9 +7,12 @@ import com.metabroadcast.common.intl.Country;
 public class ItunesEpfPricingSource {
 
     private final EpfPricing row;
-    private final Country country;
+    private final int country;
+
+    public static final int GB_CODE = 143444;
+    public static final int US_CODE = 143441;
     
-    public ItunesEpfPricingSource(EpfPricing row, Country country) {
+    public ItunesEpfPricingSource(EpfPricing row, int country) {
         this.row = row;
         this.country = country;
     }
@@ -17,7 +20,7 @@ public class ItunesEpfPricingSource {
     public EpfPricing getRow() {
         return this.row;
     }
-    public Country getCountry() {
+    public int getCountry() {
         return this.country;
     }
     
--- a/src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfPricingSource.java
+++ b/src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfPricingSource.java
@@ -1,5 +1,7 @@
 package org.atlasapi.remotesite.itunes.epf;
 
+import java.util.Map;
+
 import org.atlasapi.remotesite.itunes.epf.model.EpfPricing;
 
 import com.metabroadcast.common.intl.Country;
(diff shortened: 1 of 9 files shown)


src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfPricingSource.java
```diff
@@ -7,9 +7,12 @@
 public class ItunesEpfPricingSource {
 
     private final EpfPricing row;
-    private final Country country;
+    private final int country;
+
+    public static final int GB_CODE = 143444;
```
tfm: Are these guaranteed to be constant, or is there another table we can look these up from?

src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesPricingLocationExtractor.java
```diff
@@ -14,31 +15,44 @@
...
 
     @Override
     public Maybe<Location> extract(ItunesEpfPricingSource source) {
-        BigDecimal sdPrice = source.getRow().get(EpfPricing.SD_PRICE);
-        if (sdPrice == null) {
+        String countryCode = iso3Code(source.getCountry());
+        Integer storefrontId = source.getCountryCodes().get(countryCode);
+        Integer extractedStorefrontId = source.getRow().get(EpfPricing.STOREFRONT_ID);
+
+        if (extractedStorefrontId.equals(storefrontId)) {
```
tfm: Change this to a guard. Removes unnecessary indentation. ie:

```
if (!extractedStorefrontId.equals(storefrontId)) {
    return Maybe.nothing();
}

BigDecimal...
```

Also consider using `Objects.equal()` since this is an external data source where there's possibly bad data. If so, there's no harm in being null-safe, which `Objects.equal` is.

src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesPricingLocationExtractor.java
```diff
@@ -14,31 +15,44 @@
...
-
-        Currency currency = Currency.getInstance(new Locale("en", source.getCountry().code()));
-        policy.setPrice(new Price(currency, sdPrice.movePointRight(currency.getDefaultFractionDigits()).intValue()));
-
-        location.setPolicy(policy);
+    }
 
-        return Maybe.just(location);
+    private String iso3Code(Country country) {
+        return new Locale("en", country.code()).getISO3Country().toLowerCase();
```
tfm: Can't you use `Locale.ENGLISH`?

src/main/java/org/atlasapi/remotesite/itunes/epf/model/EpfPricing.java
```diff
@@ -51,18 +51,18 @@ public boolean equals(Object that) {
         }
         if (that instanceof EpfPricing) {
             EpfPricing other = (EpfPricing) that;
-            return Objects.equal(this.get(EPISODE_URL), other.get(EPISODE_URL));
+            return Objects.equal(this.get(VIDEO_ID), other.get(VIDEO_ID));
```
tfm: `VIDEO_ID` is the primary key of this table, is it?

tfm:
A few minor comments, so back to you @Spartakovic

src/main/java/org/atlasapi/remotesite/itunes/epf/model/EpfPricing.java
Spartakovic: it is, why?

tfm: Just wanted to confirm, given the change from `EPISODE_URL` to `VIDEO_ID`

PR #843, opened 2016-02-04
Guard against missing brand in Nitro force update controller.

emilssolmanis:
Episodes from a "top level series" might not have a brand container and have a Series set as their container. The brand update logic was relying on an Episode's container being a Brand.


src/main/java/org/atlasapi/remotesite/bbc/nitro/PidUpdateController.java
```diff
@@ -143,7 +143,15 @@ private void updateBrand(
             ImmutableSet<Brand> brand = contentAdapter.fetchBrands(
                     ImmutableList.of(parentPidRef)
             );
-            contentWriter.createOrUpdate(Iterables.getOnlyElement(brand));
+
+            /* handle The Curious Case of Top Level Episode (In The Night). The container can
```
tfm: "The Curious Case of Top Level Series", no?

PR #844, opened 2016-02-04
Pa updates ingest

Spartakovic:
PA provides us with the new type of files. It contains updates for a content that has been ingested before. The model used for PA updates differs slightly from existing model, so I used previous schema and updated existing processor to handle new data without broadcasts and channels.

1.Added new interface to extend existing PaProgrammeProcessor
2.Added scheduled updater task
3.Implemented changes to cope with content without channel
4.Added transformer from new to old model
5.Fixed tests according to the new code
6.New test for new logic


src/main/resources/TVArchive.dtd
```diff
@@ -0,0 +1,248 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Spartakovic: This schema is not being used, but I am uploading it for future reference

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -203,18 +216,27 @@ private ItemAndBroadcast getClosedEpisode(Brand brand, ProgData progData, Channe
...
-        episode.setCurie(CLOSED_CURIE+getClosedPostfix(channel));
         episode.setTitle(progData.getTitle());
         episode.setScheduleOnly(true);
-        episode.setMediaType(channel.getMediaType());
-        
-        Version version = findBestVersion(episode.getVersions());
 
-        Broadcast broadcast = broadcast(progData, channel, zone, updatedAt);
+        if (!channel.isPresent()) {
+            episode.setCanonicalUri(CLOSED_EPISODE);
```
Dalamar42: This means every PA update will end up having the same canonical URI and therefore be the same Atlas item. That can't be what we want here

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -193,8 +204,10 @@ private boolean isClosedBrand(Optional<Brand> brand) {
         return brand.isPresent() && CLOSED_BRAND.equals(brand.get().getCanonicalUri());
     }
 
-    private ItemAndBroadcast getClosedEpisode(Brand brand, ProgData progData, Channel channel, DateTimeZone zone, Timestamp updatedAt) {
-        String uri = CLOSED_EPISODE+getClosedPostfix(channel);
+    private ItemAndBroadcast getClosedEpisode(Brand brand, ProgData progData, Optional<Channel> channel, DateTimeZone zone, Timestamp updatedAt) {
```
Dalamar42: If we are given an update to a close just ignore it. There is nothing to do here

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -238,8 +260,10 @@ private String getClosedPostfix(Channel channel) {
         brand.addAlias(brandAlias);
         brand.setTitle(progData.getTitle());
         brand.setDescription(Strings.emptyToNull(progData.getSeriesSynopsis()));
-        brand.setSpecialization(specialization(progData, channel));
-        brand.setMediaType(channel.getMediaType());
+        if (channel.isPresent()) {
+            brand.setSpecialization(specialization(progData, channel));
```
Dalamar42: You know channel is present here. Pass channel.get() to the specialization(...) method

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -378,11 +404,11 @@ private DateTime fromPaDate(String paDate) {
         return PA_DATE_FORMAT.parseDateTime(paDate).withZone(DateTimeZone.UTC);
     }
     
-    private ItemAndBroadcast getFilmOrEpisode(ProgData progData, Channel channel, DateTimeZone zone, boolean isEpisode, Timestamp updatedAt) {
-        return specialization(progData, channel) == Specialization.FILM ? getFilm(progData, channel, zone, updatedAt) : getEpisode(progData, channel, zone, isEpisode, updatedAt);
+    private ItemAndBroadcast getFilmOrEpisode(ProgData progData, Optional<Channel> channel, DateTimeZone zone, boolean isEpisode, Timestamp updatedAt) {
+        return getBooleanValue(progData.getAttr().getFilm()) ? getFilm(progData, channel, zone, updatedAt) : getEpisode(progData, channel, zone, isEpisode, updatedAt);
```
Dalamar42: Have you confirmed the business logic here, i.e. that it is correct to replace the specialization check with `progData.getAttr().getFilm()`

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -407,21 +433,28 @@ private ItemAndBroadcast getFilm(ProgData progData, Channel channel, DateTimeZon
...
+            return new ItemAndBroadcast(film, Maybe.<Broadcast>nothing());
+        }
+
         
-        return new ItemAndBroadcast(film, Maybe.just(broadcast));
+
     }
     
-    private Broadcast setCommonDetails(ProgData progData, Channel channel, DateTimeZone zone, Item episode, Timestamp updatedAt) {
+    private Broadcast setCommonDetails(ProgData progData, Optional<Channel> channel, DateTimeZone zone, Item episode, Timestamp updatedAt) {
```
Dalamar42: There is some duplication between setCommonDetails and setCommonDetailsWithoutBroadcasts. Extract the common code to a third method to avoid implementing the same thing in two places.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
Dalamar42: And also comment why you will be ignoring it

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -712,8 +782,8 @@ private void setBasicDetails(ProgData progData, Item item) {
         item.addVersion(version);
     }
     
-    protected Specialization specialization(ProgData progData, Channel channel) {
-        if (MediaType.AUDIO.equals(channel.getMediaType())) {
+    protected Specialization specialization(ProgData progData, Optional<Channel> channel) {
+        if (channel.isPresent() && MediaType.AUDIO.equals(channel.get().getMediaType())) {
```
Dalamar42: We said you need to check the data to see if you can get this information from somewhere else. Can't you get an update on a radio item?

src/main/resources/TVArchive.dtd
Dalamar42: :+1: 
Add a comment to that effect to the top of the file

Dalamar42:
There are no tests for the new logic

src/test/java/org/atlasapi/remotesite/pa/archives/PaArchivesUpdaterTest.java
```diff
@@ -0,0 +1,65 @@
...
+        when(resolver.findByCanonicalUris(anyCollection())).thenReturn(resolvedContent);
+        PaProgDataUpdatesProcessor progProcessor = new PaProgrammeProcessor(writer,resolver,log);
+        paUpdatesProcessor = new PaUpdatesProcessor(progProcessor, writer);
+        updater = new PaCompleteArchivesUpdater(store,resultStore,paUpdatesProcessor);
+        when(store.localArchivesFiles(any(Predicate.class))).thenReturn(ImmutableList.of(file));
+        when(store.copyForProcessing(file)).thenReturn(file);
+    }
+
+    @Test
+    public void testRunTask() throws Exception {
```
Dalamar42: What is this testing?

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -110,24 +112,56 @@ public ContentHierarchyAndSummaries process(ProgData progData, Channel channel,
...
             item.setLastUpdated(updatedAt.toDateTimeUTC());
-            
-            return new ContentHierarchyAndSummaries(possibleBrand, possibleSeries, item, itemAndBroadcast.getBroadcast().requireValue(), 
+
+            return new ContentHierarchyAndSummaries(possibleBrand, possibleSeries, item, itemAndBroadcast.getBroadcast().requireValue(),
                     Optional.fromNullable(brandSummary), Optional.fromNullable(seriesSummary));
         } catch (Exception e) {
-        	e.printStackTrace();
-        	log.record(new AdapterLogEntry(Severity.ERROR).withCause(e).withSource(PaProgrammeProcessor.class).withDescription(e.getMessage()));
+            e.printStackTrace();
```
tfm: This does not stay. Must be removed.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -110,24 +112,56 @@ public ContentHierarchyAndSummaries process(ProgData progData, Channel channel,
...
+            log.record(new AdapterLogEntry(Severity.ERROR).withCause(e).withSource(PaProgrammeProcessor.class).withDescription(e.getMessage()));
+        }
+        return null;
```
tfm: Optional.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -421,7 +494,7 @@ private Broadcast setCommonDetails(ProgData progData, Channel channel, DateTimeZ
         //currently Welsh channels have Welsh titles/descriptions 
         // which flip the English ones, resulting in many writes. We'll only take the Welsh title if we don't
     	// already have a title from another channel
-        if (episode.getTitle() == null || !channel.getUri().contains("wales")) {
+        if (episode.getTitle() == null || (!channel.getUri().contains("wales"))) {
```
tfm: Why did we add the brackets?

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -477,6 +531,58 @@ private Broadcast setCommonDetails(ProgData progData, Channel channel, DateTimeZ
...
+    }
+
+    private void setDescription(ProgData progData, Item item, Channel channel) {
+        if (progData.getBillings() != null) {
+            for (Billing billing : progData.getBillings().getBilling()) {
+                if((item.getDescription() == null || (channel !=null && !channel.getUri().contains("wales")))
+                        && billing.getType().equals("synopsis")) {
+                    item.setDescription(billing.getvalue());
+                }
+                if ((item.getShortDescription() == null || (channel !=null && !channel.getUri().contains("wales")))
```
tfm: Why so much repetition of the channel URI condition?

PR #847, opened 2016-02-08
Mbst 13163 adding pa genre mapper for mb tags

MartinsIrbe:
Updated the logic to store the tags as TopicRefs.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
@@ -19,7 +19,7 @@ import com.google.common.collect.ImmutableSet;
 public class PaTagMap {
 
     private final ImmutableMap<String, String> paTagMap;
-    private final String TOPIC = "gb:pa:stage:";
+    private final String TOPIC = "gb:pa:prod:";
 
     public PaTagMap() {
         ImmutableMap.Builder<String, String> mapBuilder = ImmutableMap.builder();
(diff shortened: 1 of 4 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,163 @@
...
+     * This is done to get MetaBroadcast tags for a specific PA content. After that the
+     * PA content tags are added to the PA content tags field.
+     * @param genres - PA content genres that are used for mapping with MetaBroadcast tags.
+     * @return set of MetaBroadcast tags for the PA content.
+     */
+    public Set<String> map(Set<String> genres) {
+        Set<String> tags = Sets.newHashSet();
+        for (String genre : genres) {
+            if (genre.contains("http://pressassociation.com/genres/")) {
+                tags.add(paTagMap.get(genre));
```
JamieAP: Are there any PA genres that we don't have a mapping for? i.e. is this an exhaustive mapping? We're gunna get nulls in our Set otherwise

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+
+        // Checking if the tags set has only a one tag with the value - film, this means that it's action film.
+        if (tags.size() == 1 && tags.contains("film")) {
+            tags.add("action");
+        }
+        return tags;
```
JamieAP: Can we return `ImmutableSet.copyOf(tags)`?

MartinsIrbe: :+1:

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,189 @@
...
+     */
+    private Set<TopicRef> getTopicRefFromTags(Set<String> mappedTags) {
+        ImmutableSet<String> tags = ImmutableSet.copyOf(mappedTags);
+        if(tags.isEmpty()) {
+            return ImmutableSet.of();
+        }
+
+        ImmutableSet.Builder<TopicRef> topicRefBuilder = ImmutableSet.builder();
+        for (String tag : tags) {
+            topicRefBuilder.add(new TopicRef(new Topic(null, TOPIC + tag, tag), null, null, null));
```
JamieAP: We can make this nicer with: 

`
tags.stream().map(tag -> new TopicRef(new Topic(null, TOPIC + tag, tag), null, null, null)).collect(ImmutableCollectors.toSet())`

MartinsIrbe: @JamieAP yes :)

JamieAP: No we cant, Owl is still on Java 7...

MartinsIrbe: :+1:

PR #849, opened 2016-02-10
Remove brand image extraction logic

Dalamar42:
- Brand image extraction has been subsumed by the
DedupedDescriptionAndImageUpdater and the
HierarchyDescriptionAndImageUpdater that implement the most current
spec of image selection. It is interfering with the correct operation
of the current implementation and should have been removed in the
first place when those were added.


tfm:
yay, delete code. :+1:

PR #851, opened 2016-02-12
Normalize spacing around back-to-back delimeters

tfm:
A '/' is used as a delimeter around back-to-back titles.
In some cases there are surrounding spaces, in others not.
These are now normalized, to not having spaces, to allow
a positive match.

An attempt was made to remove all spacing as part of this
change; however, abbreviations for sports teams makes use
of the position of spaces to allow abbrevations to match
full names, so spaces need to be considered to avoid
mis-matches.


MartinsIrbe:
lgtm :)

PR #855, opened 2016-02-16
Move to programme_formats from formats

tfm:
The latter was deprecated and removed. We must also support switching from type Item to type Film, since there will be a number of Items becoming Films.


Dalamar42:
LGTM

PR #856, opened 2016-02-16
Added PaTagMap to map PA genreses to MetaBroadcast tags and storing t…

MartinsIrbe:
Added PaTagMap to map PA genres to MetaBroadcast tags and storing them as TopicRefs for PA content.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaModule.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaModule.java
@@ -23,9 +23,11 @@ import org.atlasapi.persistence.content.ScheduleResolver;
 import org.atlasapi.persistence.content.people.ItemsPeopleWriter;
 import org.atlasapi.persistence.content.people.PersonWriter;
 import org.atlasapi.persistence.content.schedule.mongo.ScheduleWriter;
+import org.atlasapi.persistence.ids.MongoSequentialIdGenerator;
 import org.atlasapi.persistence.logging.AdapterLog;
 import org.atlasapi.persistence.logging.AdapterLogEntry;
 import org.atlasapi.persistence.logging.AdapterLogEntry.Severity;
+import org.atlasapi.persistence.topic.TopicStore;
 import org.atlasapi.remotesite.channel4.epg.BroadcastTrimmer;
 import org.atlasapi.remotesite.channel4.epg.ScheduleResolverBroadcastTrimmer;
 import org.atlasapi.remotesite.pa.archives.PaArchivesUpdater;
@@ -51,6 +53,15 @@ import org.atlasapi.remotesite.pa.persistence.PaScheduleVersionStore;
 import org.atlasapi.remotesite.rt.RtFilmModule;
 import org.atlasapi.s3.DefaultS3Client;
 import org.atlasapi.s3.S3Client;
+
+import com.metabroadcast.common.persistence.mongo.DatabasedMongo;
+import com.metabroadcast.common.scheduling.RepetitionRule;
+import com.metabroadcast.common.scheduling.RepetitionRules;
+import com.metabroadcast.common.scheduling.SimpleScheduler;
+import com.metabroadcast.common.security.UsernameAndPassword;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.joda.time.Duration;
 import org.joda.time.LocalTime;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -59,14 +70,6 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
-
-import com.google.common.collect.ImmutableMap;
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import com.metabroadcast.common.persistence.mongo.DatabasedMongo;
-import com.metabroadcast.common.scheduling.RepetitionRule;
-import com.metabroadcast.common.scheduling.RepetitionRules;
-import com.metabroadcast.common.scheduling.SimpleScheduler;
-import com.metabroadcast.common.security.UsernameAndPassword;
 
 @Configuration
 @Import(RtFilmModule.class)
@@ -95,6 +98,7 @@ public class PaModule {
     private @Autowired ChannelResolver channelResolver;
     private @Autowired ChannelWriter channelWriter;
     private @Autowired DatabasedMongo mongo;
+    private @Autowired @Qualifier("topicStore") TopicStore topicStore;
     // to ensure the complete and daily people ingest jobs are not run simultaneously 
     private final Lock peopleLock = new ReentrantLock();
 
@@ -174,7 +178,8 @@ public class PaModule {
     }
     
     @Bean PaProgDataProcessor paProgrammeProcessor() {
-        return new PaProgrammeProcessor(contentWriter, contentBuffer(), log);
+        return new PaProgrammeProcessor(contentWriter, contentBuffer(), log, new PaTagMap(topicStore,
+                new MongoSequentialIdGenerator(mongo, "topics")));
     }
     
     @Bean PaCompleteUpdater paCompleteUpdater() {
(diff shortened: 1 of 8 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaModule.java
```diff
@@ -195,14 +200,16 @@ private FileUploadResultStore fileUploadResultStore() {
     }
 
     @Bean PaArchivesUpdater paRecentArchivesUpdater() {
-        PaProgDataUpdatesProcessor paProgDataUpdatesProcessor = new PaProgrammeProcessor(contentWriter, contentBuffer(), log);
+        PaProgDataUpdatesProcessor paProgDataUpdatesProcessor = new PaProgrammeProcessor(contentWriter, contentBuffer(), log,
+                new PaTagMap(topicStore, new MongoSequentialIdGenerator(mongo, "topics")));
```
Spartakovic: extract "new PaTagMap" to the method, because you are using it several times

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,248 @@
...
+ * This class is used in PaProgrammeProcessor for mapping the PA genres to
+ * MetaBroadcast tags for the brand, episode and series content. Tags are stored
+ * as a topics as part of the PA content. This is done so that we can use PA content
+ * in the content prioritization algorithm. This further would allow us to filter
+ * the PA content by the priority.
+ */
+public class PaTagMap {
+
+    private final ImmutableMap<String, String> paTagMap;
+    private final String PA_NAMESPACE = "pressassociation.com";
```
tfm: see other namespaces we create, they're generally of the form gb:pressassociation....

src/main/java/org/atlasapi/remotesite/pa/PaModule.java
```diff
@@ -174,7 +178,8 @@ private FileUploadResultStore fileUploadResultStore() {
     }
     
     @Bean PaProgDataProcessor paProgrammeProcessor() {
-        return new PaProgrammeProcessor(contentWriter, contentBuffer(), log);
+        return new PaProgrammeProcessor(contentWriter, contentBuffer(), log, new PaTagMap(topicStore,
+                new MongoSequentialIdGenerator(mongo, "topics")));
```
Dalamar42: Make "topics" a constant rather than a magic string. Something like TOPICS_STORE

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,248 @@
...
+    private final String METABROADCAST_TAG = "http://metabroadcast.com/tags/";
```
tfm: Is this a metabroadcast tag? or a PA tag?

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,248 @@
...
+import com.google.api.client.util.Sets;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static org.atlasapi.media.entity.Publisher.PA;
+
+/** PA genre map for MetaBroadcast tags.
+ *
+ * This class is used in PaProgrammeProcessor for mapping the PA genres to
```
Dalamar42: :+1: for the JavaDoc.

I'd avoid saying where this class is used because that can easily change without this JavaDoc changing to reflect that

Dalamar42: Leave the bits that say what it's doing and why though. Those are good

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /** This method maps PA content genres with MetaBroadcast tags.
+     *
+     * This is done to get MetaBroadcast tags for a specific PA content. After that the
+     * PA content tags are added to the PA content topicRef field.
+     * @param genres - PA content genres that are used for mapping with MetaBroadcast tags.
+     * @return set of MetaBroadcast tags as TopicRef objects for the PA content.
+     */
+    public Set<TopicRef> map(Set<String> genres) {
```
Dalamar42: This would be better named `mapGenresToTopicRefs` so that it explains what it does

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
```diff
@@ -0,0 +1,248 @@
...
+
+/** PA genre map for MetaBroadcast tags.
+ *
+ * This class is used in PaProgrammeProcessor for mapping the PA genres to
+ * MetaBroadcast tags for the brand, episode and series content. Tags are stored
+ * as a topics as part of the PA content. This is done so that we can use PA content
+ * in the content prioritization algorithm. This further would allow us to filter
+ * the PA content by the priority.
+ */
+public class PaTagMap {
```
Dalamar42: This class looks to have a rather complicated API. Do all of its methods need to be exposed as public? e.g. the method that accepts a builder. Does that one need to be public?

It would nice if there is a simple rewrite we can do to simplify the API for someone using this class

src/test/java/org/atlasapi/remotesite/pa/PaBaseProgrammeUpdaterTest.java
```diff
@@ -51,23 +67,7 @@
...
-import com.metabroadcast.common.queue.MessagingException;
-import com.metabroadcast.common.time.DateTimeZones;
-import com.metabroadcast.common.time.TimeMachine;
-import com.mongodb.ReadPreference;
-
-import junit.framework.TestCase;
+import static org.mockito.Mockito.mock;
 
 @RunWith(JMock.class)
 public class PaBaseProgrammeUpdaterTest extends TestCase {
```
Dalamar42: You should add a test here to assert that you are actually getting topic refs from the paTagMap

src/test/java/org/atlasapi/remotesite/pa/PaTagMapTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.api.client.util.Sets;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PaTagMapTest {
```
Dalamar42: There are only 2 tests here, but the class being tested has--I think--4 public methods. This relates to the comment I made on the PaTagMap class. Either there are more scenarios that we are not testing or the API is too complicated and we don't need all those methods to be public

src/main/java/org/atlasapi/remotesite/pa/PaModule.java
Dalamar42: You can also make it a bean so it won't get created multiple times though it's not necessary

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: @tfm will update this :)

src/main/java/org/atlasapi/remotesite/pa/PaModule.java
MartinsIrbe: Good point, make it as a constant. :)

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: @tfm this is MetaBroadcast tag, as we map PA genre to MB tag. Should this be PA_TAG?

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: Good point, think thought about it before. Will remove the part that says where it is being used. :)

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: Good point. Will rename this :)

src/main/java/org/atlasapi/remotesite/pa/PaTagMap.java
MartinsIrbe: @Dalamar42 there is no reason why it couldn't be private. Will change them to be private.

src/test/java/org/atlasapi/remotesite/pa/PaTagMapTest.java
MartinsIrbe: @Dalamar42 we don't need methods to be public. Also I'm testing the logic if either when specific topic exists in db and when it doesn't as the steps then are a bit different. :)

src/test/java/org/atlasapi/remotesite/pa/PaBaseProgrammeUpdaterTest.java
MartinsIrbe: @Dalamar42 good point. :)

PR #858, opened 2016-02-17
PA deletes: Generic children

JamieAP:


pom.xml
```diff
@@ -53,9 +53,6 @@
                     <source>1.7</source>
                     <target>1.7</target>
                     <fork>true</fork>
-                    <compilerArgs>
```
tfm: pourquoi?

src/main/java/org/atlasapi/system/PaContentDeactivator.java
```diff
@@ -141,6 +147,18 @@ private void deactivateContent(
...
 
+    /*
+        Generically described children are not returned by getChildRefs,
+        thus we must check the DB for any children with this content as its container.
+     */
+    private boolean hasNoGenericChildren(Content content) {
+        if (!(content instanceof Container)) {
+            return true;
+        }
+        DBObject dbQuery = where().fieldEquals("container", content.getCanonicalUri()).build();
```
tfm: constant

PR #860, opened 2016-02-17
Split equivalence result files into multiple directories

tfm:
Putting all equivalence results in the same directory causes
fun and games with lots of files in the directory. Instead, we
will hash the filename and create a two-directory deep
structure to split the files up.

We'll attempt to find a results file from both the base
directory and the new hash structure, so that old results can
still be retrieved.


emilssolmanis:
lgtm

PR #863, opened 2016-02-18
Update images and description from source of equal priority as existing

Dalamar42:
- We may see the same source multiple times as different BT VoD entries
get deduped in it and each time which images and descriptions have been
kept as part of the deduping process may have changed so we must
update every time we see a metadata source of equal priority as the
existing one because it will be the same one with potentially more up
to date metadata


emilssolmanis:
lgtm

PR #864, opened 2016-02-19
Fix NPE in PaProgrammeProcessor

Dalamar42:
- Several beans require PaTagMap to be passed in a constructor
parameter. However, PaTagMap was being instantiated in PostConstruct,
but PostConstruct is called after non-lazy beans are created so at
the time of bean creation it is null
- Added defensive null checking in PaProgrammeProcessor constructor
- Added logging to PaProgrammeProcessor
- Fixed broken tests in PaBaseProgrammeUpdaterTest that were failing
due to trying to write into a file system directory that does not
exist on developer systems


MartinsIrbe:
LGTM

Spartakovic:
cool the test is fixed :) 
lgtm

PR #867, opened 2016-02-22
Adding equivalence for EBMS Redbee

timothy22000:


Dalamar42:
Are you sure you want the repetition for this one to NEVER?

timothy22000:
hmm, I am following the other VF equivalence updaters. I think there was a reason why they were NEVER, to show equivalence before and after in the schedule EPG?

Dalamar42:
This means it will never run though unless someone triggers it. Are we ok with that?

timothy22000:
I'll check with Chris and update my PR if it is required :)

PR #869, opened 2016-02-22
extended matching in case of apostrophes

Spartakovic:


JamieAP:
@Spartakovic You've run the existing tests as well as your own?

JamieAP:
If yes :+1:

PR #872, opened 2016-02-23
Fix logic that ignores descriptions for Welsh channels

Dalamar42:
- Because Welsh channels have Welsh title and descriptions we ignore
them unless we don't have any existing values for that item. An
inverted condition meant we were only grabbing titles and descriptions
for Welsh channels instead.
- Refactored logic for handling Welsh title and descriptions to
eliminate code duplication
- Refactored logic for adding common details to items to eliminate
code duplication
- Add test for ingesting broadcast


tfm:
:+1:

PR #875, opened 2016-02-23
Update pa update controller to use POST and to be single-threaded

Dalamar42:
- PA updates are potentially quite expensive so the update controller
will only have one thread to process multiple requests any incoming
requests

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaSingleDateUpdatingController.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaSingleDateUpdatingController.java
@@ -2,37 +2,41 @@ package org.atlasapi.remotesite.pa;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
 
 import javax.annotation.PreDestroy;
 import javax.servlet.http.HttpServletResponse;
 
 import org.atlasapi.media.channel.Channel;
 import org.atlasapi.media.channel.ChannelResolver;
-import org.atlasapi.persistence.content.ScheduleResolver;
 import org.atlasapi.remotesite.pa.data.PaProgrammeDataStore;
-import org.springframework.stereotype.Controller;
-import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaSingleDateUpdatingController.java
```diff
@@ -40,20 +44,36 @@ public void shutDown() {
         executor.shutdown();
     }
 
-    @RequestMapping("/system/update/pa/{dateString}")
+    @RequestMapping(value = "/system/update/pa/{dateString}", method = RequestMethod.POST)
     public void runUpdate(@PathVariable String dateString, HttpServletResponse response) {
-        PaSingleDateUpdater updater = new PaSingleDateUpdater(Executors.newSingleThreadExecutor(), channelProcessor, fileManager, channelResolver, dateString);
+        PaSingleDateUpdater updater = new PaSingleDateUpdater(
+                Executors.newSingleThreadExecutor(),
```
emilssolmanis: Err, why have that `ExecutorService` member then..?

src/main/java/org/atlasapi/remotesite/pa/PaSingleDateUpdatingController.java
```diff
@@ -40,20 +44,36 @@ public void shutDown() {
...
+            method = RequestMethod.POST
+    )
+    public void runUpdate(@PathVariable String dateString, @PathVariable String channelString,
+            HttpServletResponse response) {
         Maybe<Channel> channel = channelResolver.fromKey(channelString);
         if (channel.hasValue()) {
-//            PaProgDataProcessor processor = Boolean.parseBoolean(fillGaps) ? new PaEmptyScheduleProcessor(this.processor, scheduleResolver) : this.processor;
-            PaSingleDateUpdater updater = new PaSingleDateUpdater(Executors.newSingleThreadExecutor(), channelProcessor, fileManager, channelResolver, dateString);
+            PaSingleDateUpdater updater = new PaSingleDateUpdater(
+                    Executors.newSingleThreadExecutor(),
```
emilssolmanis: Same question.

PR #876, opened 2016-02-24
added title expander

Spartakovic:
1.Expands dr to doctor
2.Expands 3 to three
3.added test

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/FilmEquivalenceGenerator.java
+++ b/src/main/java/org/atlasapi/equiv/generators/FilmEquivalenceGenerator.java
@@ -2,8 +2,12 @@ package org.atlasapi.equiv.generators;
 
 import static com.google.common.collect.Iterables.filter;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Pattern;
+
+import javax.annotation.Nullable;
 
 import org.atlasapi.application.v3.ApplicationConfiguration;
 import org.atlasapi.application.v3.SourceStatus;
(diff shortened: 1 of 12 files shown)


src/main/java/org/atlasapi/equiv/generators/TitleExpander.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+public class TitleExpander {
```
tfm: Very general class name here. Can you come up with something more specific?

src/main/java/org/atlasapi/equiv/generators/TitleExpander.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private final Map<String, String> WORDS_TO_EXPAND = ImmutableMap.<String,String>builder()
```
tfm: Use an OptionalMap here

src/main/java/org/atlasapi/equiv/generators/TitleExpander.java
```diff
@@ -0,0 +1,40 @@
...
+            builder.append(word).append(" ");
+        }
+        return builder.toString().trim();
+    }
+
+    private Function<String, String> expander = new Function<String, String>() {
+        @Nullable
+        @Override
+        public String apply(@Nullable String word) {
+            if (WORDS_TO_EXPAND.keySet().contains(word)) {
```
tfm: This then simplifies to return `WORDS_TO_EXPAND.get(word).or(word)`

src/main/java/org/atlasapi/equiv/generators/FilmEquivalenceGenerator.java
```diff
@@ -68,6 +77,8 @@ private ApplicationConfiguration defaultConfigWithSourcesEnabled(Iterable<Publis
         }
         
         Film film = (Film) item;
+        String expandedTitle = titleExpander.expand(film.getTitle());
+        film.setTitle(expandedTitle);
```
tfm: This is very hidden functionality. You should do this close to the search call, and when title scoring.

tfm:
:+1:

PR #880, opened 2016-02-29
added counter for generics

Spartakovic:


tfm:
We should be adding counts for generics to the PA stats so that those equivalent sets with many PA items in them can take account of generic episodes, not to equivalence itself.

PR #886, opened 2016-03-07
Reactivate content that is should not be deactive

JamieAP:
Also, don't deactivate content ingest or updated after PA's archive ID file was generated.


JamieAP:
@tfm Done and done

PR #887, opened 2016-03-08
Add sorting on arbitrary field for YV tasks

emilssolmanis:


src/main/java/org/atlasapi/query/v2/TaskController.java
```diff
@@ -138,6 +137,30 @@ private TaskQuery queryFrom(DestinationType destinationType, Publisher publisher
             TVAElementType type = TVAElementType.valueOf(typeStr.trim().toUpperCase());
             query.withTaskType(type);
         }
+
+        if (orderBy != null) {
```
Dalamar42: I'm not sure this logic belongs here. Could we put it in a parser in TaskQuery.Sort?

emilssolmanis: @Dalamar42 I'm not sure it belongs there either. This is dealing with parsing arbitrary input from the intertoobz into the internal `TaskQuery` object. If we had a `TaskQueryParser`, the whole `queryFrom` method would just go in there, but as is, I think the controller is the most suitable place.

PR #888, opened 2016-03-08
Require masterAgreementServiceType OTG

tfm:
Ingest rules did not correctly specify when content is available
to play. The presence of the masterAgreementServiceType:OTG is
also required (i.e. serviceType:OTG AND
masterAgreementServiceType:OTG required). This means only content
with both these tags are playable on TVE (they may still be present
in the feed though because they are playable on a STB).

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/btvod/BtVodVersionsExtractor.java
+++ b/src/main/java/org/atlasapi/remotesite/btvod/BtVodVersionsExtractor.java
@@ -76,7 +76,8 @@ public class BtVodVersionsExtractor {
         
         locations.addAll(createLocations(row, btTvServiceId, aliases));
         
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/remotesite/btvod/BtVodVersionsExtractor.java
```diff
@@ -210,8 +211,8 @@ private boolean hasServiceTypeOtg(BtVodEntry row) {
         return row.getServiceTypes().contains(OTG_PLATFORM);
     }
 
-    private boolean isItemTvodPlayoutAllowed(BtVodEntry row) {
```
Dalamar42: Was this unused?

Dalamar42:
LGTM

PR #890, opened 2016-03-09
Ingest revised repeat flag from PA

tfm:
Revised repeat means that a broadcast is a repeat, but has
been edited since the original broadcast. In Atlas we do not
make such a distinction, we coalesce PA's repeat and revised
repeat flag into a single field.


JamieAP:
LGTM

PR #891, opened 2016-03-10
setting headers with id when writing topics

Spartakovic:


src/main/java/org/atlasapi/query/v2/TopicWriteController.java
```diff
@@ -96,7 +101,13 @@ public Void writeContent(HttpServletRequest req, HttpServletResponse resp) {
             log.error("Error reading input for request " + req.getRequestURL(), e);
             return error(resp, HttpStatusCode.SERVER_ERROR.code());
         }
-        
+
+        String hostName = Configurer.get("local.host.name").get();
+        resp.setHeader(
+                HttpHeaders.LOCATION,
+                hostName
+                        + "/3.0/content.json?id="
```
tfm: Don't you mean topic?

src/main/java/org/atlasapi/query/v2/TopicWriteController.java
```diff
@@ -45,6 +49,7 @@
     private final ApplicationConfigurationFetcher appConfigFetcher;
     private final TopicStore store;
     private final ModelReader reader;
+    private static final NumberToShortStringCodec codec = SubstitutionTableNumberCodec.lowerCaseOnly();
```
tfm: Do topics use the uppercase or lowercase-only codec? Please double-check.

Spartakovic: they use lowercase codec

Spartakovic: yes they use lowercase

Dalamar42:
LGTM

PR #892, opened 2016-03-10
Use PA film field to decide type of film

tfm:
Instead of using rt_filmnumber, which isn't solely
set on films, we'll use the film flag as provided by PA to
decide if an Item is a Film or not.


Spartakovic:
LGTM

PR #893, opened 2016-03-10
Missing fields in post API

Spartakovic:
*duration
*countries of origin
*year
*series and episode number for episodes

### DIFF
--- a/src/main/java/org/atlasapi/input/ItemModelTransformer.java
+++ b/src/main/java/org/atlasapi/input/ItemModelTransformer.java
@@ -251,7 +251,8 @@ public class ItemModelTransformer extends ContentModelTransformer<org.atlasapi.m
         }
         Set<Restriction> restrictions = Sets.newHashSet();
         Version version = new Version();
-
+        Duration duration = new Duration(inputLocation.getDuration());
+        version.setDuration(duration);
         restrictions.add(createRestrictionForLocation(inputLocation));
 
         setToFirstRestriction(version, restrictions);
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/input/ItemModelTransformer.java
```diff
@@ -120,6 +129,10 @@ protected Item setFields(Item item, org.atlasapi.media.entity.simple.Item inputI
             version.setLastUpdated(now);
             version.setManifestedAs(encodings);
         }
+
+        if (!inputItem.getLocations().isEmpty()) {
```
tfm: Why do you need this condition?

src/main/java/org/atlasapi/input/ItemModelTransformer.java
```diff
@@ -141,6 +154,15 @@ protected Item setFields(Item item, org.atlasapi.media.entity.simple.Item inputI
         return item;
     }
 
+    private void setDuration(Version version, org.atlasapi.media.entity.simple.Item inputItem) {
+        for (org.atlasapi.media.entity.simple.Location location : inputItem.getLocations()) {
```
tfm: We're iterating over all `Location`s and using just one? Is there a better way?

src/main/java/org/atlasapi/input/ItemModelTransformer.java
```diff
@@ -66,12 +68,19 @@ protected Item createOutput(org.atlasapi.media.entity.simple.Item inputItem) {
...
     }
 
+    private Item createItem(org.atlasapi.media.entity.simple.Item inputItem) {
+        Item item = new Item();
+        item.setYear(inputItem.getYear());
+        item.setCountriesOfOrigin(inputItem.getCountriesOfOrigin());
+        return item;
+    }
+
     private Item createFilm(org.atlasapi.media.entity.simple.Item inputItem) {
```
Dalamar42: CountriesOfOrigin and Year should be handled for films as well

Dalamar42: Actually, check if you can standardise this so you only have to do it in one place

PR #894, opened 2016-03-11
Updated Opta event group titles for UEFA Champions and Europa leagues

MartinsIrbe:
Updated Opta event group titles for UEFA Champions and Europa leagues, this is done as we are currently getting the wrong titles for event groups on event queries to Atlas.


Dalamar42:
LGTM

PR #895, opened 2016-03-11
Added separate OPTA credentials for soccer and rugby queries.

MartinsIrbe:
Added separate OPTA credentials for soccer and rugby queries.


src/main/java/org/atlasapi/remotesite/opta/events/HttpOptaEventsFetcher.java
```diff
@@ -58,8 +70,23 @@ public HttpOptaEventsFetcher(Map<OptaSportType, OptaSportConfiguration> sportCon
             if (config == null) {
                 return Optional.absent();
             }
-            
-            String url = String.format(urlPattern, config.feedType(), config.competition(), config.seasonId());
+
+            String url;
+            if (sport == OptaSportType.RUGBY_AVIVA_PREMIERSHIP) {
```
tfm: I think this is too hidden, can you push into config which set of credentials

MartinsIrbe: :+1:

PR #897, opened 2016-03-11
Mbst 14600 updating opta ingester credentials for soccer and rugby queries

MartinsIrbe:
Making use of username and password in SportConfig based of soccer and rugby fetcher.


src/main/resources/config/environment.properties
```diff
@@ -424,7 +424,11 @@ opta.events.http.password=
 opta.events.http.sports.soccer.football_premier_league=f1|8|2015
 opta.events.http.sports.soccer.football_scottish_premier_league=f1|14|2015
 opta.events.http.sports.soccer.football_german_bundesliga=f1|22|2015
+opta.events.http.sports.soccer.username=
```
tfm: Not sure if I'm reading this config correctly, but don't we need a username and password for each of `opta.events.http.sports.soccer.football_premier_league`, `opta.events.http.sports.soccer.football_scottish_premier_league` etc?

src/main/java/org/atlasapi/remotesite/opta/events/HttpOptaEventsFetcher.java
```diff
@@ -58,8 +55,19 @@ public HttpOptaEventsFetcher(Map<OptaSportType, OptaSportConfiguration> sportCon
             if (config == null) {
                 return Optional.absent();
             }
-            
-            String url = String.format(urlPattern, config.feedType(), config.competition(), config.seasonId());
+
+            String urlPattern = String.format(
```
tfm: Can we combine this into a single `String.format` method, rather than two?

src/main/resources/config/environment.properties
MartinsIrbe: @tfm I don't think we need to be more specific because in `OptaEventsModule` we are using following two prefixes for creating configuration for soccer and non football tasks.
```private static final String OPTA_HTTP_SOCCER_CONFIG_PREFIX = "opta.events.http.sports.soccer.";
private static final String OPTA_HTTP_RUGBY_CONFIG_PREFIX = "opta.events.http.sports.rugby.";```
This is used in `sportConfig` method when we are using `Configurer.getParamsWithKeyMatching` to get configuration properties.
So that when we are running soccer task, we would use `OPTA_HTTP_SOCCER_CONFIG_PREFIX` as part of creating the configuration, and `opta.events.http.sports.soccer.` username and password would be used.

File: src/main/java/org/atlasapi/remotesite/opta/events/HttpOptaEventsFetcher.java (hunk 2 above)
Comment (MartinsIrbe) [inline src/main/java/org/atlasapi/remotesite/opta/events/HttpOptaEventsFetcher.java]: @tfm yes :+1:

File: src/main/java/org/atlasapi/remotesite/opta/events/OptaEventsModule.java [hunk 3]
```diff
@@ -142,7 +137,9 @@ private OptaSportConfiguration parseConfig(String sportConfig) {
             = OptaSportConfiguration.builder()
                 .withFeedType(Iterables.get(configItems, 0))
                 .withCompetition(Iterables.get(configItems, 1))
-                .withSeasonId(Iterables.get(configItems, 2));
+                .withSeasonId(Iterables.get(configItems, 2))
+                .withUsername(Iterables.get(configItems, 3))
```
tfm: Is this correct? Isn't `configItems` the pipe-delimited config?

Dalamar42:
LGTM

Dalamar42:
Added one comment on the commit: https://github.com/atlasapi/atlas/commit/5e31da69f97341ce57c83323246d14ed3dfa7cb2

otherwise LGMT

PR #899, opened 2016-03-14
Fix bug where we were ignoring restrictions on locations

Dalamar42:
- Pull restriction transformation to setFields(...) since
restrictions could exist in both broadcasts and locations and are a
field of the version in the complex model
- In the process fix a bug where we were creating a new version per
location, setting its restriction and then discarded it
- Remove special broadcast type handling in createOutput(...) since
broadcasts are handled generically in setFields(...) anyway


JamieAP:
LGTM

PR #900, opened 2016-03-14
Update aliases for MPX ingester

Dalamar42:
- When synthesising brand/series added guid/id of MPX row that
created it as an alias with 'synthesised' namespace
- When updating images/descriptions from the hierarchy add alias
with the guid of the row that propagated its images/descriptions


emilssolmanis:
lgtm

PR #901, opened 2016-03-15
running equivalence at startup

Spartakovic:


src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -157,112 +169,156 @@
...
+            List<ScheduledTask> jobsAtStartup) {
+        taskScheduler.schedule(task, repetitionRule);
+        if (!repetitionRule.equals(RepetitionRules.NEVER)) {
+            jobsAtStartup.add(task);
+        }
+        return jobsAtStartup;
+    }
+
+    private ThreadPoolExecutor createThreadPool(Integer maxThreads) {
+        return new ThreadPoolExecutor(
```
Dalamar42: Use a Guava Executor here to avoid dropping jobs if you have more jobs than the number of threads

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -157,112 +169,156 @@
     private @Autowired RecentEquivalenceResultStore equivalenceResultStore;
     
     private @Autowired KafkaMessagingModule messaging;
+
+    private final int NUMBER_OF_THREADS = 30;
```
Dalamar42: Let's try 10 threads to avoid overloading it on startup. Also try renaming this to say what they threads are for, e.g. NUM_OF_THREADS_FOR_INTERRUPTED_JOBS

Dalamar42: Or better NUM_OF_THREADS_FOR_STARTUP_JOBS

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -157,112 +169,156 @@
...
+            scheduleEquivalenceJob(publisherUpdateTask(BBC_MUSIC).withName("Music Equivalence Updater"), BBC_MUSIC_EQUIVALENCE_REPETITION, jobsAtStartup);
 
-            taskScheduler.schedule(taskBuilder(0, 7)
+            scheduleEquivalenceJob(publisherUpdateTask(PA).withName("PA Equivalence Updater"), RepetitionRules.NEVER, jobsAtStartup);
+            scheduleEquivalenceJob(publisherUpdateTask(BBC).withName("BBC Equivalence Updater"), RepetitionRules.NEVER, jobsAtStartup);
+            scheduleEquivalenceJob(publisherUpdateTask(C4).withName("C4 Equivalence Updater"), RepetitionRules.NEVER,jobsAtStartup);
+            scheduleEquivalenceJob(publisherUpdateTask(C4_PMLSD).withName("C4 PMLSD Equivalence Updater"), RepetitionRules.NEVER, jobsAtStartup);
+            scheduleEquivalenceJob(publisherUpdateTask(BBC_REDUX).withName("Redux Equivalence Updater"), RepetitionRules.NEVER, jobsAtStartup);
+            //scheduleEquivalenceJob(publisherUpdateTask(LOVEFILM).withName("Lovefilm Equivalence Updater"), RepetitionRules.every(Duration.standardHours(12)).withOffset(Duration.standardHours(10)), jobsAtStartup);
```
Dalamar42: Please check the commit log to see why this was commented out and if we can delete it

Spartakovic: Nope, nothing about commenting out

PR #903, opened 2016-03-16
Change namespace for Opta competition event groups

Dalamar42:
- This is so that API users can distinguish which event group
contains the competition title by its namespace


emilssolmanis:
lgtm :shipit:

PR #904, opened 2016-03-16
Adding equivalence updater

timothy22000:
Equiv updater for EBMS VF data

Added it before in a past PR but it seems to have gotten removed.
https://github.com/atlasapi/atlas/pull/867/files


Spartakovic:
lgtm

PR #908, opened 2016-03-17
Fix logic for MPX images and descriptions aliases

Dalamar42:
- The description, long description and images of a brand or series
could have come from different items so we should have separate aliases
for each
- The decision of which images and descriptions a content
should have is taken elsewhere. Therefore when an update source comes
that has the same priority as the existing source we should check the
actual metadata has changed before applying it and updating the alias


emilssolmanis:
:shipit:

PR #909, opened 2016-03-18
Add async option for /content POST API

Dalamar42:
- Add async parameter on the controller to switch between writing
  content immediately and sending a message to a queue for async
  writing
- Create message consumer and worker for processing content write
  messages
- The new worker is added to the QueryExecutorModule to keep all
  write responsibilities in one place and to avoid the more
  complicated dependency graph that a separate WorkerModule would
  have generated
- Remove unused AtlasProcessingWebApplicationContext and processing.xml
- Add QueryExecutorModule to application context for when running as
  processing. When running as an app server it is inherited via
  QueryWebModule instead


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -104,15 +117,40 @@ private Void deserializeAndUpdateContent(HttpServletRequest req, HttpServletResp
...
-            writeExecutor.writeContent(content, inputContent.getType(), merge);
+            if (async) {
+                sendMessage(inputStreamBytes, contentId, merge);
+            } else {
+                content.setId(contentId);
+                writeExecutor.writeContent(content, inputContent.getType(), merge);
+            }
         } catch (Exception e) {
-            log.error("Error reading input for request " + req.getRequestURL(), e);
+            log.error("Error executing request " + req.getRequestURL(), e);
```
emilssolmanis: I've corrected this in some other places already, but `.getRequestURL()` only gets you the path. You probably want to add the request arguments too.

src/main/java/org/atlasapi/query/worker/ContentWriteMessageConfiguration.java
```diff
@@ -0,0 +1,20 @@
+package org.atlasapi.query.worker;
+
+import com.metabroadcast.common.time.Timestamp;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public abstract class ContentWriteMessageConfiguration {
```
emilssolmanis: Huh? What does this actually do?

src/main/java/org/atlasapi/query/worker/ContentWriteMessageSerialiser.java
```diff
@@ -0,0 +1,44 @@
...
+import com.metabroadcast.common.queue.MessageDeserializationException;
+import com.metabroadcast.common.queue.MessageSerializationException;
+import com.metabroadcast.common.queue.MessageSerializer;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Throwables;
+
+public class ContentWriteMessageSerialiser implements MessageSerializer<ContentWriteMessage> {
+
+    private final ObjectMapper mapper;
```
emilssolmanis: You can (and should) make this static, they're designed to be thread-safe and all that jazz. You should optimally only have a few of these around your app, by their recommendations, usually even just one.

Dalamar42: True. I'll fix that.

src/main/java/org/atlasapi/query/worker/ContentWriteMessageConfiguration.java
Dalamar42: It's a pattern we have to for storing the Jackson annotations separately from the model object and then adding them as mix-in annotations. I can see a couple of reasons for it (to avoid adding a million annotations to things like Content for example), but I'm not entirely sold. For now I'm just maintaining the pattern.

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
Dalamar42: Ack

emilssolmanis:
Other than a couple of minor things, lgtm.

src/main/java/org/atlasapi/query/worker/ContentWriteMessageConfiguration.java
emilssolmanis: Not sure I fully understand how it works, but OK.

PR #913, opened 2016-04-04
Amazon deduping

Spartakovic:


Dalamar42:
Don't forget to squash your commits before merging. And make sure the remaining commit has a descriptive message about what is being done

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -119,29 +135,80 @@ public void prepare() {
...
+
+        } else {
+
+            if (content instanceof Episode){
+                Episode episode = (Episode) content;
+                String title = episode.getTitle();
+                String episodeNumber = episode.getEpisodeNumber().toString();
+                String seasonNumber = episode.getSeriesNumber().toString();
+                String brandTitle = episode.getShortDescription();
+                manualHash = title.concat(brandTitle).concat(episodeNumber).concat(seasonNumber).concat(EPISODE);
```
ljeyaratnam: Why not use a StringBuilder here? Instead of converting Episode and Series number to String. 
Also you can use `+ ` instead of concat() method as it will convert the argument to String. However StringBuilder would be better.

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -119,29 +135,80 @@ public void prepare() {
...
+    private Content mergeAliasesAndLocations(Content item, Content seen) {
+        Version version = Iterables.getOnlyElement(item.getVersions());
+        Alias alias = Iterables.getOnlyElement(Iterables.filter(item.getAliases(), AMAZON_ALIAS));
+        version.addAlias(alias);
+        seen.addAlias(alias);
+        seen.addVersion(version);
+        return seen;
+    }
+
+    private Predicate<Alias> AMAZON_ALIAS = new Predicate<Alias>() {
```
Dalamar42: Per code style fields should be above methods and the constructor. Also this should be final and should not be snake case since it is not a constant

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -119,29 +135,80 @@ public void prepare() {
...
-                if (content instanceof Item) {
-                    write(contentMerger.merge(ContentMerger.asItem(identified), (Item) content));
-                } else if (content instanceof Container) {
-                    write(contentMerger.merge(ContentMerger.asContainer(identified), (Container) content));
+                String title = content.getTitle();
+                manualHash = title.concat(SERIES);
+            }
+
+            Content seen = seenContent.get(manualHash);
+            if (seen == null && !seenContent.containsValue(content)) {
```
ljeyaratnam: if (!seenContent.containsKey(manualHash) && !seenContent.containsValue(content)) maybe?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -119,29 +135,80 @@ public void prepare() {
...
+            if (content instanceof Episode){
+                Episode episode = (Episode) content;
+                String title = episode.getTitle();
+                String episodeNumber = episode.getEpisodeNumber().toString();
+                String seasonNumber = episode.getSeriesNumber().toString();
+                String brandTitle = episode.getShortDescription();
+                manualHash = title.concat(brandTitle).concat(episodeNumber).concat(seasonNumber).concat(EPISODE);
+            } else if (content instanceof Film) {
+                Film film = (Film) content;
+                manualHash = "";
```
ljeyaratnam: Use a StringBuilder?

Dalamar42:
LGTM, but you might want a second opinion from someone who is more familiar with the business case

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
Spartakovic: but if it will be final it will have to be snake case as well?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -119,29 +135,80 @@ public void prepare() {
...
-            seenContent.add(content);
-            Maybe<Identified> existing = resolve(content.getCanonicalUri());
-            if (existing.isNothing()) {
-                write(content);
+            checkForDuplicatesSaveForProcessing(content);
+        }
+    }
+
+    private void checkForDuplicatesSaveForProcessing(Content content) {
+        String manualHash;
```
tfm: Why `manualHash`? This is more a `key` than a hash I think

PR #919, opened 2016-04-08
Changing links to parentRefs after extracting content

Spartakovic:
-Previsouly episodes and series could have parentRefs to content that will be deactivated. Now before writing content in we change contentRefs of content of the same brand to one common brand.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
+++ b/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
@@ -2,6 +2,9 @@ package org.atlasapi.remotesite.amazonunbox;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -228,17 +253,32 @@ public void finish() {
...
-            if (existing.isNothing()) {
-                write(content);
-            } else {
-                Identified identified = existing.requireValue();
-                if (content instanceof Item) {
-                    write(contentMerger.merge(ContentMerger.asItem(identified), (Item) content));
-                } else if (content instanceof Container) {
-                    write(contentMerger.merge(ContentMerger.asContainer(identified), (Container) content));
-                }
+        Iterable<Content> brands = Iterables.filter(seenContent.values(), IS_BRAND);
```
tfm: Just do `Iterables.filter(..., Brand.class)`

PR #922, opened 2016-04-12
Added title transformation to the broadcast subset title scorer

Spartakovic:
Same normalization we do for title scorer. To handle cases like Power Rangers: R.P.M. and Power Rangers RPM


src/test/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorerTest.java
```diff
@@ -33,6 +33,10 @@ public void testMatches() {
             itemWithTitle("Doctor Who?"), 
             itemWithTitle("Doctor Who Confidential")
         ));
+        assertEquals(Score.ONE, score(
```
Dalamar42: You need a test case for the `&` -> `and` and `the Something` -> `Something`

Dalamar42:
Two more test cases are needed. Otherwise LGTM

PR #923, opened 2016-04-14
Make YV channel resolver prefer manual overrides

emilssolmanis:
Historically, PA don't provide YV IDs for all channels, usually radio has some missing. We add manual
override aliases for these. Sometimes PA decides to add the YV IDs, which leads to the existence of both
an override and an ingest alias on two different channel IDs.

This then causes problems with the YV channel resolver initialization, because it caches things and
these two IDs clash, which then leads to processing not starting up.

The fix is preferring the manual override even if an ingested ID is present, because we want to
keep ingesting schedule data on the same channel that we started doing it on historically. I.e.,
the override is older and therefore gets precedence.


Dalamar42:
LGTM

PR #924, opened 2016-04-14
transform and simplify complex award

ljeyaratnam:


Dalamar42:
You need to add tests for the transformation before this can be merged

PR #925, opened 2016-04-15
Don't cache channels for youview equivalence

tfm:
The set of channels may change over the lifetime of a JVM.
Instead we'll use a Supplier<Iterable<Channel>> so they
can be recomputed when needed.


emilssolmanis:
:shipit:

Dalamar42:
LGTM

PR #929, opened 2016-04-18
Catching exception thrown when API key has been revoked and returning HTTP response 403

MartinsIrbe:
Catching ApiKeyNotFoundException, RevokedApiKeyException, InvalidIpForApiKeyException exceptions to return error 4xx when revoked api key is being used.


src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -97,7 +100,13 @@ public void listChannels(HttpServletRequest request, HttpServletResponse respons
             @RequestParam(value = PLATFORM_ID_KEY, required = false) String platformId,
             @RequestParam(value = ADVERTISED, required = false) String advertised) throws IOException {
         try {
-            final ApplicationConfiguration appConfig = appConfig(request);
+            final ApplicationConfiguration appConfig;
+            try {
+                appConfig = appConfig(request);
+            } catch (ApiKeyNotFoundException | RevokedApiKeyException | InvalidIpForApiKeyException ex) {
+                outputter.writeError(request, response, FORBIDDEN);
```
emilssolmanis: This seems like going against the current pattern. Why can't we just reuse the same `errorViewFor` pattern and add these exceptions into the `AtlasErrorSummary` class?

src/main/java/org/atlasapi/query/v2/SearchController.java
```diff
@@ -25,13 +28,19 @@
...
+
+import com.metabroadcast.common.http.HttpStatusCode;
 import com.metabroadcast.common.query.Selection;
 import com.metabroadcast.common.text.MoreStrings;
 import org.atlasapi.media.entity.Specialization;
 
 @Controller
 public class SearchController extends BaseController<QueryResult<Identified,?extends Identified>> {
 
+    private static final AtlasErrorSummary FORBIDDEN = new AtlasErrorSummary(new NullPointerException())
```
emilssolmanis: Duplication.

src/main/java/org/atlasapi/query/v2/ScheduleController.java
```diff
@@ -66,6 +74,14 @@ public void schedule(@RequestParam(required=false) String from,
             @RequestParam(required=false) String publisher, 
             HttpServletRequest request, HttpServletResponse response) throws IOException {
         try {
+            Maybe<ApplicationConfiguration> requestedConfig;
```
Dalamar42: If you are going to move this here then you should also remove this whole block:
```
String apiKey = request.getParameter("apiKey");
boolean apiKeySupplied = apiKey != null;
if (apiKeySupplied && requestedConfig.isNothing()) {
        throw new IllegalArgumentException("Unknown API key: " + apiKey);
}
```

PR #936, opened 2016-04-26
Create canonical URIs for Films based on their prodId rather than RT

Spartakovic:
film number. Resolving existing content uses both URI styles to avoid
creating duplicate content. Aliases of the alternate style is added to
the content.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
@@ -486,33 +486,22 @@ public class PaProgrammeProcessor implements PaProgDataProcessor, PaProgDataUpda
     }
 
     private Item getBasicFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Timestamp updatedAt) {
-        String rtFilmAlias = PaHelper.getFilmRtAlias(progData.getRtFilmnumber());
+        String filmAlias = PaHelper.getAlias(progData.getProgId());
         String filmUri = PaHelper.getFilmUri(identifierFor(progData));
-        ResolvedContent possiblePreviousData = contentResolver.findByCanonicalUris(ImmutableList.of(rtFilmAlias, filmUri));
+        Maybe<Identified> possiblePreviousData = contentResolver.findByUris(ImmutableList.of(
+                filmAlias,
+                filmUri
+        )).getFirstValue();
 
         Film film;
-        Maybe<Identified> oldFilm = possiblePreviousData.get(rtFilmAlias);
-        Maybe<Identified> newFilm = possiblePreviousData.get(filmUri);
-        if (oldFilm.hasValue()) {
-            Identified previous = oldFilm.requireValue();
+        if (possiblePreviousData.hasValue()) {
+            Identified previous = possiblePreviousData.requireValue();
             if (previous instanceof Film) {
                 film = (Film) previous;
             } else {
                 film = new Film();
                 Item.copyTo((Episode) previous, film);
             }
-            film.addAlias(PaHelper.getProgIdAlias(progData.getProgId()));
-            film.addAliasUrl(filmUri);
-        } else if (newFilm.hasValue()) {
-            Identified previous = newFilm.requireValue();
-            if (previous instanceof Film) {
-                film = (Film) previous;
-            } else {
-                film = new Film();
-                Item.copyTo((Episode) previous, film);
-            }
-            film.addAlias(PaHelper.getProgIdAlias(progData.getProgId()));
-            film.addAliasUrl(rtFilmAlias);
         } else {
             film = getBasicFilm(progData);
         }
@@ -849,8 +838,6 @@ public class PaProgrammeProcessor implements PaProgDataProcessor, PaProgDataUpda
     
     private Film getBasicFilm(ProgData progData) {
         Film film = new Film(PaHelper.getFilmUri(identifierFor(progData)), PaHelper.getEpisodeCurie(identifierFor(progData)), Publisher.PA);
-        film.addAliasUrl(PaHelper.getFilmRtAlias(progData.getRtFilmnumber()));
-        film.addAlias(PaHelper.getProgIdAlias(progData.getProgId()));
         setBasicDetails(progData, film);
         
         return film;
--- a/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
@@ -30,6 +30,7 @@ import org.atlasapi.media.entity.Version;
 import org.atlasapi.media.util.ItemAndBroadcast;
 import org.atlasapi.persistence.content.ContentResolver;
 import org.atlasapi.persistence.content.ContentWriter;
+import org.atlasapi.persistence.content.ResolvedContent;
 import org.atlasapi.persistence.logging.AdapterLog;
 import org.atlasapi.persistence.logging.AdapterLogEntry;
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -485,19 +486,33 @@ private Item getFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Times
     }
 
     private Item getBasicFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Timestamp updatedAt) {
+        String rtFilmAlias = PaHelper.getFilmRtAlias(progData.getRtFilmnumber());
         String filmUri = PaHelper.getFilmUri(identifierFor(progData));
-        Maybe<Identified> possiblePreviousData = contentResolver.findByCanonicalUris(ImmutableList.of(filmUri)).getFirstValue();
+        ResolvedContent possiblePreviousData = contentResolver.findByCanonicalUris(ImmutableList.of(rtFilmAlias, filmUri));
```
tfm: Can't be `findByCanonicalUris` since we need to look up aliases, no?

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -485,19 +486,33 @@ private Item getFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Times
...
 
         Film film;
-        if (possiblePreviousData.hasValue()) {
-            Identified previous = possiblePreviousData.requireValue();
+        Maybe<Identified> oldFilm = possiblePreviousData.get(rtFilmAlias);
```
tfm: Why did all this logic change? Once we've found the film via whichever ID, the code should be identical.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -833,8 +848,9 @@ private Version findBestVersion(Iterable<Version> versions) {
     }
     
     private Film getBasicFilm(ProgData progData) {
-        Film film = new Film(PaHelper.getFilmUri(identifierFor(progData)), PaHelper.getFilmCurie(identifierFor(progData)), Publisher.PA);
-        
+        Film film = new Film(PaHelper.getFilmUri(identifierFor(progData)), PaHelper.getEpisodeCurie(identifierFor(progData)), Publisher.PA);
+        film.addAliasUrl(PaHelper.getFilmRtAlias(progData.getRtFilmnumber()));
+        film.addAlias(PaHelper.getProgIdAlias(progData.getProgId()));
```
tfm: We are already adding aliases, no?

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
```diff
@@ -435,7 +531,7 @@ private ProgData setupProgData() {
         inputProgData.setTime("11:40");
         Attr threeDAttr = new Attr();
         threeDAttr.setThreeD("yes");
-        threeDAttr.setFilm("yes");
+        threeDAttr.setFilm("no");
```
tfm: Why did this change?

tfm:
Quite a large number of deltas here; wouldn't expect as much to have changed. I've commented the bits I'd have thought would've been the same.

tfm:
back to you, @Spartakovic

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
Spartakovic: because it suppose to create episode progData, not films, otherwise all content would be generated as films

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -485,19 +486,24 @@ private Item getFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Times
...
             Identified previous = possiblePreviousData.requireValue();
             if (previous instanceof Film) {
                 film = (Film) previous;
-            }
-            else {
+            } else {
                 film = new Film();
                 Item.copyTo((Episode) previous, film);
             }
+            film.addAlias(PaHelper.getProgIdAlias(progData.getProgId()));
```
tfm: The existing production code already adds aliases, doesn't it? Otherwise we have problems with back-population.

Spartakovic: You are right, it already sets it I will fix it. But does it change anything? Aliases are sets, so it wouldn't change anything?

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
tfm: I don't understand; this unit test shouldn't have changed the 3d attribute.

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
tfm: It's confusing for someone else at a future date.

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
Spartakovic: the change is for other unit tests, they use mocked up contentresolver that expects episodes, and not films

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -485,19 +486,23 @@ private Item getFilmWithoutBroadcast(ProgData progData, DateTimeZone zone, Times
...
             Identified previous = possiblePreviousData.requireValue();
             if (previous instanceof Film) {
                 film = (Film) previous;
-            }
-            else {
+            } else {
                 film = new Film();
                 Item.copyTo((Episode) previous, film);
             }
+            film.addAliasUrl(filmUri);
```
tfm: Weren't we already doing this?

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -833,8 +838,8 @@ private Version findBestVersion(Iterable<Version> versions) {
     }
     
     private Film getBasicFilm(ProgData progData) {
-        Film film = new Film(PaHelper.getFilmUri(identifierFor(progData)), PaHelper.getFilmCurie(identifierFor(progData)), Publisher.PA);
-        
+        Film film = new Film(PaHelper.getFilmUri(identifierFor(progData)), PaHelper.getEpisodeCurie(identifierFor(progData)), Publisher.PA);
+        film.addAliasUrl(PaHelper.getFilmRtAlias(progData.getRtFilmnumber()));
```
tfm: Weren't we already doing this?

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
```diff
@@ -425,6 +429,98 @@ public void testCreateBroadcastFromProgData() throws Exception {
         assertThat(broadcast.getLastUpdated(), is(updatedAt.toDateTimeUTC()));
     }
 
+    @Test
+    public void testExtractsNewFilmWithEpisodeUri() {
```
tfm: Do we have a test that content that exists with the alias is resolved and used?

PR #937, opened 2016-04-28
content buffer returns content by uris

Spartakovic:


emilssolmanis:
*Your* code is fine, but while we're on this, can we please make this a `LoadingCache` rather than a `HashMap` as well?

tfm:
This breaks existing functionality. You can't store aliases and canonical URIs in the same map, since you'll then expose aliases in the `findByCanonicalUris` call.

PR #938, opened 2016-04-28
Updated the current channel group save logic so that instead of compa…

MartinsIrbe:
…ring the difference between the existing channel group and the channels that has been fetched from the BT API and then updating the channel group, to update the channel group without checking the difference. This has been done as for some weird reason if channels ends up with no reference to channel group however the channel group still reference the channel, the reference to channels in the channel group isn't updated.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
+++ b/src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
@@ -1,7 +1,5 @@
 package org.atlasapi.remotesite.bt.channels;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
```diff
@@ -116,31 +117,17 @@ public AbstractBtChannelGroupSaver(Publisher publisher, ChannelGroupResolver cha
...
-    
-                    @Override
-                    public boolean apply(ChannelNumbering input) {
-                        return !input.getChannelGroup().equals(channelGroup.getId());
-                    }
-                })
-            );
-            channelWriter.createOrUpdate(channel);
+
+    private void addCurrentChannelsToChannelGroup(final ChannelGroup channelGroup, Set<Long> currentChannels) {
```
Dalamar42: `setCurrentChannelsToChannelGroup` since you are not replacing them, not adding to them

Dalamar42:
LGTM

src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
MartinsIrbe: 👍

PR #940, opened 2016-05-03
Opta-Ingester: allow aliases for venues, added a large number of venues

cognitive-i:
Fix production issue with FA cup not showing from Opta feed


src/main/java/org/atlasapi/remotesite/events/EventsUtility.java
```diff
@@ -63,16 +63,26 @@ public EventsUtility(TopicStore topicStore)  {
...
         return Optional.of(resolveOrCreateDbpediaTopic(
-                location, Topic.Type.PLACE, DBPEDIA_NAMESPACE, value.get(), DBPEDIA
+                value.get().title, Topic.Type.PLACE, DBPEDIA_NAMESPACE, value.get().uri, DBPEDIA
         ));
     }
-    
-    public abstract Optional<String> fetchLocationUrl(String location);
+
+    public static class LocationTitleUri {
+        final public String title;
```
Dalamar42: Add getter so it doesn't matter it's an inner class.

Move inner class to bottom per code style

Dalamar42:
LGTM

PR #944, opened 2016-05-10
Adding Glycerin query executor that implements Iterable

MartinsIrbe:
This has been done to reduce the heap overhead of off schedule ingest.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
@@ -7,6 +7,8 @@ import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import javax.annotation.Nullable;
 
@@ -92,7 +94,7 @@ public class GlycerinNitroClipsAdapter {
             if (Iterables.isEmpty(refs)) {
                 return ImmutableMultimap.of();
             }
-            
+
             Iterable<com.metabroadcast.atlas.glycerin.model.Clip> nitroClips
                 = Iterables.transform(Iterables.filter(getNitroClips(refs), isClip), toClip);
             
@@ -107,7 +109,7 @@ public class GlycerinNitroClipsAdapter {
                 }));
                 return ImmutableMultimap.of();
             }
-            
+
             Iterable<List<Clip>> clipParts = Iterables.partition(nitroClips, BATCH_SIZE);
             ImmutableListMultimap.Builder<String, org.atlasapi.media.entity.Clip> clips
                 = ImmutableListMultimap.builder();
@@ -122,8 +124,8 @@ public class GlycerinNitroClipsAdapter {
 
     public List<org.atlasapi.media.entity.Clip> clipsFor(PidReference ref) throws NitroException {
         try {
-            Iterable<com.metabroadcast.atlas.glycerin.model.Clip> nitroClips
-                    = Iterables.transform(Iterables.filter(getNitroClip(ref), isClip), toClip);
+            Iterable<com.metabroadcast.atlas.glycerin.model.Clip> nitroClips = Iterables
+                    .transform(Iterables.filter(getNitroClip(ref), isClip), toClip);
 
             if (Iterables.isEmpty(nitroClips)) {
                 log.warn("No programmes found for clipRefs {}", ref, new Function<PidReference, String>() {
@@ -135,12 +137,12 @@ public class GlycerinNitroClipsAdapter {
                 });
             }
 
-            Iterable<List<Clip>> clipParts = Iterables.partition(nitroClips, BATCH_SIZE);
-            ImmutableList.Builder<org.atlasapi.media.entity.Clip> clips = ImmutableList.builder();
-            for (List<Clip> clipPart : clipParts) {
-                clips.addAll(extractClips(clipPart).values());
+            ImmutableList.Builder<org.atlasapi.media.entity.Clip> extractedClips = ImmutableList.builder();
+            List<Clip> clips = Lists.newArrayList(nitroClips);
+            for (Clip clip : clips) {
+                extractedClips.add(extractClip(clip));
             }
-            return clips.build();
+            return extractedClips.build();
         } catch (GlycerinException e) {
             throw new NitroException(ref.toString(), e);
         }
@@ -159,6 +161,25 @@ public class GlycerinNitroClipsAdapter {
             extracted.put(BbcFeeds.nitroUriForPid(clip.getClipOf().getPid()), clipExtractor.extract(source));
         }
         return extracted.build();
+    }
+
+    private org.atlasapi.media.entity.Clip extractClip(Clip clipPart) throws GlycerinException {
+        final List<Availability> availabilities = getNitroAvailabilities(clipPart);
+        final List<Version> versions = versions(clipPart);
+
+        NitroItemSource<Clip> source = NitroItemSource.valueOf(clipPart, availabilities,
+                ImmutableList.<Broadcast>of(), versions);
+
+        return clipExtractor.extract(source);
+    }
+
+    private List<Version> versions(Clip clip) throws GlycerinException {
+        VersionsQuery query = VersionsQuery.builder()
+                .withDescendantsOf(clip.getPid())
+                .withPageSize(pageSize)
+                .build();
+
+        return exhaust(glycerin.execute(query));
     }
 
     private ListMultimap<String, Version> versions(List<Clip> clips) throws GlycerinException {
@@ -187,7 +208,7 @@ public class GlycerinNitroClipsAdapter {
         if (clipPart.isEmpty()) {
             return ImmutableListMultimap.of();
         }
-        
+
         AvailabilityQuery query = AvailabilityQuery.builder()
                 .withDescendantsOf(toPid(clipPart))
                 .withPageSize(pageSize)
@@ -200,7 +221,17 @@ public class GlycerinNitroClipsAdapter {
             }
         });
     }
-    
+
+    private List<Availability> getNitroAvailabilities(Clip clip) throws GlycerinException {
+        AvailabilityQuery query = AvailabilityQuery.builder()
+                .withDescendantsOf(clip.getPid())
+                .withPageSize(pageSize)
+                .build();
+
+        GlycerinResponse<Availability> availabilities = glycerin.execute(query);
+        return availabilities.getResults();
+    }
+
     private Iterable<String> toPid(List<Clip> clipPart) {
         return Lists.transform(clipPart, new Function<Clip, String>() {
             @Override
@@ -250,8 +281,8 @@ public class GlycerinNitroClipsAdapter {
                 exhaustingProgrammeCallable(query));
 
         try {
-            return ImmutableList.copyOf(future.get());
-        } catch (InterruptedException | ExecutionException e) {
+            return ImmutableList.copyOf(future.get(10L, TimeUnit.MINUTES));
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
             if (e.getCause() instanceof GlycerinException) {
                 throw (GlycerinException) e.getCause();
             }
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
@@ -14,6 +14,8 @@ import org.atlasapi.remotesite.bbc.BbcFeeds;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroClipExtractor;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroItemSource;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroUtil;
+
+import com.google.common.base.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -116,7 +118,32 @@ public class GlycerinNitroClipsAdapter {
         } catch (GlycerinException e) {
             throw new NitroException(NitroUtil.toPids(refs).toString(), e);
         }
-        
+    }
+
+    public List<org.atlasapi.media.entity.Clip> clipsFor(PidReference ref) throws NitroException {
+        try {
+            Iterable<com.metabroadcast.atlas.glycerin.model.Clip> nitroClips
+                    = Iterables.transform(Iterables.filter(getNitroClip(ref), isClip), toClip);
+
+            if (Iterables.isEmpty(nitroClips)) {
+                log.warn("No programmes found for clipRefs {}", ref, new Function<PidReference, String>() {
+
+                    @Override
+                    public String apply(@Nullable PidReference pidRef) {
+                        return pidRef.getPid();
+                    }
+                });
+            }
+
+            Iterable<List<Clip>> clipParts = Iterables.partition(nitroClips, BATCH_SIZE);
+            ImmutableList.Builder<org.atlasapi.media.entity.Clip> clips = ImmutableList.builder();
+            for (List<Clip> clipPart : clipParts) {
+                clips.addAll(extractClips(clipPart).values());
+            }
+            return clips.build();
+        } catch (GlycerinException e) {
+            throw new NitroException(ref.toString(), e);
+        }
     }
 
     private Multimap<String, org.atlasapi.media.entity.Clip> extractClips(List<Clip> clipPart) throws GlycerinException {
@@ -200,6 +227,31 @@ public class GlycerinNitroClipsAdapter {
         
         ListenableFuture<List<ImmutableList<Programme>>> all = Futures.allAsList(futures);
         
+        try {
+            return ImmutableList.copyOf(Iterables.concat(all.get()));
+        } catch (InterruptedException | ExecutionException e) {
+            if (e.getCause() instanceof GlycerinException) {
+                throw (GlycerinException) e.getCause();
+            }
+            throw Throwables.propagate(e);
+        }
+    }
+
+    private ImmutableList<Programme> getNitroClip(PidReference ref) throws GlycerinException {
+
+        List<ListenableFuture<ImmutableList<Programme>>> futures = Lists.newArrayList();
+
+        ProgrammesQuery query = ProgrammesQuery.builder()
+                .withEntityType(EntityTypeOption.CLIP)
+                .withChildrenOf(ref.toString())
+                .withMixins(IMAGES)
+                .withPageSize(pageSize)
+                .build();
+
+        futures.add(executor.submit(exhaustingProgrammeCallable(query)));
+
+        ListenableFuture<List<ImmutableList<Programme>>> all = Futures.allAsList(futures);
+
         try {
             return ImmutableList.copyOf(Iterables.concat(all.get()));
         } catch (InterruptedException | ExecutionException e) {
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
@@ -17,7 +17,6 @@ import org.atlasapi.remotesite.bbc.nitro.extract.NitroClipExtractor;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroItemSource;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroUtil;
 
-import com.google.common.base.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -125,7 +124,7 @@ public class GlycerinNitroClipsAdapter {
     public List<org.atlasapi.media.entity.Clip> clipsFor(PidReference ref) throws NitroException {
         try {
             Iterable<com.metabroadcast.atlas.glycerin.model.Clip> nitroClips = Iterables
-                    .transform(Iterables.filter(getNitroClip(ref), isClip), toClip);
+                    .transform(Iterables.filter(getNitroClips(ref), isClip), toClip);
 
             if (Iterables.isEmpty(nitroClips)) {
                 log.warn("No programmes found for clipRefs {}", ref, new Function<PidReference, String>() {
@@ -138,8 +137,7 @@ public class GlycerinNitroClipsAdapter {
             }
 
             ImmutableList.Builder<org.atlasapi.media.entity.Clip> extractedClips = ImmutableList.builder();
-            List<Clip> clips = Lists.newArrayList(nitroClips);
-            for (Clip clip : clips) {
+            for (Clip clip : nitroClips) {
                 extractedClips.add(extractClip(clip));
             }
             return extractedClips.build();
@@ -268,7 +266,7 @@ public class GlycerinNitroClipsAdapter {
         }
     }
 
-    private ImmutableList<Programme> getNitroClip(PidReference ref) throws GlycerinException {
+    private ImmutableList<Programme> getNitroClips(PidReference ref) throws GlycerinException {
 
         ProgrammesQuery query = ProgrammesQuery.builder()
                 .withEntityType(EntityTypeOption.CLIP)
@@ -310,6 +308,4 @@ public class GlycerinNitroClipsAdapter {
         }
         return programmes.build();
     }
-
-    
 }
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
@@ -239,8 +239,6 @@ public class GlycerinNitroClipsAdapter {
 
     private ImmutableList<Programme> getNitroClip(PidReference ref) throws GlycerinException {
 
-        List<ListenableFuture<ImmutableList<Programme>>> futures = Lists.newArrayList();
-
         ProgrammesQuery query = ProgrammesQuery.builder()
                 .withEntityType(EntityTypeOption.CLIP)
                 .withChildrenOf(ref.toString())
@@ -248,12 +246,11 @@ public class GlycerinNitroClipsAdapter {
                 .withPageSize(pageSize)
                 .build();
 
-        futures.add(executor.submit(exhaustingProgrammeCallable(query)));
-
-        ListenableFuture<List<ImmutableList<Programme>>> all = Futures.allAsList(futures);
+        ListenableFuture<ImmutableList<Programme>> future = executor.submit(
+                exhaustingProgrammeCallable(query));
 
         try {
-            return ImmutableList.copyOf(Iterables.concat(all.get()));
+            return ImmutableList.copyOf(future.get());
         } catch (InterruptedException | ExecutionException e) {
             if (e.getCause() instanceof GlycerinException) {
                 throw (GlycerinException) e.getCause();
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
@@ -2,9 +2,9 @@ package org.atlasapi.remotesite.bbc.nitro;
 
 import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 
+import com.google.common.base.Predicate;
 import org.atlasapi.media.entity.Brand;
 import org.atlasapi.media.entity.Clip;
 import org.atlasapi.media.entity.Item;
@@ -20,34 +20,24 @@ import com.metabroadcast.atlas.glycerin.Glycerin;
 import com.metabroadcast.atlas.glycerin.GlycerinException;
 import com.metabroadcast.atlas.glycerin.GlycerinResponse;
 import com.metabroadcast.atlas.glycerin.model.Availability;
-import com.metabroadcast.atlas.glycerin.model.Broadcast;
 import com.metabroadcast.atlas.glycerin.model.Episode;
 import com.metabroadcast.atlas.glycerin.model.PidReference;
 import com.metabroadcast.atlas.glycerin.model.Programme;
-import com.metabroadcast.atlas.glycerin.model.Version;
 import com.metabroadcast.atlas.glycerin.queries.AvailabilityQuery;
-import com.metabroadcast.atlas.glycerin.queries.BroadcastsQuery;
 import com.metabroadcast.atlas.glycerin.queries.ProgrammesQuery;
-import com.metabroadcast.atlas.glycerin.queries.VersionsQuery;
 import com.metabroadcast.common.time.Clock;
 
-import com.google.api.client.util.Lists;
 import com.google.common.base.Function;
-import com.google.common.base.Predicates;
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimap;
-import com.google.common.collect.Multimaps;
-import com.google.common.util.concurrent.Futures;
-import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -64,6 +54,21 @@ public class GlycerinNitroContentAdapter implements NitroContentAdapter {
     private static final int NITRO_BATCH_SIZE = 10;
 
     private static final Logger log = LoggerFactory.getLogger(GlycerinNitroContentAdapter.class);
+    private static final Function<Programme, Episode> TO_EPISODE = new Function<Programme, Episode>() {
+
+        @Nullable
+        @Override
+        public Episode apply(@Nullable Programme input) {
+            return input.getAsEpisode();
+        }
+    };
+    public static final Predicate<Programme> IS_EPISODE = new Predicate<Programme>() {
+
+        @Override
+        public boolean apply(Programme input) {
+            return input.isEpisode();
+        }
+    };
 
     private final Glycerin glycerin;
     private final GlycerinNitroClipsAdapter clipsAdapter;
@@ -98,8 +103,7 @@ public class GlycerinNitroContentAdapter implements NitroContentAdapter {
         }
         try {
             checkRefType(refs, "brand");
-            ImmutableList<Programme> programmes =
-                    fetchProgrammes(makeProgrammeQueries(refs));
+            Iterable<Programme> programmes = fetchProgrammes(makeProgrammeQueries(refs));
             Multimap<String, Clip> clips = clipsAdapter.clipsFor(refs);
             ImmutableSet.Builder<Brand> fetched = ImmutableSet.builder();
             for (Programme programme : programmes) {
@@ -128,8 +132,7 @@ public class GlycerinNitroContentAdapter implements NitroContentAdapter {
         }
         try {
             checkRefType(refs, "series");
-            ImmutableList<Programme> programmes =
-                    fetchProgrammes(makeProgrammeQueries(refs));
+            Iterable<Programme> programmes = fetchProgrammes(makeProgrammeQueries(refs));
             Multimap<String, Clip> clips = clipsAdapter.clipsFor(refs);
             ImmutableSet.Builder<Series> fetched = ImmutableSet.builder();
             for (Programme programme : programmes) {
@@ -162,15 +165,9 @@ public class GlycerinNitroContentAdapter implements NitroContentAdapter {
     }
 
     @Override
-    public ImmutableSet<Item> fetchEpisodes(ProgrammesQuery programmesQuery) throws NitroException {
+    public Iterable<Item> fetchEpisodes(ProgrammesQuery programmesQuery) throws NitroException {
         try {
-            ImmutableList<Programme> programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
-
-            if (programmes.isEmpty()) {
-                log.warn("No programmes found for queries {}", programmesQuery.toString());
-                return ImmutableSet.of();
-            }
-
+            Iterable<Programme> programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
             return fetchEpisodesFromProgrammes(programmes);
         } catch (GlycerinException e) {
             throw new NitroException(programmesQuery.toString(), e);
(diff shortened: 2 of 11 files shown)


src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -128,8 +127,7 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
         }
         try {
             checkRefType(refs, "series");
-            ImmutableList<Programme> programmes =
-                    fetchProgrammes(makeProgrammeQueries(refs));
+            GlycerinQueryExecutor programmes = fetchProgrammes(makeProgrammeQueries(refs));
```
emilssolmanis: The type definition should be `Iterable<Programme>` here.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -164,14 +162,14 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
     @Override
     public ImmutableSet<Item> fetchEpisodes(ProgrammesQuery programmesQuery) throws NitroException {
         try {
-            ImmutableList<Programme> programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
+            GlycerinQueryExecutor programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
```
emilssolmanis: Same, the type definition has to be `Iterable<Programme>` here.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -182,20 +180,21 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
             throws NitroException {
         try {
             Iterable<ProgrammesQuery> programmesQueries = makeProgrammeQueries(refs);
-            ImmutableList<Programme> programmes = fetchProgrammes(programmesQueries);
+            GlycerinQueryExecutor programmes = fetchProgrammes(programmesQueries);
```
emilssolmanis: Ditto.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -182,20 +180,21 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
 
-            if (programmes.isEmpty()) {
+            Iterator<Programme> programmesIterator = programmes.iterator();
+            if (programmesIterator.hasNext()) {
+                return fetchEpisodesFromProgrammes(programmes.iterator());
```
emilssolmanis: Why not just make `fetchEpisodesFromProgrammes` work fine with an empty iterable rather than doing all these hacks?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -182,20 +180,21 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
             }
-
-            return fetchEpisodesFromProgrammes(programmes);
         } catch (GlycerinException e) {
             throw new NitroException(refs.toString(), e);
         }
     }
 
-    private ImmutableSet<Item> fetchEpisodesFromProgrammes(ImmutableList<Programme> programmes)
+    private ImmutableSet<Item> fetchEpisodesFromProgrammes(Iterator<Programme> programmes)
```
emilssolmanis: Methods should take an `Iterable`, not an `Iterator`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -242,21 +241,20 @@ public PidReference apply(Episode input) {
...
+            if (programme.isEpisode()) {
+                episodes.add(programme.getAsEpisode());
+            }
+        }
+
+        return episodes.build();
     }
 
-    private ImmutableList<Programme> fetchProgrammes(Iterable<ProgrammesQuery> queries)
+    private ImmutableList<Programme> exhaustingFetchProgrammes(Iterable<ProgrammesQuery> queries)
```
emilssolmanis: The naming is off, something like `eagerFetchProgrammes` or `fetchAllProgrammes` would be more descriptive. `exhausingFetchProgrammes` just sounds like it's a bit tiring for the computer.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -276,6 +274,12 @@ public Episode apply(Programme input) {
         }
     }
 
+    private GlycerinQueryExecutor fetchProgrammes(Iterable<ProgrammesQuery> queries)
```
emilssolmanis: This should return an `Iterable`, not a concrete implementation.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -242,21 +241,20 @@ public PidReference apply(Episode input) {
         return sources.build();
     }
 
-    private ImmutableList<Episode> getAsEpisodes(ImmutableList<Programme> programmes) {
-        return ImmutableList.copyOf(Iterables.filter(Iterables.transform(programmes,
-                new Function<Programme, Episode>() {
+    private ImmutableList<Episode> getAsEpisodes(Iterator<Programme> programmes) {
```
emilssolmanis: Are we sure having this return a `List` rather than an `Iterable` doesn't defeat our purpose of making this whole thing lazy?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+import com.metabroadcast.atlas.glycerin.GlycerinException;
+import com.metabroadcast.atlas.glycerin.GlycerinResponse;
+import com.metabroadcast.atlas.glycerin.model.Programme;
+import com.metabroadcast.atlas.glycerin.queries.ProgrammesQuery;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+
+public class GlycerinQueryExecutor implements Iterable<Programme> {
```
emilssolmanis: The name has to change. `Executor` implies it's an `ExecutorService` or something similar. Maybe something like `PaginatedProgrammeRequest`?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private Glycerin client;
+    private Iterator<ProgrammesQuery> currentIterator;
```
emilssolmanis: We discussed this yesterday f2f, this needs to change, the name is very confusing.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+    private ProgrammeIterator iterator;
```
emilssolmanis: And same for this, needs a much clearer name.

emilssolmanis: For that matter, at least the client and the query iterable should also be final.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(GlycerinQueryExecutor.class);
```
emilssolmanis: Style guide, static finals go at the very top.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public Iterator<Programme> iterator() {
+        return iterator;
+    }
+
+    private static class ProgrammeIterator implements Iterator<Programme> {
+
+        private Glycerin client;
+        private Iterator<ProgrammesQuery> queryIterator;
```
emilssolmanis: At least the client and query iterator should be final.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+        private GlycerinResponse<Programme> programmeGlycerinResponse;
+
+        private Iterator<Programme> currentProgramme;
```
emilssolmanis: Naming, if this is a collection, it should be plural.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+        public ProgrammeIterator(Glycerin client, Iterator<ProgrammesQuery> queryIterator) {
+            this.client = client;
+            this.queryIterator = queryIterator;
+        }
+
+        @Override
+        public boolean hasNext() {
```
emilssolmanis: A JavaDoc describing how this works is definitely needed.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public Iterator<Programme> iterator() {
+        return iterator;
+    }
+
+    private static class ProgrammeIterator implements Iterator<Programme> {
+
+        private Glycerin client;
+        private Iterator<ProgrammesQuery> queryIterator;
+        private GlycerinResponse<Programme> programmeGlycerinResponse;
```
emilssolmanis: Naming. It's pretty obvious just by the type that this is both a `programme` and a `glycerin` response. Maybe `currentResponse`?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
emilssolmanis: Naming. Type info doesn't belong in the variable name, this should just be `queries`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+
+            if (queryIterator.hasNext()) {
+                try {
+                    this.programmeGlycerinResponse = client.execute(queryIterator.next());
+
+                    currentProgramme = programmeGlycerinResponse.getResults().iterator();
+                    if (currentProgramme.hasNext()) {
+                        return true;
+                    } else {
+                        if (queryIterator.hasNext()) {
```
emilssolmanis: This is wrong. You're immediatly going to the next query, but it could be the case that we just need to fetch the next page.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
```diff
@@ -0,0 +1,89 @@
...
+                                return true;
+                            } else {
+                                return false;
+                            }
+                        } else {
+                            return false;
+                        }
+                    }
+                } catch (GlycerinException e) {
+                    log.warn("Couldn't get programmes from Nitro - {}", e);
```
emilssolmanis: This is **very** bad, you're swallowing an exception here. This should definitely propagate up rather than just pretending you've exhausted the iterator.

src/test/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapterTest.java
```diff
@@ -1,5 +1,9 @@
 package org.atlasapi.remotesite.bbc.nitro;
```
emilssolmanis: You haven't added any new tests here, and you definitely need to. For example, you haven't tested that the pagination logic works as you'd expect it to, the only test in here immediatly does a `hasNext() -> false` and therefore you're only testing the case of 1 request + 1 single paged response.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinQueryExecutor.java
MartinsIrbe: That's why previously `if (currentProgramme.hasNext())` would return true. `currentProgramme` is a field, therefore `if (queryIterator.hasNex())` would be executed only once there is no next page for the current programme.

src/test/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapterTest.java
MartinsIrbe: I'm testing pagination logic in `PaginatedProgrammeRequestTest`

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -164,9 +162,9 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
     @Override
     public ImmutableSet<Item> fetchEpisodes(ProgrammesQuery programmesQuery) throws NitroException {
         try {
-            ImmutableList<Programme> programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
+            Iterable<Programme> programmes = fetchProgrammes(ImmutableList.of(programmesQuery));
 
-            if (programmes.isEmpty()) {
+            if (!programmes.iterator().hasNext()) {
```
emilssolmanis: We talked about this, this isn't a good way of doing this, if only because calling `.iterator()` might produce a heavy throwaway object. I.e., imagine you'd have done the 1st call to Nitro in the constructor. Just let `fetchEpisodesFromProgrammes` handle empty iterators properly and do the warning there if really necessary. We're fine with getting rid of this log message if it's hard to do.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -242,38 +234,23 @@ public PidReference apply(Episode input) {
         return sources.build();
     }
 
-    private ImmutableList<Episode> getAsEpisodes(ImmutableList<Programme> programmes) {
-        return ImmutableList.copyOf(Iterables.filter(Iterables.transform(programmes,
-                new Function<Programme, Episode>() {
+    private Iterable<Episode> getAsEpisodes(Iterator<Programme> programmes) {
```
emilssolmanis: This should take an `Iterable`, not an actual `Iterator`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -182,22 +180,16 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
             throws NitroException {
```
emilssolmanis: The primary purpose of this whole thing is to make these calls not load everything into memory. You've only done part of the work and made it somewhat lazy internally, but this public method that's actually used in the ingest still materialises everything fully. Check how the `GlycerinNitroContentAdapter` is actually used from `OffScheduleContentIngestTask`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,92 @@
...
+        private GlycerinResponse<Programme> currentResponse;
+        private Iterator<Programme> currentProgrammes;
+
+        public ProgrammeIterator(Glycerin client, Iterator<ProgrammesQuery> programmeQueries) {
+            this.client = client;
+            this.programmeQueries = programmeQueries;
+        }
+
+        /**
+         * If current programmes is not null (has been created) and current programmes has next Programme, return true.
```
emilssolmanis: You need to explain the why, not the how. Everyone can read the code, this should be much higher level.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,92 @@
...
+         * programmes.
+         * Else return false, in case if there is no more programme queries and current programmes.
+         * If the query execution fails, GlycerinException would be thrown.
+         * @return returns true if Iterator has more objects.
+         */
+        @Override
+        public boolean hasNext() {
+            if (currentProgrammes != null && currentProgrammes.hasNext()) {
+                return true;
+            } else if (programmeQueries.hasNext()) {
```
emilssolmanis: Can remove a level of nesting by just guarding, e.g.,
```
if (currentProgrammes != null && currentProgrammes.hasNext()) {
    return true;
}

if (!programmeQueries.hasNext()) {
    return false;
}
...
```

src/test/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequestTest.java
```diff
@@ -0,0 +1,98 @@
...
+                .withPid(pid)
+                .withMixins(TITLES, PEOPLE)
+                .withPageSize(pageSize)
+                .build();
+
+        this.paginatedProgrammeRequest = new PaginatedProgrammeRequest(client, ImmutableList.of(programmesQuery));
+    }
+
+    @Test
+    public void testIfIteratorIsReturned() throws GlycerinException {
```
emilssolmanis: The naming could be better. Usually test names should include expected behaviour as well, so, e.g., this would be something like `programmesRequestReturnsNonEmptyIterator`, but, that said, it's unclear what this test actually tries to test.

src/test/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequestTest.java
```diff
@@ -0,0 +1,98 @@
...
+import static com.metabroadcast.atlas.glycerin.queries.ProgrammesMixin.TITLES;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PaginatedProgrammeRequestTest {
+
+    private Glycerin client = mock(Glycerin.class);
```
emilssolmanis: We generally use `@Mock` for this. Check for uses of them across Atlas if you need an example.

src/test/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequestTest.java
emilssolmanis: Also note that as of JUnit 4 you don't need to (and shouldn't) name test case methods `test*`. The annotation already conveys that info.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Iterator;
+
+public class PaginatedProgrammeRequest implements Iterable<Programme> {
+
+    private final Iterator<ProgrammesQuery> programmeQueries;
+    private final ProgrammeIterator currentProgrammes;
+
+    public PaginatedProgrammeRequest(Glycerin client, Iterable<ProgrammesQuery> queries) throws GlycerinException {
+        this.programmeQueries = queries.iterator();
+        this.currentProgrammes = new ProgrammeIterator(client, programmeQueries);
```
emilssolmanis: This is also wrong. `iterator()` needs to be reusable, so you have to save these verbatim and create a new `ProgrammeIterator` every time in `.iterator()`, otherwise you'll get in serious trouble if the iterable gets reused somewhere by a client.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+         * This has been done to decrease the heap overhead when querying Nitro.
+         * @return returns true if the current Programme is present, else returns false.
+         */
+        @Override
+        public boolean hasNext() {
+            if (currentProgrammes != null && currentProgrammes.hasNext()) {
+                return true;
+            }
+
+            if (programmeQueries.hasNext()) {
```
emilssolmanis: As I said, you can get rid of this level of nesting with just a guard.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+                try {
+                    if (currentProgrammes == null) { // Getting the first page.
+                        currentResponse = client.execute(programmeQueries.next());
+                        currentProgrammes = currentResponse.getResults().iterator();
+                    } else if (currentResponse.hasNext()) { // Getting the next page.
+                        currentProgrammes = currentResponse.getNext().getResults().iterator();
+                    } else if (!currentResponse.hasNext()) { // Getting the next Programme.
```
emilssolmanis: The comment is off, you're getting the next response, not the next programme.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+                        currentResponse = client.execute(programmeQueries.next());
+                        currentProgrammes = currentResponse.getNext().getResults().iterator();
```
emilssolmanis: This is clearly wrong. `response.getNext()` gets the next page before you've even processed the one you have.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+                    return checkCurrentProgramme(); // Check if current Programme is present.
+                } catch (GlycerinException e) {
+                    throw Throwables.propagate(e);
+                }
+            } else {
+                return false;
+            }
+        }
+
+        private boolean checkCurrentProgramme() throws GlycerinException {
```
emilssolmanis: Naming, `check*` implies an assertion and / or a `void` method.

emilssolmanis:
.. and you still need to make the actual public method return an `Iterable` and use that in `OffScheduleIngestTask`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -4,6 +4,8 @@
 
 import java.util.Set;
 
+import com.google.common.base.Optional;
+import com.google.common.collect.*;
```
emilssolmanis: We don't use star imports.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -73,36 +70,38 @@ protected void runTask() {
...
+                if (possibleEpisodeId.isPresent() && possibleContainerId.isPresent()) {
+                    reportStatus("Locking item ID");
 
-            reportStatus("Locking item IDs");
+                    String episodeId = possibleEpisodeId.get();
+                    String containerId = possibleContainerId.get();
 
-            lock.lock(episodeIds);
-            lock.lock(containerIds);
+                    episodeIds.add(episodeId);
```
emilssolmanis: This is a bit silly, you're locking them all and then only unlocking at the very end in a giant batch. That's completely unnecessary and will lead to this still being slow, why not just unlock once you're done with the item?

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -73,36 +70,38 @@ protected void runTask() {
...
+                    ResolveOrFetchResult<Item> items = localOrRemoteFetcher.resolveItems(ImmutableList.of(fetchedItem));
 
-            Iterable<Series> series = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Series.class);
-            Iterable<Brand> brands = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Brand.class);
+                    ImmutableSet<Container> resolvedSeries = localOrRemoteFetcher.resolveOrFetchSeries(items.getAll());
+                    ImmutableSet<Container> resolvedBrands = localOrRemoteFetcher.resolveOrFetchBrand(items.getAll());
 
-            reportStatus("Writing items");
-            writeContent(items, series, brands);
+                    Iterable<Series> series = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Series.class);
```
emilssolmanis: Don't just blindly nest things. By pulling this logic inside that loop, this is now no more a collection, this should just be a single Series and Brand now, adjust the code accordingly.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -73,36 +70,38 @@ protected void runTask() {
...
+                    Iterable<Brand> brands = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Brand.class);
+
+                    reportStatus("Writing items");
```
emilssolmanis: The statuses need updating as well, you're not "writing items" any more. That used to be a step that wrote out a giant batch of items, that is no longer the case.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -162,118 +150,73 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
-        for (NitroItemSource<Episode> source : sources) {
-            Item item = itemExtractor.extract(source);
-            item.setClips(clips.get(item.getCanonicalUri()));
-            fetched.add(item);
-        }
-        return fetched.build();
+        Iterable<Episode> episodes = getAsEpisodes(programmes);
+        Iterable<NitroItemSource<Episode>> sources = toItemSources(episodes);
+
+        return new PaginatedEpisodes(sources, itemExtractor, clipsAdapter);
```
emilssolmanis: The name off, it's not `PaginatedEpisodes`, the logic inside it is really more like `LazyNitroEpisodeExtractor`.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -162,118 +150,73 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
-                        return null;
-                    }
-                }), Predicates.notNull()));
+    private Iterable<Episode> getAsEpisodes(Iterable<Programme> programmes) {
+        return Iterables.transform(Iterables.filter(programmes, IS_EPISODE()), TO_EPISODE());
     }
 
-    private ImmutableList<Programme> fetchProgrammes(Iterable<ProgrammesQuery> queries)
-            throws GlycerinException {
+    private Function<Programme, Episode> TO_EPISODE() {
```
emilssolmanis: Why are you returning a closure? This should just be a constant.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -162,118 +150,73 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
+            @Nullable
+            @Override
+            public Episode apply(@Nullable Programme input) {
+                return input.getAsEpisode();
+            }
+        };
+    }
 
-        List<ListenableFuture<ImmutableList<Programme>>> futures = Lists.newArrayList();
+    private Predicate<Programme> IS_EPISODE() {
```
emilssolmanis: Same as above, this should be a constant.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -2,12 +2,13 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import java.util.Set;
-
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableList;
```
emilssolmanis: The import order is definitely wrong. You need to import the newest IntelliJ settings.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -68,114 +65,112 @@ protected void runTask() {
...
 
-            reportStatus("Writing items");
-            writeContent(items, series, brands);
+                if (writeSuccessful) {
+                    written++;
+                } else {
+                    failed++;
+                }
+
+                lock.unlock(episodeId);
```
emilssolmanis: We talked about this, this needs to go somewhere where it'll guaranteedly unlock them if there's an exception.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -68,114 +65,112 @@ protected void runTask() {
...
+                lock.unlock(containerId);
+            }
+            reportStatus(String.format("Written %d items of which %d failed", written, failed));
```
emilssolmanis: This needs to go on the inside. This is used to report progress and it's the only thing we see on the scheduler page.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -68,114 +65,112 @@ protected void runTask() {
...
         } catch (NitroException e) {
             throw Throwables.propagate(e);
         } catch (InterruptedException e) {
```
emilssolmanis: Needs to be handled on the inside, per item.

src/main/java/org/atlasapi/remotesite/bbc/nitro/OffScheduleContentIngestTask.java
```diff
@@ -68,114 +65,112 @@ protected void runTask() {
...
+            contentWriter.createOrUpdate(item);
+            return true;
+        } catch (Exception e) {
+            log.error(item.getCanonicalUri(), e);
+            return false;
         }
     }
 
-    private Series getSeries(Item item, ImmutableMap<String, Series> seriesIndex) {
+    private Series getSeries(Item item, Series series) {
```
emilssolmanis: These privates are unused now, as far as I can tell?

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedEpisodes.java
```diff
@@ -0,0 +1,86 @@
...
+import com.metabroadcast.atlas.glycerin.model.Episode;
+import com.metabroadcast.atlas.glycerin.model.PidReference;
+import org.atlasapi.media.entity.Clip;
+import org.atlasapi.media.entity.Item;
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroEpisodeExtractor;
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroItemSource;
+
+import java.util.Iterator;
+
+public class PaginatedEpisodes implements Iterable<Item> {
```
emilssolmanis: Missing JavaDocs.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedEpisodes.java
```diff
@@ -0,0 +1,86 @@
...
+                                GlycerinNitroClipsAdapter clipsAdapter) {
+            this.episodes = episodes.iterator();
+            this.itemExtractor = itemExtractor;
+            this.clipsAdapter = clipsAdapter;
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (episodes.hasNext()) {
+                this.episode = episodes.next();
```
emilssolmanis: Again, we discussed this yesterday. You can't do this, `.hasNext()` has to be idempotent. Consecutive calls to `.hasNext()` can't advance the iterator unless `.next()` was called.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedEpisodes.java
```diff
@@ -0,0 +1,86 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        @Override
+        public Item next() {
+            Item item = itemExtractor.extract(episode);
+            Multimap<String, Clip> clips = getClips(
```
emilssolmanis: Same as with everything else, this doesn't need to be a map any more.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedNitroItemSources.java
```diff
@@ -0,0 +1,228 @@
...
+            this.episodes = episodes.iterator();
+            this.pageSize = pageSize;
+            this.executor = executor;
+            this.glycerin = glycerin;
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (episodes.hasNext()) {
+                episode = episodes.next();
```
emilssolmanis: Same as the other, this is definitely not OK.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedNitroItemSources.java
```diff
@@ -0,0 +1,228 @@
...
+                    versions
+            );
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
+
+        private List<Availability> availabilities(Episode episode) throws GlycerinException {
```
emilssolmanis: This can be massively simplified now that you only have one episode. This has a list of `ListenableFuture`s, which we don't need any more. As we agreed yesterday, this should probably just return a `ListenableFuture` since the availabilities, broadcasts and versions can all be fetched in parallel.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedProgrammeRequest.java
```diff
@@ -0,0 +1,90 @@
...
+                    currentResponse = client.execute(programmeQueries.next());
+                    currentProgrammes = currentResponse.getResults().iterator();
+                }
+                return hasNextProgramme(); // Check if current Programme is present.
+            } catch (GlycerinException e) {
+                throw Throwables.propagate(e);
+            }
+        }
+
+        private boolean hasNextProgramme() throws GlycerinException {
```
emilssolmanis: 1. it doesn't actually throw anything
2. why does this method even exist?

src/main/java/org/atlasapi/remotesite/bbc/nitro/PidUpdateController.java
```diff
@@ -64,7 +64,7 @@ public void updatePidFromNitro(HttpServletResponse response, @PathVariable("pid"
             return;
         }
 
-        if (items.isEmpty()) {
+        if (!items.iterator().hasNext()) {
```
emilssolmanis: Might be simpler if you actually do something like `Iterables.getOnly` and then check if the item exists, or just catch the exception and handle this there. Otherwise this is just awkward.

src/main/java/org/atlasapi/remotesite/bbc/nitro/v1/PaginatedProgrammeEpisodes.java
```diff
@@ -0,0 +1,52 @@
...
+        private Programme programme;
+
+        public PaginatedProgrammeIterator(Iterable<Programme> programmes) {
+            this.programmes = programmes.iterator();
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (programmes.hasNext()) {
+                programme = programmes.next();
```
emilssolmanis: Same issue as above.

emilssolmanis:
Back to you @MartinsIrbe

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
```diff
@@ -209,6 +236,31 @@ public String apply(Clip input) {
             throw Throwables.propagate(e);
         }
     }
+
+    private ImmutableList<Programme> getNitroClip(PidReference ref) throws GlycerinException {
+
+        List<ListenableFuture<ImmutableList<Programme>>> futures = Lists.newArrayList();
```
emilssolmanis: Why do you have a list of futures when you only ever add 1 future to it?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -2,9 +2,9 @@
 
 import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 
+import com.google.common.base.Predicate;
```
emilssolmanis: Wrong order of imports.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -64,6 +54,21 @@
     private static final int NITRO_BATCH_SIZE = 10;
 
     private static final Logger log = LoggerFactory.getLogger(GlycerinNitroContentAdapter.class);
+    private static final Function<Programme, Episode> TO_EPISODE = new Function<Programme, Episode>() {
+
+        @Nullable
```
emilssolmanis: This isn't actually `@Nullable` though.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroContentAdapter.java
```diff
@@ -162,118 +165,54 @@ private void checkRefType(Iterable<PidReference> refs, String type) {
...
-            sources.add(NitroItemSource.valueOf(
-                    episode,
-                    availabilities.get(episode.getPid()),
-                    broadcasts.get(episode.getPid()),
-                    versions.get(episode.getPid())));
-        }
-        return sources.build();
+        return new PaginatedNitroItemSources(
+                episodes,
+                NITRO_BATCH_SIZE,
```
emilssolmanis: This used to be an unused member inside it, is this actually used now or just left here?

src/main/java/org/atlasapi/remotesite/bbc/nitro/LazyNitroEpisodeExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+        public EpisodesIterator(Iterable<NitroItemSource<Episode>> episodes, NitroEpisodeExtractor itemExtractor,
+                                GlycerinNitroClipsAdapter clipsAdapter) {
+            this.episodes = episodes.iterator();
+            this.itemExtractor = itemExtractor;
+            this.clipsAdapter = clipsAdapter;
+        }
+
+        @Override
+        public boolean hasNext() {
+            if (episodes.hasNext()) {
```
emilssolmanis: Again, just return the boolean value.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
```diff
@@ -116,7 +118,32 @@ public String apply(@Nullable PidReference pidRef) {
...
+                log.warn("No programmes found for clipRefs {}", ref, new Function<PidReference, String>() {
+
+                    @Override
+                    public String apply(@Nullable PidReference pidRef) {
+                        return pidRef.getPid();
+                    }
+                });
+            }
+
+            Iterable<List<Clip>> clipParts = Iterables.partition(nitroClips, BATCH_SIZE);
```
emilssolmanis: Why do we need to partition these?

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
```diff
@@ -209,6 +236,28 @@ public String apply(Clip input) {
...
+                .withChildrenOf(ref.toString())
+                .withMixins(IMAGES)
+                .withPageSize(pageSize)
+                .build();
+
+        ListenableFuture<ImmutableList<Programme>> future = executor.submit(
+                exhaustingProgrammeCallable(query));
+
+        try {
+            return ImmutableList.copyOf(future.get());
```
emilssolmanis: Probably want to add a timeout to this, to avoid it blocking indefinitely.

src/main/java/org/atlasapi/remotesite/bbc/nitro/LazyNitroEpisodeExtractor.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.common.base.Throwables;
+import com.google.common.collect.Multimap;
+import com.metabroadcast.atlas.glycerin.model.Episode;
+import com.metabroadcast.atlas.glycerin.model.PidReference;
+import org.atlasapi.media.entity.Clip;
+import org.atlasapi.media.entity.Item;
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroEpisodeExtractor;
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroItemSource;
+
+import java.util.Iterator;
```
emilssolmanis: The imports are wrong.

src/main/java/org/atlasapi/remotesite/bbc/nitro/LazyNitroEpisodeExtractor.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return episodes.hasNext();
+        }
+
+        @Override
+        public Item next() {
+            this.episode = episodes.next();
```
emilssolmanis: This looks like a local, why is this a field?

src/main/java/org/atlasapi/remotesite/bbc/nitro/LazyNitroEpisodeExtractor.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        public List<Clip> getClips(String pid, String uri) {
+            PidReference pidReference = new PidReference();
+            pidReference.setPid(pid);
+            pidReference.setHref(uri);
+
+            List<Clip> clips;
+            try {
+                clips = clipsAdapter.clipsFor(pidReference);
```
emilssolmanis: Sould simplify to `try { return clipsAdapter.clipsFor(pidReference); ... `

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedNitroItemSources.java
```diff
@@ -0,0 +1,190 @@
+package org.atlasapi.remotesite.bbc.nitro;
+
+import com.google.api.client.util.Lists;
+import com.google.common.base.Throwables;
+import com.google.common.collect.*;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
```
emilssolmanis: Wrong imports.

src/main/java/org/atlasapi/remotesite/bbc/nitro/PaginatedNitroItemSources.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return episodes.hasNext();
+        }
+
+        @Override
+        public NitroItemSource<Episode> next() {
+            episode = episodes.next();
```
emilssolmanis: This is a local, shouldn't be a field.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
```diff
@@ -116,7 +120,32 @@ public String apply(@Nullable PidReference pidRef) {
...
+
+                    @Override
+                    public String apply(@Nullable PidReference pidRef) {
+                        return pidRef.getPid();
+                    }
+                });
+            }
+
+            ImmutableList.Builder<org.atlasapi.media.entity.Clip> extractedClips = ImmutableList.builder();
+            List<Clip> clips = Lists.newArrayList(nitroClips);
```
emilssolmanis: This list seems unnecessary.

src/main/java/org/atlasapi/remotesite/bbc/nitro/GlycerinNitroClipsAdapter.java
```diff
@@ -209,6 +267,28 @@ public String apply(Clip input) {
             throw Throwables.propagate(e);
         }
     }
+
+    private ImmutableList<Programme> getNitroClip(PidReference ref) throws GlycerinException {
```
emilssolmanis: Naming, if you're getting a list, it's probably plural, `Clips`

src/main/java/org/atlasapi/remotesite/bbc/nitro/LazyNitroEpisodeExtractor.java
```diff
@@ -0,0 +1,84 @@
...
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroEpisodeExtractor;
+import org.atlasapi.remotesite.bbc.nitro.extract.NitroItemSource;
+
+import com.metabroadcast.atlas.glycerin.model.Episode;
+import com.metabroadcast.atlas.glycerin.model.PidReference;
+
+import com.google.common.base.Throwables;
+
+/**
+ * Extracts Atlas items from given Nitro episodes to reduce the heap overhead.
```
emilssolmanis: This description doesn't make sense. Extracting atlas items from nitro doesn't by itself "reduce heap overhead". It's the paginating part that's important. You need to describe why this exists in the first place, as well as what this does.

emilssolmanis:
other than that, lgtm

PR #953, opened 2016-05-16
Handle availabilities and versions disappearing in Nitro.

emilssolmanis:
In theory, these things should never just magically vanish, they should only ever be revoked. In practice,
theory and practice aren't the same, and we need to gracefully cope with this case to then propagate these
changes to our other partners. This tries to fuzzy match versions and availabilities and revokes them
if it can't find them. This aims for recall over precision, so we *might* mistakenly revoke some
availabilities when we mismatch, but we should always catch any actual deletions.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/ContentMerger.java
+++ b/src/main/java/org/atlasapi/remotesite/ContentMerger.java
@@ -35,37 +35,20 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 public class ContentMerger {
 
-    interface TopicMergeStrategy {
-        Content mergeTopics(Content current, Content extracted);
-    }
-
-    interface VersionMergeStrategy {
-        Content mergeVersions(Content current, Content extracted);
-    }
-    
-    interface AliasMergeStrategy {
-        Item mergeAliases(Item current, Item extracted);
-    }
-
-    public static abstract class MergeStrategy {
-        private MergeStrategy(){}
-
-        public static final LeaveEverythingAlone KEEP = new LeaveEverythingAlone();
-        public static final ReplaceEverything REPLACE = new ReplaceEverything();
-        public static final StandardMerge MERGE = new StandardMerge();
-        public static final BbcNitroRevokingMerge NITRO_VERSIONS_REVOKE = new BbcNitroRevokingMerge();
-
-        public static ReplaceTopicsBasedOnEquivalence replaceTopicsBasedOn(final Equivalence<TopicRef> equivalence) {
-            Preconditions.checkNotNull(equivalence);
-            return new ReplaceTopicsBasedOnEquivalence(equivalence);
-        }
-    }
-
     private final VersionMergeStrategy versionMergeStrategy;
     private final TopicMergeStrategy topicsMergeStrategy;
     private final AliasMergeStrategy aliasMergeStrategy;
 
-    public ContentMerger(VersionMergeStrategy versionMergeStrategy, TopicMergeStrategy topicsMergeStrategy, 
+    public static Container asContainer(Identified identified) {
+        return castTo(identified, Container.class);
+    }
+
+    public static Item asItem(Identified identified) {
+        return castTo(identified, Item.class);
+    }
+
+    public ContentMerger(VersionMergeStrategy versionMergeStrategy,
+            TopicMergeStrategy topicsMergeStrategy,
             AliasMergeStrategy aliasMergeStrategy) {
         this.versionMergeStrategy = checkNotNull(versionMergeStrategy);
         this.topicsMergeStrategy = checkNotNull(topicsMergeStrategy);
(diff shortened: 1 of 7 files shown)


src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -280,6 +289,182 @@ public Content mergeTopics(Content current, Content extracted) {
...
+                            return input.getCanonicalUri();
+                        }
+                    }
+            );
+
+            Map<String, Version> mergedVersions = Maps.newHashMap();
+
+            for (Map.Entry<String, Version> currVersionEntry : currVersions.entrySet()) {
+                if (!newVersions.containsKey(currVersionEntry.getKey())) {
+                    // BBC deleted it without properly revoking, invalidate all locations
```
Dalamar42: This code isn't BBC specific so the comment should be more generic

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -280,6 +289,182 @@ public Content mergeTopics(Content current, Content extracted) {
...
+                        public String apply(@Nullable Version input) {
+                            return input.getCanonicalUri();
+                        }
+                    }
+            );
+
+            Map<String, Version> mergedVersions = Maps.newHashMap();
+
+            for (Map.Entry<String, Version> currVersionEntry : currVersions.entrySet()) {
+                if (!newVersions.containsKey(currVersionEntry.getKey())) {
```
Dalamar42: This would be a bit more readable and wouldn't need the comments if each branch was a separate private method

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -280,6 +289,182 @@ public Content mergeTopics(Content current, Content extracted) {
...
+        }
+
+        private Set<Encoding> mergeEncodings(Set<Encoding> old, Set<Encoding> ingested) {
+            HashSet<Encoding> result = Sets.newHashSet();
+
+            // blats out locations on all old encodings that are not present in the ingested set
+            for (Encoding oldEncoding : old) {
+                Optional<Encoding> ingestedEncoding = findEqualEncoding(oldEncoding, ingested);
+                if (!ingestedEncoding.isPresent()) {
+                    // encoding got nuked, revoke old one, add new one
```
Dalamar42: This is talking about adding a new encoding, but the code only sets the locations on the old one to unavailable

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -280,6 +289,182 @@ public Content mergeTopics(Content current, Content extracted) {
...
+        private boolean locationFuzzyEqual(Location needle, Location candidate) {
+            return Objects.equal(needle.getLastUpdated(), candidate.getLastUpdated()) &&
+                    Objects.equal(needle.getAvailable(), candidate.getAvailable()) &&
+                    Objects.equal(needle.getTransportType(), candidate.getTransportType()) &&
+                    Objects.equal(needle.getUri(), candidate.getUri()) &&
+                    policyFuzzyEquals(needle.getPolicy(), candidate.getPolicy());
+        }
+
+        private boolean policyFuzzyEquals(@Nullable Policy needle, @Nullable Policy candidate) {
+            if (needle == candidate) {
```
Dalamar42: You probably want to return false if `null == null` instead of true

src/test/java/org/atlasapi/remotesite/ContentMergerTest.java
```diff
@@ -332,6 +334,116 @@ public void testMergesContainers() {
         );
     }
 
+    @Test
+    public void revokesVersionDeletedInNitro() {
```
Dalamar42: This test is not Nitro specific

src/test/java/org/atlasapi/remotesite/ContentMergerTest.java
```diff
@@ -332,6 +334,116 @@ public void testMergesContainers() {
...
+            @Override
+            public boolean apply(Version input) {
+                return "http://example.org/2".equals(input.getCanonicalUri());
+            }
+        });
+        assertThat(merged2, is(notNullValue()));
+    }
+
+    @Test
+    public void revokesAvailabilityDeletedInNitro() {
```
Dalamar42: Ditto

src/test/java/org/atlasapi/remotesite/ContentMergerTest.java
Dalamar42: I would add a couple of tests to test the different branches of the merge, i.e. that non-revoked versions and locations are also handled properly

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -280,6 +289,182 @@ public Content mergeTopics(Content current, Content extracted) {
...
+
+        @Override
+        public Content mergeVersions(Content current, Content extracted) {
+            Map<String, Version> currVersions = Maps.uniqueIndex(
+                    current.getVersions(),
+                    new Function<Version, String>() {
+
+                        @Override
+                        public String apply(@Nullable Version input) {
+                            return input.getCanonicalUri();
```
Dalamar42: Are you sure about this? I don't think all versions have a non-null canonical URI

Dalamar42:
Are we using ingester locks to ensure there are no race conditions around the merge logic?

Dalamar42:
When you are done with this could you please do some aggressive reformatting of `ContentMerger.java` to conform to the code style? i.e. move interfaces and inner classes to the bottom and whatever else you can think of.

src/main/java/org/atlasapi/remotesite/ContentMerger.java
emilssolmanis: Yeah, "new" as in the ingested one. Will clarify.

PR #956, opened 2016-05-18
Feature/dont display revoked locations

emilssolmanis:
Review 2nd commit only. 1st is a bunch of formatting fixup.

### DIFF
--- a/src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
+++ b/src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
@@ -568,19 +568,26 @@ public class ItemModelSimplifier
             ApplicationConfiguration config) {
         DateTime now = new DateTime(DateTimeZones.UTC);
         for (Location location : encoding.getAvailableAt()) {
+            if (!location.getAvailable()) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
```diff
@@ -568,19 +568,26 @@ private void addTo(org.atlasapi.media.entity.simple.Item simpleItem, Version ver
...
-                    || available(location.getPolicy(), now)) {
+                    || withinDateWindow(location, now)) {
                 addTo(simpleItem, version, encoding, location, item, annotations, config);
             }
         }
     }
 
-    private boolean available(Policy policy, DateTime now) {
-        return policy.getAvailabilityStart() == null
+    private boolean withinDateWindow(Location location, DateTime now) {
```
Dalamar42: I think `withinAvailabilityWindow` is a bit clearer

emilssolmanis: @Dalamar42 otherwise lgtm..?

PR #959, opened 2016-05-19
Fetch items 12h in future for Nitro off-schedule ingest.

emilssolmanis:
Currently we fetch only available items, which by definition means we lag behind by some sampling
duration. Nitro has the option to query for content that will become available in the future,
so we should ingest content before it actually becomes available.


MartinsIrbe:
lgtm

Dalamar42:
:shipit:

Spartakovic:
lgtm

PR #969, opened 2016-05-27
Stop fetching broadcasts for on-schedule ingest items and reuse the o…

emilssolmanis:
…nes from the discovery call


Dalamar42:
I don't love the null you are passing around, but I don't see you having much choice except by passing an `Optional<ImmutableListMultiMap<Broadcast> killMe`

LGTM

MartinsIrbe:
lgtm

emilssolmanis:
@Dalamar42 yeah, normally it'd just be an empty map, but that doesn't quite work here.

PR #971, opened 2016-05-27
Use available_versions mixin instead of separate calls to /versions a…

emilssolmanis:
…nd /availabilities in Nitro


MartinsIrbe:
LGTM

PR #974, opened 2016-06-02
Implement a /system/alive endpoint that always returns 200

Dalamar42:
- This is the minimum viable test to assert the jetty is up and
  serving HTTP requests


JamieAP:
LGTM

PR #975, opened 2016-06-02
Reject big schedule requests unless from privileged key

Dalamar42:


src/main/java/org/atlasapi/query/v2/ScheduleController.java
```diff
@@ -207,4 +222,28 @@ private boolean allNullOrEmpty(String... params) {
...
+        // These API keys are allowed to make big schedule requests
+        if (PRIVILEGED_KEYS.contains(apiKey)) {
+            return;
+        }
+
+        Interval interval = new Interval(fromWhen, toWhen);
+        if (interval.toDuration().isLongerThan(MAX_SCHEDULE_REQUEST_DURATION)) {
+            throw new IllegalArgumentException(MessageFormat.format(
+                    "Schedule requests for more than {} are not allowed",
+                    PeriodFormat.getDefault().print(MAX_SCHEDULE_REQUEST_DURATION.toPeriod())
```
emilssolmanis: Might make the format a [static] instance field to avoid the repeated calls.

emilssolmanis:
:shipit:

src/main/java/org/atlasapi/query/v2/ScheduleController.java
```diff
@@ -207,4 +224,28 @@ private boolean allNullOrEmpty(String... params) {
...
         return channels.build();
     }
+
+    private void checkRequestedInterval(
+            HttpServletRequest request,
+            DateTime fromWhen,
+            DateTime toWhen
+    ) {
+        String apiKey = request.getParameter(API_KEY_QUERY_PARAMETER);
+        if (apiKey == null) {
```
JamieAP: Dying to be extracted somewhere common

src/main/java/org/atlasapi/query/v2/ScheduleController.java
```diff
@@ -127,7 +144,7 @@ public void schedule(@RequestParam(required = false) String from,
                         "You must specify at least one publisher that you have permission to view");
             }
 
-            if (!(Strings.isNullOrEmpty(channelId) ^ Strings.isNullOrEmpty(channel))) {
```
JamieAP: Thank you

src/main/java/org/atlasapi/query/v2/ScheduleController.java
```diff
@@ -34,17 +35,31 @@
...
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static org.atlasapi.application.query.IpCheckingApiKeyConfigurationFetcher.API_KEY_QUERY_PARAMETER;
 
 @Controller
 public class ScheduleController extends BaseController<Iterable<ScheduleChannel>> {
 
     private static final Range<Integer> COUNT_RANGE = Range.closed(1, 10);
 
+    private static final ImmutableSet<String> PRIVILEGED_KEYS = ImmutableSet.<String>builder()
```
JamieAP: It would probably be nicer if `privileged` was a field on an Application, , so it could be configured via DB, maybe it can be in Deer.

This'll require a deploy every time we want to add a privileged key, although I guess that isn't often.

src/main/java/org/atlasapi/query/v2/ScheduleController.java
Dalamar42: It already is in `atlas-applications`. That's where I'm getting the `API_KEY_QUERY_PARAMETER` from. To expose the full apiKey fetching logic I would have to change the config fetcher, its interface and the BaseController in Owl. It felt excessive in the context of this.

PR #976, opened 2016-06-03
Reject big schedule requests unless from privileged key

Dalamar42:


atish-metabroadcast:
I hate that we're using properties here rather than Mongo.  OLGTM

Dalamar42:
Yeah I agree, but this is too temporary a solution to warrant messing with the applications collection

PR #977, opened 2016-06-03
Run YV schedule equivalence jobs for +8 days in the future

Dalamar42:
- Those YV jobs run at 2130 in the evening so to ensure we have at
  least +7 day coverage from the next day onwards we will run them
  for +8 days


atish-metabroadcast:
LGTM, although hygiene would have been nice as a separate commit

PR #985, opened 2016-06-09
Stop ingesting Nitro versions with null durations

emilssolmanis:
This is bad data from Nitro and we don't want to ingest it. It'll usually get corrected by the BBC,
at which point the next ingest can pick it up, but having no duration is not valid, so we don't
want it in Atlas.


Dalamar42:
LGTM

PR #986, opened 2016-06-09
Get the BBC channel list to ingest from Mongo rather than hard-coding it

emilssolmanis:


MartinsIrbe:
LGTM

src/main/java/org/atlasapi/remotesite/bbc/nitro/BbcNitroModule.java
```diff
@@ -211,20 +209,9 @@ GlycerinNitroChannelAdapter nitroChannelAdapter(Glycerin glycerin) {
...
-                    new Function<String, Channel>() {
-                        @Override
-                        public Channel apply(String input) {
-                            return channelResolver.fromUri(input).requireValue();
-                        }
-                    }
-                ));
-            }
-        };
+        return () -> ImmutableSet.copyOf(channelResolver.allChannels(ChannelQuery.builder()
```
Dalamar42: This is an expensive mongo call. Depending on how often this is happening you might want to cache the result for a little bit

PR #987, opened 2016-06-10
Create filter that adds a request's processing time as a response header

Dalamar42:
- This is intended to give us a measure of which requests are
  intrinsically expensive by measuring the time it takes to process
  them and ignoring any time waiting for a request thread to become
  available


JamieAP:
LGTM

PR #990, opened 2016-06-13
Add support to query channels by their type

emilssolmanis:
This means
 * stop ingesting media type for masterbrands, since that field is meaningless in that context
 * expose the channelType field for querying via API


Dalamar42:
LGTM

PR #992, opened 2016-06-14
MBST-15461: Content unpublish endpoint:

atish-metabroadcast:
To use:
  curl -i -XPOST 'http://localhost:8080/system/content/unpublish/{:id}'

Note:
This endpoint does not recursively unpublish versions.   TBD if required


Dalamar42:
LGTM

PR #996, opened 2016-06-20
MBST-15183: Channel4: remove legacy ingester.  Now use PMLSD throughout

atish-metabroadcast:
Monster refactor with the delete key and moving to using C4 PMLSD.

* Note equiv usages where I've switched from C4 to C4PMLSD
* YouView app mappings have changed too  (C4PMLSD to 4OD)


.gitignore
```diff
@@ -1,7 +1,7 @@
 .classpath
 .project
 .idea/
-.iml
+*.iml
```
Dalamar42: This shouldn't be in the project .gitignore, but only in our individual global .gitignore. Same goes for the eclipse entries

Dalamar42:
LGTM

PR #998, opened 2016-06-21
MBST-15461: Unpublisher endpoint: mandate the user specifies a publisher

atish-metabroadcast:
to prevent 'friendly fire'


Dalamar42:
LGTM though it would be a good idea to create a publish endpoint as well so we can correct friendly fire if need be :)

PR #999, opened 2016-06-21
MBST-15461: Unpublish endpoint: ID codec configuration fix

atish-metabroadcast:
SubstitutionTableNumberCodec by default uses the upper case alphabet


Spartakovic:
why it should be lowercase? I thought we are using it only in Deer

atish-metabroadcast:
Content IDs are shared between Deer and Owl.  If incorrect codec is used then the wrong numerical ID is hit from conversion.

Spartakovic:
oh right, my bad, lgtm

PR #1012, opened 2016-06-24
Adding channel POST endpoint.

MartinsIrbe:
+  Implemented channel translator from simple model to complex model.

### DIFF
--- a/src/main/java/org/atlasapi/input/ChannelModelTransformer.java
+++ b/src/main/java/org/atlasapi/input/ChannelModelTransformer.java
@@ -1,10 +1,12 @@
 package org.atlasapi.input;
 
 import java.math.BigInteger;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.atlasapi.media.channel.ChannelNumbering;
+import org.atlasapi.media.channel.ChannelType;
 import org.atlasapi.media.entity.Alias;
 import org.atlasapi.media.entity.Image;
 import org.atlasapi.media.entity.MediaType;
@@ -16,23 +18,32 @@ import org.atlasapi.media.entity.simple.PublisherDetails;
 import com.metabroadcast.common.base.Maybe;
 import com.metabroadcast.common.ids.NumberToShortStringCodec;
 
+import com.google.api.client.util.Lists;
 import com.google.common.base.Optional;
-import com.google.common.collect.Iterables;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
 import org.joda.time.LocalDate;
+
+import static com.google.gdata.util.common.base.Preconditions.checkNotNull;
 
 public class ChannelModelTransformer implements ModelTransformer<Channel, org.atlasapi.media.channel.Channel> {
 
     private final NumberToShortStringCodec v4Codec;
     private final ImageModelTranslator imageTranslator;
 
-    public ChannelModelTransformer(
+    private ChannelModelTransformer(
             NumberToShortStringCodec v4Codec,
             ImageModelTranslator imageTranslator
     ) {
-        this.v4Codec = v4Codec;
-        this.imageTranslator = imageTranslator;
+        this.v4Codec = checkNotNull(v4Codec);
+        this.imageTranslator = checkNotNull(imageTranslator);
+    }
+
+    public static ChannelModelTransformer create(
+            NumberToShortStringCodec v4Codec,
+            ImageModelTranslator imageTranslator
+    ) {
+        return new ChannelModelTransformer(v4Codec, imageTranslator);
     }
 
     @Override
@@ -42,20 +53,43 @@ public class ChannelModelTransformer implements ModelTransformer<Channel, org.at
         complex.withUri(simple.getUri());
         complex.withHighDefinition(simple.getHighDefinition());
         complex.withRegional(simple.getRegional());
+        complex.withRegion(simple.getRegion());
         complex.withAdult(simple.getAdult());
         complex.withTitle(simple.getTitle());
-
-        Set<Alias> aliases = simple.getV4Aliases().stream().map(simpleAlias -> new Alias(
-                simpleAlias.getNamespace(),
-                simpleAlias.getValue()
-        )).collect(Collectors.toSet());
-        complex.withAliases(aliases);
+        complex.withChannelType(ChannelType.valueOf(simple.getChannelType().toUpperCase()));
 
         if (simple.getMediaType() != null) {
-            Optional<MediaType> mediaType = MediaType.fromKey(simple.getMediaType().toUpperCase());
+            Optional<MediaType> mediaType = MediaType.fromKey(simple.getMediaType());
             if (mediaType.isPresent()) {
                 complex.withMediaType(mediaType.get());
+            } else {
+                throw new IllegalArgumentException("Couldn't get media type from simple model, "
+                        + "media type either not AUDIO or VIDEO.");
             }
+        }
+
+        if (simple.getPublisherDetails() != null) {
+            Maybe<Publisher> publisher = Publisher.fromKey(simple.getPublisherDetails().getKey());
+            if (!publisher.isNothing()) {
+                complex.withSource(publisher.requireValue());
+            } else {
+                throw new IllegalArgumentException("Publisher is not present while trying to set "
+                        + "source.");
+            }
+        }
+
+        if (simple.getBroadcaster() != null) {
+            Maybe<Publisher> publisher = Publisher.fromKey(simple.getBroadcaster().getKey());
+            if (!publisher.isNothing()) {
+                complex.withBroadcaster(publisher.requireValue());
+            } else {
+                throw new IllegalArgumentException("Couldn't extract publisher from simple "
+                        + "model broadcaster field.");
+            }
+        }
+
+        if (!simple.getTargetRegions().isEmpty()) {
+            complex.withTargetRegions(simple.getTargetRegions());
         }
 
         if (simple.getStartDate() != null) {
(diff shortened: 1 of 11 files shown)


src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+import org.joda.time.DateTime;
+import org.joda.time.Duration;
+import org.joda.time.LocalDate;
+
+public class ChannelModelTransformer implements ModelTransformer<Channel, org.atlasapi.media.channel.Channel> {
+
+    private final NumberToShortStringCodec v4Codec;
+    private final ImageModelTranslator imageTranslator;
+
+    public ChannelModelTransformer(
```
Dalamar42: Code style. Constructor must be private. Make a static factory method named `create(...)` to create a new transformer

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    public org.atlasapi.media.channel.Channel transform(Channel simple) {
+        org.atlasapi.media.channel.Channel.Builder complex = org.atlasapi.media.channel.Channel.builder();
+
+        complex.withUri(simple.getUri());
+        complex.withHighDefinition(simple.getHighDefinition());
+        complex.withRegional(simple.getRegional());
+        complex.withAdult(simple.getAdult());
+        complex.withTitle(simple.getTitle());
+
+        Set<Alias> aliases = simple.getV4Aliases().stream().map(simpleAlias -> new Alias(
```
Dalamar42: Break the method chain with newlines for readability

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+                simpleAlias.getNamespace(),
+                simpleAlias.getValue()
+        )).collect(Collectors.toSet());
+        complex.withAliases(aliases);
+
+        if (simple.getMediaType() != null) {
+            Optional<MediaType> mediaType = MediaType.fromKey(simple.getMediaType().toUpperCase());
```
Dalamar42: Don't silently swallow the failure if `MediaType.fromKey()` returns absent. That means the simple model contains an invalid media type string. Throw a descriptive exception and fail the call.

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+            complex.withAdvertiseFrom(new DateTime(simple.getAdvertisedFrom()));
+        }
+
+        if (simple.getGenres() != null && !simple.getGenres().isEmpty()) {
+            complex.withGenres(simple.getGenres());
+        }
+
+        if (simple.getPublisherDetails() != null) {
+            Maybe<Publisher> publisher = Publisher.fromKey(simple.getPublisherDetails().getKey());
+            if (!publisher.isNothing()) {
```
Dalamar42: We absolutely do not allow data to be written without a source. Same as above. If empty throw descriptive exception and fail the call

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+                complex.withSource(publisher.requireValue());
+            }
+        }
+
+        if (simple.getBroadcaster() != null) {
+            Maybe<Publisher> publisher = Publisher.fromKey(simple.getBroadcaster().getKey());
+            if (!publisher.isNothing()) {
```
Dalamar42: Ditto as above

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+                complex.withBroadcaster(publisher.requireValue());
+            }
+        }
+
+        if (simple.getImages() != null && !simple.getImages().isEmpty()) {
+            complex.withImages(Iterables.transform(
```
Dalamar42: Atlas Owl is now Java 8. `Iterables.transform(...)` should only be used if you need lazy evaluation which you don't here. Use Java 8 streams instead.

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+                    simple.getImages(),
+                    input -> imageTranslator.transform(input)
+            ));
+        }
+
+        if (simple.getAvailableFrom() != null && !simple.getAvailableFrom().isEmpty()) {
+            complex.withAvailableFrom(Iterables.transform(
```
Dalamar42: Ditto

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+
+        if (simple.getParent() != null) {
+            org.atlasapi.media.channel.Channel.Builder parent
+                    = org.atlasapi.media.channel.Channel.builder();
+            parent.withUri(simple.getParent().getUri());
+            complex.withParent(parent.build());
+        }
+
+        if (simple.getVariations() != null && !simple.getVariations().isEmpty()) {
+            complex.withVariations(Iterables.transform(
```
Dalamar42: Ditto

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+                    simple.getVariations(),
+                    channel -> transform(channel)
+            ));
+        }
+
+        simple.getRelatedLinks().stream().forEach(relatedLink -> {
```
Dalamar42: Newlines to break up method chain

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+            complex.withRelatedLink(translateRelatedLink(relatedLink));
+        });
+
+        if (simple.getChannelGroups() != null && !simple.getChannelGroups().isEmpty()) {
+            complex.withChannelNumbers(Iterables.transform(
```
Dalamar42: Java 8 stream

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+        complex.withImage(simple.getImage());
+        complex.withShortName(simple.getShortName());
+        complex.withSourceId(simple.getSourceId());
+        complex.withThumbnail(simple.getThumbnail());
+
+        return complex.build();
+    }
+
+    private Publisher translatePublisherDetails(PublisherDetails simple) {
+        Maybe<Publisher> complex = Publisher.fromKey(simple.getKey());
```
Dalamar42: Same as above. If you don't have a valid publisher throw exception

src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -365,6 +371,18 @@ private ContentWriteController contentWriteController() {
         );
     }
 
+    private ChannelWriteController channelWriteController() {
+        return new ChannelWriteController(
```
Dalamar42: Code style. Use static factory methods or builders for all classes. Constructors should be private

src/main/java/org/atlasapi/query/v2/ChannelWriteController.java
```diff
@@ -0,0 +1,120 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ChannelWriteController {
+
+    private final Logger LOG = LoggerFactory.getLogger(ChannelWriteController.class);
```
Dalamar42: Code style. Lowercase `log`

src/main/java/org/atlasapi/query/v2/ChannelWriteController.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private final ApplicationConfigurationFetcher appConfigFetcher;
+    private final ChannelStore store;
+    private final ModelReader reader;
+    private final ChannelModelTransformer channelTransformer;
+
+    public ChannelWriteController(
```
Dalamar42: Constructor should be private. Make a static factory method

src/main/java/org/atlasapi/query/v2/ChannelWriteController.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ChannelWriteController {
```
Dalamar42: Missing `@Controller` annotation

src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -365,6 +371,18 @@ private ContentWriteController contentWriteController() {
         );
     }
 
+    private ChannelWriteController channelWriteController() {
```
Dalamar42: This should public with `@Bean` annotation

src/main/java/org/atlasapi/query/v2/QueryController.java
```diff
@@ -210,4 +213,9 @@ public Void postContent(HttpServletRequest req, HttpServletResponse resp) {
     public Void putContent(HttpServletRequest req, HttpServletResponse resp) {
         return contentWriteController.putContent(req, resp);
     }
+
+    @RequestMapping(value="/3.0/channels.json", method = RequestMethod.POST)
```
Dalamar42: Why is this here?
- First, this is the query controller for content. The POST endpoint for channels does not belong here
- Second, you have already defined this endpoint in its own controller where it should be.

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
MartinsIrbe: 👍

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
MartinsIrbe: 👍

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
MartinsIrbe: 👍

src/main/java/org/atlasapi/input/ChannelModelTransformer.java
MartinsIrbe: 👍

src/main/java/org/atlasapi/query/v2/ChannelWriteController.java
MartinsIrbe: @Dalamar42 `ContentWriteController` doesn't have one either. Not sure if we really need this annotation.

src/main/java/org/atlasapi/query/QueryWebModule.java
MartinsIrbe: @Dalamar42 `ContentWriteController` is private and doesn't have @Bean annotation.

src/main/java/org/atlasapi/query/v2/QueryController.java
MartinsIrbe: @Dalamar42 I took an example from ContentWriteController. :) I have removed this as you are right it doesn't make any sense to have the annotation here as the entry point is in `QueryController` 👍

src/main/java/org/atlasapi/query/QueryWebModule.java
Dalamar42: It looks like `ContentWriteController` is not declared properly and is intended defined inside the QueryController. This needs to be declared a bean and made public so it can be discovered by Spring

src/main/java/org/atlasapi/query/v2/ChannelWriteController.java
Dalamar42: As below `ContentWriteController` is defined in a weird way. You absolutely need the `@Controller` annotation to ensure Spring discovers this class as a controller

PR #1018, opened 2016-06-30
Channel post endpoint fix

MartinsIrbe:


src/main/java/org/atlasapi/query/v2/ChannelController.java
```diff
@@ -186,6 +219,11 @@ public void listChannel(HttpServletRequest request, HttpServletResponse response
         }
     }
 
+    @RequestMapping(value={"/3.0/channels.*", "/channels.*"}, method = RequestMethod.POST)
```
ljeyaratnam: You should keep this code in ChannelWriteController and pass an instance of this object in ChannelController. Check how ContentController and ContentWriteController are implemented

MartinsIrbe: @ljeyaratnam there is no `ContentController`

ljeyaratnam:
LGTM

PR #1022, opened 2016-07-12
Change Opta ingester to 2016 season

Dalamar42:
- Change Premier League, Scottish Premier League, Bundesliga to 2016
  season. Champions League, Europa League and FA Cup do not yet have
  2016 data available so they will be updated later


Spartakovic:
lgtm

PR #1023, opened 2016-07-12
Change Wikipedia API URL to HTTPS

Dalamar42:
- Wikipedia API has started rejecting HTTP calls and forcing use of
  HTTPS


emilssolmanis:
lgtm

PR #1033, opened 2016-07-25
Externalise to config the name of MPX feed for new content

Dalamar42:
- This is done because the new content feed for the Vol-E environment
  has a different name than one in the production environment that we
  had hardcoded before
- Also remove unused `newFeedContentMatchingPredicate` from BtVodUpdater


emilssolmanis:
LGTM technically, whether it's logically correct I've no idea.

PR #1051, opened 2016-07-29
Add key to Nitro channels

Dalamar42:
- Even though channel keys are deprecated they are still required as
  parts of the code assume a channel will always have a key and can
  NPE if it's missing


emilssolmanis:
lgtm

emilssolmanis:
Actually scratch that lgtm, what happens if the keys are not unique? I.e., we might have keys for legacy PA ingested BBC channels that clash with SIDs, in fact that's quite likely to happen.

PR #1058, opened 2016-08-15
When desserialing channel we set key from uri

Spartakovic:
because key is compulsory field in some part of the system

### DIFF
--- a/src/main/java/org/atlasapi/input/ChannelModelTransformer.java
+++ b/src/main/java/org/atlasapi/input/ChannelModelTransformer.java
@@ -24,6 +24,7 @@ import org.joda.time.DateTime;
 import org.joda.time.Duration;
 import org.joda.time.LocalDate;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/input/ChannelModelTransformer.java
```diff
@@ -50,7 +50,11 @@ public static ChannelModelTransformer create(
     public org.atlasapi.media.channel.Channel transform(Channel simple) {
         org.atlasapi.media.channel.Channel.Builder complex = org.atlasapi.media.channel.Channel.builder();
 
+        if (simple.getUri() == null) {
```
Dalamar42: Simpler:
```
checkArgument(simple.getUri() != null, "Channel uri should be provided");
```

Throws the same exception and saves you the boilerplate and nesting

Dalamar42:
Minor comment, OLGTM

PR #1067, opened 2016-08-17
Add a scorer for matching title and description

AdamDM:
- Created equivalence scorer for matching title and description
  keywords. This helps with some content that has a description but
  no title or vv.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/scorers/DescriptionTitleMatchingScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/DescriptionTitleMatchingScorer.java
@@ -1,0 +1,107 @@
+package org.atlasapi.equiv.scorers;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import com.google.common.collect.ImmutableSet;
+
+import org.atlasapi.equiv.results.description.ResultDescription;
+import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/equiv/scorers/DescriptionTitleMatchingScorer.java
```diff
@@ -0,0 +1,124 @@
...
+
+        for (Item suggestion : suggestions) {
+            equivalents.addEquivalent(suggestion, score(subject, suggestion, desc));
+        }
+
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item suggestion, ResultDescription desc) {
+        Score score = Score.NULL_SCORE;
```
Spartakovic: don't see why you need to instantiate it here, just inline it with expression below

src/main/java/org/atlasapi/equiv/scorers/DescriptionTitleMatchingScorer.java
```diff
@@ -0,0 +1,124 @@
...
+        List<String> subjectList = descriptionToProcessedList(subject.getDescription());
+
+        List candidateTitleList = titleToProcessedList(candidate.getTitle());
+        List subjectTitleList = titleToProcessedList(subject.getTitle());
+
+        subjectList.retainAll(candidateTitleList);
+        candidateList.retainAll(subjectTitleList);
+
+        // calibrate the division factor
+        return subjectList.size() > (candidateTitleList.size()/2) || candidateList.size() > (subjectTitleList.size()/2);
```
jamiesprax: line >100 chars

PR #1072, opened 2016-08-19
Fixing compilation errors caused by common ingest changes.

MartinsIrbe:


src/main/java/org/atlasapi/remotesite/knowledgemotion/KnowledgeMotionUpdater.java
```diff
@@ -102,12 +104,22 @@ private boolean writeRow(KnowledgeMotionDataRow row, ProcessingResult.Builder re
...
             log.debug("Failed to update row {}", row.getId(), e);
-            resultBuilder.error(row.getId(), "While merging content: " + e.getMessage());
+            resultBuilder.addEntity(Entity.failure()
+                    .withId(row.getId())
+                    .withError(String.format(
+                            "While merging content: %s \n%s",
+                            row.getId(),
+                            e.getMessage()
+                    ))
+                    .build());
```
dragospitica: you should set the other fields of the entity as well, otherwise LGTM!

PR #1076, opened 2016-08-23
Mbst 16442 atlas voila owl k8 stage

lhopki01:


Dalamar42:
I'm not terribly fond of how we keep the same defaults in multiple docker files, but I don't think we have a choice here.

Dalamar42:
LGTM

grrywlsn:
LGTM

PR #1077, opened 2016-08-25
Feature/equiv title description scorer

AdamDM:
added scorers for Levenshtein distance and description matching and implemented their use in EquivModule

### DIFF
--- a/src/main/java/org/atlasapi/equiv/EquivModule.java
+++ b/src/main/java/org/atlasapi/equiv/EquivModule.java
@@ -59,7 +59,9 @@ import org.atlasapi.equiv.results.scores.ScoreThreshold;
 import org.atlasapi.equiv.scorers.BroadcastAliasScorer;
 import org.atlasapi.equiv.scorers.ContainerHierarchyMatchingScorer;
 import org.atlasapi.equiv.scorers.CrewMemberScorer;
+import org.atlasapi.equiv.scorers.DescriptionMatchingScorer;
 import org.atlasapi.equiv.scorers.EquivalenceScorer;
+import org.atlasapi.equiv.scorers.LDistanceTitleSubsetBroadcastItemScorer;
 import org.atlasapi.equiv.scorers.SequenceContainerScorer;
 import org.atlasapi.equiv.scorers.SequenceItemScorer;
 import org.atlasapi.equiv.scorers.SeriesSequenceItemScorer;
@@ -314,7 +316,7 @@ public class EquivModule {
         ));
         
         EquivalenceUpdater<Item> standardItemUpdater = standardItemUpdater(MoreSets.add(acceptablePublishers, LOVEFILM), 
-                ImmutableSet.of(new TitleMatchingItemScorer(), new SequenceItemScorer(Score.ONE), new DescriptionTitleMatchingScorer())).build();
+                ImmutableSet.of(new TitleMatchingItemScorer(), new SequenceItemScorer(Score.ONE), new DescriptionTitleMatchingScorer(), DescriptionMatchingScorer.makeScorer())).build();
         EquivalenceUpdater<Container> topLevelContainerUpdater = topLevelContainerUpdater(MoreSets.add(acceptablePublishers, LOVEFILM));
 
         Set<Publisher> nonStandardPublishers = ImmutableSet.copyOf(Sets.union(
@@ -712,9 +714,10 @@ public class EquivModule {
         return standardItemUpdater(sources, ImmutableSet.of(
             new TitleMatchingItemScorer(), 
             new SequenceItemScorer(Score.ONE),
-            new TitleSubsetBroadcastItemScorer(contentResolver, titleMismatch, 80/*percent*/),
+            new LDistanceTitleSubsetBroadcastItemScorer(contentResolver, titleMismatch, 80/*percent*/),
             new BroadcastAliasScorer(Score.nullScore()),
-            new DescriptionTitleMatchingScorer()
+            new DescriptionTitleMatchingScorer(),
+                DescriptionMatchingScorer.makeScorer()
         ), filter).build();
     }
 
--- a/src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
@@ -1,0 +1,97 @@
+package org.atlasapi.equiv.scorers;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.atlasapi.equiv.results.description.ResultDescription;
+import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.media.entity.Item;
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+
+public class DescriptionMatchingScorer implements EquivalenceScorer<Item> {
+
+    public static final String NAME = "Description Matching";
+    private final Set<String> commonWords = ImmutableSet.of(
+            "the", "in", "a", "and", "&", "of", "to", "show"
+    );
+
+    private DescriptionMatchingScorer() {
+    }
+
+    public static DescriptionMatchingScorer makeScorer(){
+        return new DescriptionMatchingScorer();
+    }
+
+    @Override
+    public ScoredCandidates<Item> score(Item subject, Set<? extends Item> candidates,
+            ResultDescription desc) {
+        DefaultScoredCandidates.Builder<Item> equivalents = DefaultScoredCandidates.fromSource(NAME);
+
+        for (Item candidate : candidates) {
+            equivalents.addEquivalent(candidate, score(subject, candidate, desc));
+        }
+
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item candidate, ResultDescription desc) {
+        Score score = Score.nullScore();
+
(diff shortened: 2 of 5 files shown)


src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -314,7 +316,7 @@ public boolean apply(Publisher input) {
         ));
         
         EquivalenceUpdater<Item> standardItemUpdater = standardItemUpdater(MoreSets.add(acceptablePublishers, LOVEFILM), 
-                ImmutableSet.of(new TitleMatchingItemScorer(), new SequenceItemScorer(Score.ONE), new DescriptionTitleMatchingScorer())).build();
+                ImmutableSet.of(new TitleMatchingItemScorer(), new SequenceItemScorer(Score.ONE), new DescriptionTitleMatchingScorer(), DescriptionMatchingScorer.makeScorer())).build();
```
emilssolmanis: Line length, this is definitely past 100 chars. Throw some whitespace in there.

src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -712,9 +714,10 @@ private SourceSpecificEquivalenceUpdater roviUpdater(Publisher roviSource, Immut
         return standardItemUpdater(sources, ImmutableSet.of(
             new TitleMatchingItemScorer(), 
             new SequenceItemScorer(Score.ONE),
-            new TitleSubsetBroadcastItemScorer(contentResolver, titleMismatch, 80/*percent*/),
+            new LDistanceTitleSubsetBroadcastItemScorer(contentResolver, titleMismatch, 80/*percent*/),
             new BroadcastAliasScorer(Score.nullScore()),
-            new DescriptionTitleMatchingScorer()
+            new DescriptionTitleMatchingScorer(),
+                DescriptionMatchingScorer.makeScorer()
```
emilssolmanis: Weird indent

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+import org.atlasapi.media.entity.Item;
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+
+public class DescriptionMatchingScorer implements EquivalenceScorer<Item> {
+
+    public static final String NAME = "Description Matching";
+    private final Set<String> commonWords = ImmutableSet.of(
```
emilssolmanis: It if's a constant, make it static.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+        Score score = Score.nullScore();
+
+        score = score(subject, candidate);
+        desc.appendText("%s (%s) scored: %s", candidate.getTitle(), candidate.getCanonicalUri(), score);
+
+        return score;
+    }
+
+    private Score score(Item subject, Item candidate) {
+        return descriptionMatch(subject, candidate)? Score.ONE: Score.nullScore();
```
emilssolmanis: Whitespace is weird, add a space before `?` and `:`

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+        candidateList.retainAll(subjectList);
+
+        //calibrate here
+        return new Double(candidateList.size())/new Double(subjectList.size()) > 0.2;
+    }
+
+    private List<String> descriptionToProcessedList(String description) {
+        List<String> descriptionList = new LinkedList<>();
+        if (!Strings.isNullOrEmpty(description)) {
+            descriptionList = filterCommon(new HashSet<>(tokenize(description)));
```
emilssolmanis: This whole thing can be done much more readable and with less indirection if you just used streams.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+
+        return score;
+    }
+
+    private Score score(Item subject, Item candidate) {
+        return descriptionMatch(subject, candidate)? Score.ONE: Score.nullScore();
+    }
+
+    private boolean descriptionMatch(Item subject, Item candidate) {
+        List<String> candidateList = descriptionToProcessedList(candidate.getDescription());
```
emilssolmanis: You have `List`s here, yet below `filterCommon` actually converts them to `Set`s and then back to a list. At this point, these all only have unique items and should therefore be `Set`s to match the semantics. For a `List` I'd expect these to be the tokenized descriptions **with tokens in order of appearance**, maybe minus the common words.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+        List<String> subjectList = descriptionToProcessedList(subject.getDescription());
+
+        candidateList.retainAll(subjectList);
+
+        //calibrate here
+        return new Double(candidateList.size())/new Double(subjectList.size()) > 0.2;
+    }
+
+    private List<String> descriptionToProcessedList(String description) {
+        List<String> descriptionList = new LinkedList<>();
```
emilssolmanis: Also not much point using `LinkedList`s everywhere. Bigger mem overhead etc., and you don't need the unique properties they provide. Generally you'd only use them as an impl of the `Queue` iface.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        return descriptionList;
+    }
+
+    private List<String> filterCommon(Set<String> words) {
+        return new ArrayList(Sets.difference(words, commonWords));
+    }
+
+    private List<String> tokenize(String target) {
+        Pattern pattern = Pattern.compile("\\b([A-Z]\\w*)\\b");
```
emilssolmanis: Why do we only want to match capitalised words in a description matcher?

src/main/java/org/atlasapi/equiv/scorers/LDistanceTitleSubsetBroadcastItemScorer.java
```diff
@@ -0,0 +1,136 @@
...
+                candidateWords.toString());
+        double stringsTotalLength = (double) (
+                subjectWords.toString().length() + candidateWords.toString().length()
+        );
+        return 1 - (lDistance / stringsTotalLength) >= threshold;
+    }
+
+    private String sanitize(String title) {
+        return titleTransformer.expand(title)
+                .replaceAll("[^\\d\\w\\s]", "").toLowerCase();
```
emilssolmanis: Why? Needs a comment.

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private Score score(Item subject, Item candidate) {
+        DefaultDescription desc = new DefaultDescription();
+        ScoredCandidates<Item> scores = scorer.score(subject, ImmutableSet.of(candidate), desc);
+        return scores.candidates().get(candidate);
+    }
+
+    @Test
+    public void testMatch() {
```
emilssolmanis: Naming methods according to `^test*` was a JUnit 3 thing. That's not necessary in JUnit 4.

Also the names need to be much more descriptive. Any report primarily only shows the test class and method name, and it should be immediately obvious what broke.

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.atlasapi.equiv.results.description.DefaultDescription;
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.media.entity.Item;
+
+import com.google.common.collect.ImmutableSet;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
```
emilssolmanis: We don't do star imports.

src/test/java/org/atlasapi/equiv/scorers/LDistanceTitleSubsetBroadcastItemScorerTest.java
```diff
@@ -0,0 +1,94 @@
...
+    private final LDistanceTitleSubsetBroadcastItemScorer scorer
+            = new LDistanceTitleSubsetBroadcastItemScorer(resolver, Score.nullScore(), 80);
+
+    @Test
+    public void testMatches() {
+        assertEquals(Score.ONE, score(
+                itemWithTitle("The Ren & Stimpy Show"),
+                itemWithTitle("Ren and Stimpy!")
+        ));
+        assertEquals(Score.ONE, score(
```
emilssolmanis: These should all be separate test cases, not bundled together, since presumably they test different aspects of the logic.

src/test/java/org/atlasapi/equiv/scorers/LDistanceTitleSubsetBroadcastItemScorerTest.java
```diff
@@ -0,0 +1,94 @@
...
+        item1.addVersion(version);
+        assertEquals(Score.nullScore(), score(
+                item1,
+                item2
+        ));
+    }
+
+    @Test
+    public void testSymbols() {
+        assertEquals(Score.ONE, score(itemWithTitle("Sponge Bob Square Pants"), itemWithTitle("Sponge Bob Square Pants: Episode 1")));
```
emilssolmanis: Line length.

PR #1082, opened 2016-09-01
Add unpublish content by id/uri

jamiesprax:
If a DELETE request is accepted with a valid API key then the ID / URI parameter from the request is sed to resolve a piece of content.
If the API key has write permissions for that contents publisher then it will be flagged as unpublished.

A corresponding unit test has been added.
More parameters have been added to initialise the ContentWriteController as they are used for the above.
The endpoint has been added to the QueryController.


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -93,6 +120,115 @@ public WriteResponse putContent(HttpServletRequest req, HttpServletResponse resp
...
+    @Nullable
+    @RequestMapping(value = "/3.0/content.json", method = RequestMethod.DELETE)
+    public WriteResponse unpublishContent(HttpServletRequest req, HttpServletResponse resp) {
+        return setPublishStatus(req, resp, false);
+    }
+
+    @Nullable
+    private WriteResponse setPublishStatus(HttpServletRequest req, HttpServletResponse resp,
+            boolean publishStatus)
+    {
```
Dalamar42: Code style. The `)` and the `{` should be on the same line. If this line is too long you can do:
```
private WriteResponse setPublishStatus(
    HttpServletRequest req, 
    HttpServletResponse resp,
    boolean publishStatus
) {
```

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -93,6 +120,115 @@ public WriteResponse putContent(HttpServletRequest req, HttpServletResponse resp
...
+        }
+
+        // find some content from URI
+        @SuppressWarnings("deprecation")
+        Maybe<Identified> identified;
+
+        ResolvedContent resolvedContent =
+                contentResolver.findByCanonicalUris(Lists.newArrayList(contentUri));
+
+        if(!resolvedContent.isEmpty()) {
```
Dalamar42: Code style. Space after `if`

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -93,6 +120,115 @@ public WriteResponse putContent(HttpServletRequest req, HttpServletResponse resp
...
+            identified = resolvedContent.getFirstValue();
+        } else {
+            return error(req, resp, AtlasErrorSummary.forException(
+                    new NoSuchElementException("could not resolve content for URI")
+            ));
+        }
+
+        // If we didn't find one, return an error
+        if(!identified.hasValue()) {
```
Dalamar42: Code style. Space after `if`

src/test/java/org/atlasapi/query/v2/ContentWriteControllerTest.java
```diff
@@ -96,12 +121,55 @@ public void setUp() throws Exception {
         when(writeExecutor.parseInputStream(any(InputStream.class), anyBoolean())).thenReturn(inputContent);
         when(idGenerator.getId(any(Content.class))).thenReturn(contentId);
         when(response.getOutputStream()).thenReturn(outputStream);
+
+        when(lookupEntry.uri()).thenReturn(uri);
+        ArrayList<LookupEntry> entryList = new ArrayList<>();
```
Dalamar42: You can do:
`ImmutableList<LookupEntry> entryList = ImmutableList.of(lookupEntry);`

Dalamar42: or `Lists.newArrayList(lookupEntry)` on the rare occasion that you need it to be mutable

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -69,18 +85,29 @@
     private final MessageSender<ContentWriteMessage> messageSender;
     private final AtlasModelWriter<QueryResult<Identified, ? extends Identified>> outputWriter;
 
+    private final LookupEntryStore lookupEntryStore;
+    private final ContentResolver contentResolver;
+    private final ContentWriter contentWriter;
+
     public ContentWriteController(
```
Dalamar42: Given you are on this class, it might be worth changing the constructor to private and making a static factory method to bring this in line with the code style. Leaving the campsite a little cleaner than we found it every time and all that.

Dalamar42:
Some minor comments. OLGTM

PR #1083, opened 2016-09-01
Add equiv scorer for matching descriptions

AdamDM:
- New scorer matches description to description
- Helps matching content with missing fields
- Potentially useful for sport where titles often mismatch

### DIFF
--- a/src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
@@ -1,0 +1,102 @@
+package org.atlasapi.equiv.scorers;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import org.atlasapi.equiv.results.description.ResultDescription;
+import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.media.entity.Item;
+
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+
+public class DescriptionMatchingScorer implements EquivalenceScorer<Item> {
+
+    public static final String NAME = "Description Matching";
+    private static final Set<String> commonWords = ImmutableSet.of(
+            "the", "in", "a", "and", "&", "of", "to", "show"
+    );
+
+    // Proportion threshold of key words that need to match between the two descriptions
+    // to conclude there is a match
+    public static final double PROPORTION_CROSSOVER = 0.2;
+
+    // Capitalised word finding regex
+    public static final String REGEX = "\\b([A-Z]\\w*)\\b";
+
+    private DescriptionMatchingScorer() {
+    }
+
+    public static DescriptionMatchingScorer makeScorer(){
+        return new DescriptionMatchingScorer();
+    }
+
+    @Override
+    public ScoredCandidates<Item> score(Item subject, Set<? extends Item> candidates,
+            ResultDescription desc) {
+        DefaultScoredCandidates.Builder<Item> equivalents = DefaultScoredCandidates.fromSource(NAME);
+
+        for (Item candidate : candidates) {
+            equivalents.addEquivalent(candidate, score(subject, candidate, desc));
+        }
+
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item candidate, ResultDescription desc) {
+        Score score = Score.nullScore();
+
+        score = score(subject, candidate);
+        desc.appendText("%s (%s) scored: %s", candidate.getTitle(), candidate.getCanonicalUri(), score);
+
+        return score;
+    }
+
+    private Score score(Item subject, Item candidate) {
+        return descriptionMatch(subject, candidate) ? Score.ONE : Score.nullScore();
+    }
+
+    private boolean descriptionMatch(Item subject, Item candidate) {
+        Set<String> candidateList = descriptionToProcessedList(candidate.getDescription());
+        Set<String> subjectList = descriptionToProcessedList(subject.getDescription());
+
+        candidateList.retainAll(subjectList);
+
+        return new Double(candidateList.size())/new Double(subjectList.size()) > PROPORTION_CROSSOVER;
+    }
+
+    private Set<String> descriptionToProcessedList(String description) {
+        if (!Strings.isNullOrEmpty(description)) {
+            return tokenize(description).stream()
+            .filter(s -> !commonWords.contains(s))
+            .collect(Collectors.toSet());
+        } else {
+            return Sets.newHashSet();
+        }
+    }
+
+    private List<String> tokenize(String target) {
+        Pattern pattern = Pattern.compile(REGEX);
+        Matcher subjectMatcher = pattern.matcher(target);
+        List<String> targetList = new LinkedList<>();
+        while (subjectMatcher.find()) {
+            targetList.add(subjectMatcher.group(1));
+        }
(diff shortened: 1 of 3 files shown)


src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class DescriptionMatchingScorerTest {
+
+    private DescriptionMatchingScorer scorer;
+
+    public DescriptionMatchingScorerTest() {
+        this.scorer = DescriptionMatchingScorer.makeScorer();
```
jamiesprax: You should use the

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,96 @@
...
+    private Set<String> descriptionToProcessedList(String description) {
+        return !Strings.isNullOrEmpty(description) ?
+               tokenize(description).stream()
+                .filter(s -> !commonWords.contains(s))
+                .collect(Collectors.toSet()) :
+                Sets.newHashSet();
+    }
+
+    private List<String> tokenize(String target) {
+        Pattern pattern = Pattern.compile("\\b([A-Z]\\w*)\\b");
```
jamiesprax: Extract this string

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,96 @@
...
+        //calibrate here
+        return new Double(candidateList.size())/new Double(subjectList.size()) > PROPORTION_CROSSOVER;
+    }
+
+    private Set<String> descriptionToProcessedList(String description) {
+        return !Strings.isNullOrEmpty(description) ?
+               tokenize(description).stream()
+                .filter(s -> !commonWords.contains(s))
+                .collect(Collectors.toSet()) :
+                Sets.newHashSet();
```
jamiesprax: This is a little confusing to read as the y part of a `x ? y : z` is what usually happens when x expression is true, but you've negated the expression so now they're swapped. Without noticing the `!` it looks like y happens when `description` is null or empty

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,96 @@
...
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+
+public class DescriptionMatchingScorer implements EquivalenceScorer<Item> {
+
+    public static final String NAME = "Description Matching";
+    private static final Set<String> commonWords = ImmutableSet.of(
+            "the", "in", "a", "and", "&", "of", "to", "show"
+    );
+    //proportion crossover between key words in two descriptions threshold
```
Dalamar42: This is not very easy to follow. Maybe 

```
// Proportion threshold of key words that need to match between the two descriptions
// to conclude there is a match

File: src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java [hunk 5]
```diff
@@ -0,0 +1,96 @@
...
+    public static final double PROPORTION_CROSSOVER = 0.2;
+
+    private DescriptionMatchingScorer() {
+    }
+
+    public static DescriptionMatchingScorer makeScorer(){
```
Dalamar42: :+1:

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -0,0 +1,96 @@
...
+        return descriptionMatch(subject, candidate) ? Score.ONE : Score.nullScore();
+    }
+
+    private boolean descriptionMatch(Item subject, Item candidate) {
+        Set<String> candidateList = descriptionToProcessedList(candidate.getDescription());
+        Set<String> subjectList = descriptionToProcessedList(subject.getDescription());
+
+        candidateList.retainAll(subjectList);
+
+        //calibrate here
```
Dalamar42: This is not very clear. Is this a TODO to calibrate in the future or does this code do the calibration. Also instead of calibration I think you mean normalise

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
Dalamar42: This is hard to follow.

- As @jamiesprax in if statements and especially in ternary operators positive conditions are generally preferred over negative conditions as they are easier to read and comprehend.

- With long ternary operators the code style is:
```
condition
? if branch
: else branch
```
however...

- This is definitely too much code to use a ternary operator. A standard if-else would be much more readable here.

```
if (condition) {
    return one thing;
} else {
    return another thing;
}
```

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
Dalamar42: That ^ and give a helpful name of what it is this regex is supposed to match

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
Dalamar42: Tests are not initialised in their constructor. There shouldn't be a constructor at all here.
Instead:

- For things that need to happen once only you use `@BeforeClass`. This is rare and usually only used to do expensive work like initialise a database
- For things that need to happen before every test like creating this scorer you use `@Before`

There is a code style on how these methods need to be named. Check the Junit documentation.

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private Score score(Item subject, Item candidate) {
+        DefaultDescription desc = new DefaultDescription();
+        ScoredCandidates<Item> scores = scorer.score(subject, ImmutableSet.of(candidate), desc);
+        return scores.candidates().get(candidate);
+    }
+
+    @Test
+    public void testDescriptionsMatchWhenTheyShould() {
```
Dalamar42: don't prefix tests with `test`. That's the old Junit style from the time of Junit3 that discovered tests by looking for methods that started with `test`. It's not needed any more

Dalamar42: It would be good to name this more descriptively. When should the descriptions match? `Should` is not obvious enough for someone unfamiliar with the code

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+        Item subject = new Item();
+        Item candidate = new Item();
+        subject.setDescription("Football related content and Manchester United talk");
+        candidate.setDescription("Manchester united play Football");
+        assertEquals(Score.ONE, score(candidate, subject));
+        assertEquals(Score.ONE, score(subject, candidate));
+    }
+
+    @Test
+    public void testDescriptionsMismatchWhenTheyShould() {
```
Dalamar42: ditto

src/test/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorerTest.java
```diff
@@ -0,0 +1,57 @@
...
+        Item subject = new Item();
+        Item candidate = new Item();
+        subject.setDescription("Football related content and Manchester United talk");
+        candidate.setDescription("Something nothing Actually related");
+        assertEquals(Score.nullScore(), score(candidate, subject));
+        assertEquals(Score.nullScore(), score(subject, candidate));
+    }
+
+    @Test
+    public void testCapitalsAreNecessaryForMatching() {
```
Dalamar42: This is a good name. You just need to remove the `test` bit

PR #1086, opened 2016-09-05
When writing new content set actively published

Dalamar42:
- The write executor was not updating the actively published flag in
  the existing content which means if the existing content was
  unpublished it was impossible to re-publish it via the API


emilssolmanis:
:shipit:

PR #1088, opened 2016-09-08
Switch opta ingester to 2016 seasons for UCL/EL

Dalamar42:


emilssolmanis:
lgtm? I suppose? I've no idea what this actually does and I think you're the only person who's worked on this ingester. 😅

PR #1089, opened 2016-09-09
model changes for rt film feed - localized title

AdamDM:


src/main/java/org/atlasapi/input/DescribedModelTransformer.java
```diff
@@ -79,9 +84,24 @@ protected T setFields(T result, F inputContent) {
...
         result.setPresentationChannel(inputContent.getPresentationChannel());
 
+        result.setLocalizedTitles(transformLocalizedTitles(inputContent.getTitles()));
+
         return result;
     }
 
+    private Set<LocalizedTitle> transformLocalizedTitles(Set<org.atlasapi.media.entity.simple.LocalizedTitle> titles) {
+        Set<LocalizedTitle> newTitles = new HashSet<>();
+        for (org.atlasapi.media.entity.simple.LocalizedTitle title: titles) {
```
jakekeeys: Couldn't you use a stream rather than a for each loop

PR #1091, opened 2016-09-12
added peppa and pig to list of common words to filter out 

AdamDM:
in description title scorer

also changed the ordering for how common words are filtered out


jakekeeys:
lgtm

PR #1092, opened 2016-09-13
Add flags for disabling the 4 MPX ingesters

jamiesprax:
A corresponding change was made to the BtVodModule [here](https://gitlab.com/mbst/infra/merge_requests/184/diffs) to disable the current MPX ingesters running in the monolith as they have been split up outside of it.


src/main/java/org/atlasapi/remotesite/btvod/BtVodModule.java
```diff
@@ -155,6 +155,9 @@
     private String btPortalContentGroupsBaseUri;
 
     // Prod config
+    @Value("${bt.vod.mpx.prod.enabled}")
+    private boolean prodEnabled;
```
emilssolmanis: Keep the naming consistent. The rest of the flags are `btVodMpxProdFeedBaseUrl`, so make this one `btVodMpxProdEnabled`. Same for the others.

emilssolmanis:
Commit message. _the 4 ingesters_ says very little without digging into the content. It should stand independent of the diff.

emilssolmanis:
olgtm

PR #1093, opened 2016-09-13
changes to pom file to get rid of multiple logging implementations

AdamDM:


emilssolmanis:
Assuming that's the only one, lgtm

emilssolmanis:
Actually scratch that, the indentation is messed up somehow. Also if that's the only other impl, olgtm

PR #1094, opened 2016-09-16
changes to only allow multiple items of the same publisher equivalate…

AdamDM:
… to content if their broadcasts overlap

this checks through all the candidates and their versions and their broadcasts within versions and sees if any of the broadcasts are inclusive of another's broadcast. If so the target content is allowed to equivalate to multiple pieces of content with the same publisher.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
+++ b/src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
@@ -1,61 +1,55 @@
 package org.atlasapi.equiv.results;
 
-import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 import org.atlasapi.equiv.results.combining.ScoreCombiner;
 import org.atlasapi.equiv.results.description.ReadableDescription;
 import org.atlasapi.equiv.results.description.ResultDescription;
 import org.atlasapi.equiv.results.extractors.EquivalenceExtractor;
+import org.atlasapi.equiv.results.extractors.MultipleCandidateExtractor;
 import org.atlasapi.equiv.results.filters.EquivalenceFilter;
 import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
-import org.atlasapi.equiv.results.scores.Score;
 import org.atlasapi.equiv.results.scores.ScoredCandidate;
 import org.atlasapi.equiv.results.scores.ScoredCandidates;
-import org.atlasapi.media.entity.Brand;
-import org.atlasapi.media.entity.Broadcast;
 import org.atlasapi.media.entity.Content;
-import org.atlasapi.media.entity.Identified;
 import org.atlasapi.media.entity.Publisher;
-import org.atlasapi.media.entity.Series;
-import org.atlasapi.media.entity.Version;
-
-import com.metabroadcast.common.stream.MoreCollectors;
 
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMap.Builder;
 import com.google.common.collect.ImmutableMultimap;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.SortedSetMultimap;
 import com.google.common.collect.TreeMultimap;
-import org.joda.time.DateTime;
-import org.joda.time.Duration;
-import org.joda.time.Interval;
 
 public class DefaultEquivalenceResultBuilder<T extends Content> implements EquivalenceResultBuilder<T> {
-
-    public static <T extends Content> EquivalenceResultBuilder<T> create(ScoreCombiner<T> combiner, EquivalenceFilter<T> filter, EquivalenceExtractor<T> marker) {
-        return new DefaultEquivalenceResultBuilder<T>(combiner, filter, marker);
-    }
 
     private final ScoreCombiner<T> combiner;
     private final EquivalenceExtractor<T> extractor;
     private final EquivalenceFilter<T> filter;
-    private static final Duration BROADCAST_TIME_FLEXIBILITY = Duration.standardMinutes(5);
-    private static double PUBLISHER_MATCHING_EQUIV_THRESHOLD = 0.3;
 
-    public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFilter<T> filter, EquivalenceExtractor<T> extractor) {
+    private final MultipleCandidateExtractor<T> multipleCandidateExtractor;
+
+    public DefaultEquivalenceResultBuilder(
+            ScoreCombiner<T> combiner,
+            EquivalenceFilter<T> filter,
+            EquivalenceExtractor<T> extractor
+    ) {
         this.combiner = combiner;
         this.filter = filter;
         this.extractor = extractor;
+
+        this.multipleCandidateExtractor = MultipleCandidateExtractor.create();
+    }
+
+    public static <T extends Content> EquivalenceResultBuilder<T> create(
+            ScoreCombiner<T> combiner,
+            EquivalenceFilter<T> filter,
+            EquivalenceExtractor<T> marker
+    ) {
+        return new DefaultEquivalenceResultBuilder<T>(combiner, filter, marker);
     }
 
     @Override
@@ -87,30 +81,39 @@ public class DefaultEquivalenceResultBuilder<T extends Content> implements Equiv
         return filteredCandidates;
     }
 
-    private Multimap<Publisher, ScoredCandidate<T>> extract(T target, List<ScoredCandidate<T>> filteredCandidates, ResultDescription desc) {
+    private Multimap<Publisher, ScoredCandidate<T>> extract(
+            T target,
+            List<ScoredCandidate<T>> filteredCandidates,
+            ResultDescription desc
+    ) {
         desc.startStage("Extracting strong equivalents");
-        SortedSetMultimap<Publisher, ScoredCandidate<T>> publisherBins = publisherBin(filteredCandidates);
+        SortedSetMultimap<Publisher, ScoredCandidate<T>> publisherBins =
+                publisherBin(filteredCandidates);
         
-        ImmutableMultimap.Builder<Publisher, ScoredCandidate<T>> builder = ImmutableMultimap.builder();
+        ImmutableMultimap.Builder<Publisher, ScoredCandidate<T>> builder =
+                ImmutableMultimap.builder();
         
         for (Publisher publisher : publisherBins.keySet()) {
             desc.startStage(String.format("Publisher: %s", publisher));
             
-            ImmutableSortedSet<ScoredCandidate<T>> copyOfSorted = ImmutableSortedSet.copyOfSorted(publisherBins.get(publisher));
-            Optional<ScoredCandidate<T>> extracted = extractor.extract(copyOfSorted.asList().reverse(), target, desc);
-            Set<ScoredCandidate<T>> allowedCandidates = candidatesThatCanBeEquivalatedToSamePublisher(copyOfSorted, extracted, target);
+            ImmutableSortedSet<ScoredCandidate<T>> copyOfSorted =
+                    ImmutableSortedSet.copyOfSorted(publisherBins.get(publisher));
 
-            if(!extracted.isPresent()) {
-                // no candidates - do nothing
-            }
-            else if (isSeriesOrBrand(extracted.get().candidate())) {
-                builder.put(publisher, extracted.get());
-            } else if(allowedCandidates.size() > 0 && !isSeriesOrBrand(target)) {
-                for (ScoredCandidate<T> scoredCandidate : allowedCandidates) {
-                    builder.put(publisher, scoredCandidate);
-                }
+            Optional<Set<ScoredCandidate<T>>> multipleExtractedCandidates =
+                    multipleCandidateExtractor.extract(
+                            copyOfSorted.asList().reverse(), target
+                    );
+
+            if (multipleExtractedCandidates.isPresent()) {
+                builder.putAll(
+                        publisher,
+                        multipleExtractedCandidates.get()
+                );
             } else {
-                if(extracted.isPresent()) {
+                Optional<ScoredCandidate<T>> extracted = extractor.extract(
+                        copyOfSorted.asList().reverse(), target, desc
+                );
+                if (extracted.isPresent()) {
                     builder.put(publisher, extracted.get());
                 }
             }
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,8 +107,19 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
         return builder.build();
     }
 
+    private boolean isSeriesOrBrand(T target) {
+        if (target instanceof Brand || target instanceof Series) {
+            return true;
+        }
+        return false;
+    }
```
jamiesprax: you can just write
`return (target instanceof Brand || target instanceof Series)`

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,8 +107,19 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
         return builder.build();
     }
 
+    private boolean isSeriesOrBrand(T target) {
+        if (target instanceof Brand || target instanceof Series) {
```
emilssolmanis: just `return target instanceof Brand || target instanceof Series`. ;)

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
AdamDM: true..

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,8 +107,19 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
...
+            return true;
+        }
+        return false;
+    }
+
     private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidates) {
         ScoredCandidate<T> previous = null;
+
+        // copy of candidates necessary to iterate through twice to compare all with all
+        Set<ScoredCandidate<T>> candidatesTwo = ImmutableSet.copyOf(candidates);
```
emilssolmanis: 1. rubbish name.
2. why does this even exist?

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+            for (ScoredCandidate<T> candidateTwo : candidatesTwo) {
+                if (broadcastsMatchCheck(candidate, candidateTwo)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private boolean broadcastsMatchCheck(ScoredCandidate<T> candidate, ScoredCandidate<T> candidateTwo) {
```
emilssolmanis: Typically, those arguments would be called something like `lhs, rhs`, `a, b` or similar, i.e., equal in "importance". Having one be `candidate` implies no other candidate exists.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+    }
+
+    private boolean broadcastsMatchCheckCrossover(Broadcast broadcast, Broadcast broadcastTwo) {
+        DateTime broadcastTransmissionStart = broadcast.getTransmissionTime();
+        DateTime broadcastTransmissionEnd = broadcast.getTransmissionEndTime();
+        DateTime broadcastTwoTransmissionStart = broadcastTwo.getTransmissionTime();
+        DateTime broadcastTwoTransmissionEnd = broadcastTwo.getTransmissionEndTime();
+
+        if (broadcastTransmissionStart.isAfter(broadcastTwoTransmissionEnd) &&
+                broadcastTransmissionStart.isBefore(broadcastTwoTransmissionEnd)) {
```
jamiesprax: I think there is an error in this logic.
It basically reads `if (A > B && A < B) return true;`

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+                if (broadcastsMatchCheck(candidate, candidateTwo)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private boolean broadcastsMatchCheck(ScoredCandidate<T> candidate, ScoredCandidate<T> candidateTwo) {
+        for (Version version : candidate.candidate().getVersions()) {
```
emilssolmanis: All of these nested for loops can probably be rewritten via `flatMap`s and `anyMatch`es.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+            for (Broadcast broadcastTwo : version.getBroadcasts()) {
+                if (broadcastsMatchCheckInclusive(broadcast, broadcastTwo)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    private boolean broadcastsMatchCheckCrossover(Broadcast broadcast, Broadcast broadcastTwo) {
```
emilssolmanis: Same re arg names.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
AdamDM: This code was meant as optional in case we wanted that included - it's no longer needed and never actually got tested (which is why I didn't find the bug). Good spot though! It's just this method that isn't used

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+        DateTime broadcastTwoTransmissionStart = broadcastTwo.getTransmissionTime();
+        DateTime broadcastTwoTransmissionEnd = broadcastTwo.getTransmissionEndTime();
+
+        if (broadcastTransmissionStart.isAfter(broadcastTwoTransmissionStart) &&
+                broadcastTransmissionStart.isBefore(broadcastTwoTransmissionEnd) &&
+                broadcastTransmissionEnd.isAfter(broadcastTwoTransmissionStart) &&
+                broadcastTransmissionEnd.isBefore(broadcastTwoTransmissionEnd)) {
+            return true;
+        } else {
+            return false;
```
jamiesprax: same as above for this just `return (conditions...)`

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+        return false;
+    }
+
+    private boolean broadcastsMatchCheckCrossover(Broadcast broadcast, Broadcast broadcastTwo) {
+        DateTime broadcastTransmissionStart = broadcast.getTransmissionTime();
+        DateTime broadcastTransmissionEnd = broadcast.getTransmissionEndTime();
+        DateTime broadcastTwoTransmissionStart = broadcastTwo.getTransmissionTime();
+        DateTime broadcastTwoTransmissionEnd = broadcastTwo.getTransmissionEndTime();
+
+        if (broadcastTransmissionStart.isAfter(broadcastTwoTransmissionEnd) &&
```
emilssolmanis: This can all be collapsed down to `return condition`, potentially by extracting those 2 if conditions to variables.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -112,8 +129,68 @@ private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidat
...
+        }
+    }
+
+    private boolean broadcastsMatchCheckInclusive(Broadcast broadcast, Broadcast broadcastTwo) {
+        DateTime broadcastTransmissionStart = broadcast.getTransmissionTime();
+        DateTime broadcastTransmissionEnd = broadcast.getTransmissionEndTime();
+        DateTime broadcastTwoTransmissionStart = broadcastTwo.getTransmissionTime();
+        DateTime broadcastTwoTransmissionEnd = broadcastTwo.getTransmissionEndTime();
+
+        if (broadcastTransmissionStart.isAfter(broadcastTwoTransmissionStart) &&
```
emilssolmanis: Same, just return the boolean.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
AdamDM: 1. Yeah I know
2. I tried a for loop over the set with a for loop over the same set within it and it did strange things. Doing this fixed it. I need to iterate through the same set within itself because I want to compare every candidate with every other candidate

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,19 +112,64 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
         return builder.build();
     }
 
-    private boolean canBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidates) {
-        ScoredCandidate<T> previous = null;
+    private boolean isSeriesOrBrand(T target) {
+        return (target instanceof Brand || target instanceof Series);
+    }
+
+    private boolean isSeriesOrBrand(ScoredCandidate<T> target) {
```
Dalamar42: You don't need both of those methods. You can just keep `boolean isSeriesOrBrand(T target)` and then for the second one just call `isSeriesOrBrand(scoredCandidate.candidate())`

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,19 +112,64 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
...
+        return (target.candidate() instanceof Brand || target.candidate() instanceof Series);
+    }
+
+    private Set<ScoredCandidate<T>> candidatesThatCanBeEquivalatedToSamePublisher(Set<ScoredCandidate<T>> candidates) {
+        // copy of candidates necessary to iterate through twice to compare all with all
+        Set<ScoredCandidate<T>> candidatesCopy = ImmutableSet.copyOf(candidates);
```
Dalamar42: You don't need to copy the set. The input is immutable anyway, but even if it wasn't you are not mutating it in this method at all.

Doing `for(T entry : set)` creates an iterator from the `set`. Doing it twice, creates two iterators

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,19 +112,64 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
...
-            if(previous!= null) {
-                Score score = candidate.score();
-                Score previousScore = previous.score();
-                if (!(Math.abs(score.asDouble()-previousScore.asDouble()) < 0.3)) {
-                    return false;
+            for (ScoredCandidate<T> candidateTwo : candidatesCopy) {
+                if (!(candidateTwo == candidate) &&
+                        broadcastsMatchCheck(candidate, candidateTwo) &&
+                        Math.abs(candidate.score().asDouble()-candidateTwo.score().asDouble())
+                                < 0.3) {
```
Dalamar42: Magic value. Extract this as a constant with a descriptive name.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,19 +112,64 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
...
+                .anyMatch(b -> broadcastsMatchCheckInclusive(b, broadcast));
+    }
+
+    private boolean broadcastsMatchCheckInclusive(Broadcast broadcastOne, Broadcast broadcastTwo) {
+        DateTime broadcastTransmissionStart = broadcastOne.getTransmissionTime();
+        DateTime broadcastTransmissionEnd = broadcastOne.getTransmissionEndTime();
+        DateTime broadcastTwoTransmissionStart = broadcastTwo.getTransmissionTime();
+        DateTime broadcastTwoTransmissionEnd = broadcastTwo.getTransmissionEndTime();
+
+        return (broadcastTransmissionStart.isAfter(
```
Dalamar42: If I understand correctly what you are trying to do here then I think this is wrong. In the 2nd and 3rd clause you are comparing transmission start with transmission end and vv which I think is not what's needed here.

Are we trying to say that the broadcasts must start within 5mins of each other and end within 5mins of each other?

If so the correct condition would be to compute the duration between the two start times and the duration between the two end times and assert that the absolute value of those durations is not longer than the flexibility using http://joda-time.sourceforge.net/apidocs/org/joda/time/base/AbstractDuration.html#isLongerThan(org.joda.time.ReadableDuration)

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -28,7 +37,7 @@
     private final DefaultEquivalenceResultBuilder resultBuilder = new DefaultEquivalenceResultBuilder(combiner, filter, extractor);
 
     @Test
-    public void checkEquivalatesToSeveralPa() {
+    public void checkDoesNotEquivalateToSeveralPa() {
```
Dalamar42: This test name is a little misleading. The code does not check if the candidates are PA or not. It's checks if it's series or brand and if the broadcasts match. The test name should reflect whichever of these conditions is actually causing this test case not to match.

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -39,7 +48,275 @@ public void checkEquivalatesToSeveralPa() {
...
+        assertTrue(equivalenceResult.strongEquivalences().values().size() == 1);
+    }
+
+    @Test
+    public void checkWontEquivalateToBrands() {
+        Item item = new Item();
+        item.setPublisher(Publisher.ARQIVA);
+        item.setCanonicalUri("target");
+
+        Brand candidate1 = brandWithBroadcast("2500", "6500");
```
Dalamar42: The broadcast times are too far apart so these candidates may fail to match even if the code that checks if they are brands is not working. Make the broadcasts identical so that you are only checking the brand aspect of it.

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -39,7 +48,275 @@ public void checkEquivalatesToSeveralPa() {
...
+        assertTrue(equivalenceResult.strongEquivalences().values().size() == 1);
+    }
+
+    @Test
+    public void checkWontEquivalateToSeries() {
+        Item item = new Item();
+        item.setPublisher(Publisher.ARQIVA);
+        item.setCanonicalUri("target");
+
+        Series candidate1 = seriesWithBroadcast("2500", "6500");
```
Dalamar42: Ditto as above

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -39,7 +48,275 @@ public void checkEquivalatesToSeveralPa() {
...
+                "2500",
+                "9500",
+                "4400",
+                "6500",
+                "9600",
+                5.0,
+                4.8,
+                4.9
+        );
+        assertTrue(equivalenceResult.strongEquivalences().values().size() == 2);
```
Dalamar42: You need to check which two equivalences came back, not just that there is two of them.

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -39,7 +48,275 @@ public void checkEquivalatesToSeveralPa() {
...
+
+        EquivalenceResult equivalenceResult = resultBuilder.resultFor(
+                item,
+                equivalents,
+                new DefaultDescription()
+        );
+        return equivalenceResult;
+    }
+
+    private Item itemWithBroadcast(String startTime, String endTime) {
```
Dalamar42: Why are you passing strings here instead of `LocalTime`? It seems like unnecessary complexity

src/test/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilderTest.java
```diff
@@ -39,7 +48,275 @@ public void checkEquivalatesToSeveralPa() {
...
+                Integer.parseInt(endTime.substring(3, 4))
+        );
+
+        Broadcast broadcast1 = new Broadcast("", dateTime1, dateTime2);
+        version.setBroadcasts(ImmutableSet.of(broadcast1));
+        item.setVersions(ImmutableSet.of(version));
+        return item;
+    }
+
+    private Brand brandWithBroadcast(String startTime, String endTime) {
```
Dalamar42: Ditto strings

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -101,19 +112,64 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
...
-                Score score = candidate.score();
-                Score previousScore = previous.score();
-                if (!(Math.abs(score.asDouble()-previousScore.asDouble()) < 0.3)) {
-                    return false;
+            for (ScoredCandidate<T> candidateTwo : candidatesCopy) {
+                if (!(candidateTwo == candidate) &&
+                        broadcastsMatchCheck(candidate, candidateTwo) &&
+                        Math.abs(candidate.score().asDouble()-candidateTwo.score().asDouble())
+                                < 0.3) {
+                    if (!allowedCandidates.contains(candidate) && !isSeriesOrBrand(candidate)) {
```
Dalamar42: There seems to be a bug in the old model where versions were moved to `Content` from `Item`. That seems wrong. `Brand` and `Series` cannot have broadcasts. This can be simplified to filter out brands and series before you start the O(n^2) comparisons to improve performance and make it a little easier to read.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
AdamDM: I swear doing this fixed a strange bug I was having, but it appears to be gone now!

emilssolmanis:
Approved, make sure to rebase + squash to clean up the history.

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -37,6 +49,8 @@
     private final ScoreCombiner<T> combiner;
     private final EquivalenceExtractor<T> extractor;
     private final EquivalenceFilter<T> filter;
+    private static final Duration BROADCAST_TIME_FLEXIBILITY = Duration.standardMinutes(5);
+    private static double PUBLISHER_MATCHING_EQUIV_THRESHOLD = 0.3;
```
Dalamar42: `final`

src/main/java/org/atlasapi/equiv/results/DefaultEquivalenceResultBuilder.java
```diff
@@ -83,12 +97,19 @@ public DefaultEquivalenceResultBuilder(ScoreCombiner<T> combiner, EquivalenceFil
             desc.startStage(String.format("Publisher: %s", publisher));
             
             ImmutableSortedSet<ScoredCandidate<T>> copyOfSorted = ImmutableSortedSet.copyOfSorted(publisherBins.get(publisher));
-            if(canBeEquivalatedToSamePublisher(copyOfSorted)) {
-                for (ScoredCandidate<T> scoredCandidate : copyOfSorted) {
+            Optional<ScoredCandidate<T>> extracted = extractor.extract(copyOfSorted.asList().reverse(), target, desc);
+            Set<ScoredCandidate<T>> allowedCandidates = candidatesThatCanBeEquivalatedToSamePublisher(copyOfSorted, extracted, target);
+
+            if(!extracted.isPresent()) {
```
Dalamar42: It's simpler to do:
```
if (extracted.isPresent() && isSeriesOrBrand(extracted.get().candidate())) {
    builder.put(publisher, extracted.get());
}
```

Dalamar42: Also code style please. There should be a space between `if` and the `(`

PR #1095, opened 2016-09-20
Feature/filter stale channels by dates

emilssolmanis:


src/main/java/org/atlasapi/remotesite/bbc/nitro/channels/ChannelIngestTask.java
```diff
@@ -66,7 +66,21 @@ protected void runTask() {
...
-            writeAndMergeChannels(filteredServices);
+
+            ImmutableList.Builder<Channel> withUris = ImmutableList.builder();
+            for (Channel channel : filteredServices) {
+                if (channel.getCanonicalUri() != null) {
+                    withUris.add(channel);
+                } else {
+                    log.warn(
+                            "Got channel without URI; this generally means it has no DVB locator "
+                                    + "in Nitro and no hard-coded override. Find Romain. {}",
```
dvllio: Maybe use something like "If the channel is missing a DVB locator in Nitro and doesn't appear in the hardcoded list, ask the BBC for the corresponding DVB locator." instead of "Find Romain."

emilssolmanis: This was obviously just a test to see if anyone actually reviews things, you know I would *never* merge something like this.

Dalamar42:
LGTM, I think...

PR #1096, opened 2016-09-21
Bump Mongo driver to 3.3.0

emilssolmanis:


src/main/java/org/atlasapi/AtlasModule.java
```diff
@@ -104,11 +101,7 @@ private boolean isProcessing() {
 
             @Override
             public ServerAddress apply(String input) {
-                try {
-                    return new ServerAddress(input, 27017);
-                } catch (UnknownHostException e) {
-                    return null;
-                }
+                return new ServerAddress(input, 27017);
```
jakekeeys: maybe const for the port

jakekeeys:
olgtm

PR #1097, opened 2016-09-22
Added alias writing to locations with PID and CRID

jamiesprax:
This will replace the service for converting youview PIDs -> CRIDs
and CRIDs -> PIDs as now a locations.alias.namespace/value call
will be able to resolve the content by the PID/CRID on the content.


src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/AddLocationAliasesToVersionTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+import com.google.common.collect.ImmutableList;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+public class AddLocationAliasesToVersionTest {
```
Dalamar42: Please name test classes after the class the test rather than by what they do.
Format: `SomeClass`, `SomeClassTest`

This makes it easier to discover what tests, if any exist for a particular class, because the IDE `go to test` shortcut will find them.

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/AddLocationAliasesToVersionTest.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final String AUDIO_DESC_VERSION_PID = "p048k3cj";
+    private static final String AUDIO_DESC_TYPE_ID = "DubbedAudioDescribed";
+    private static final String ORIGINAL_VERSION_PID = "b07wshjw";
+    private static final String ORIGINAL_TYPE_ID = "Original";
+
+    private static final String PID_NAMESPACE = "gb:bbc:nitro:prod:version:pid";
+    private static final String CRID_NAMESPACE = "gb:yv:prod:version:crid";
+    private static final String CRID_VALUE = "crid://nitro.bbc.co.uk/iplayer/youview/b38aea5d5ab29c7330f6d05a90f997aa";
+
+    private static final long MINUTES_45 = 2700000L;
```
Dalamar42: You can write this as `45 * 60 * 1000L` to make it more obvious.

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/AddLocationAliasesToVersionTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private final NitroEpisodeExtractor episodeExtractor = new NitroEpisodeExtractor(new SystemClock(),
```
Dalamar42: Alternatively you can use the Mockito JUnit runner and create mocks with the `@Mock` annotation

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/AddLocationAliasesToVersionTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+        buildValidEpisodeAliasList();
+
+        episode = new Episode();
+        episode.setPid(EPISODE_PID);
+
+    }
+
+    @Test
+    public void testLocationAliasesAreProperlyHashedAndSetOnEpisode() throws DatatypeConfigurationException {
```
Dalamar42: By convention unit test methods throw Exception rather than any more specific ones.

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/AddLocationAliasesToVersionTest.java
jamiesprax: This test has now been moved into NitroEpisodeExtractorTest

PR #1102, opened 2016-10-05
Add support for broadcastAssertions request parameter when writing

Dalamar42:
- The broadcast assertions are effectively a promise made by the
  writer that if any broadcasts existed in the asserted
  channel/dateTime ranges for the written item then these broadcasts
  would have been provided in this write. Therefore Atlas can use
  these assertions to remove any existing broadcast in the given
  range and replace it with the ones provided
- The existing behaviour of removing all existing broadcasts and
  replacing them with the new ones is preserved if no broadcast
  assertions are provided


AdamDM:
LGTM

PR #1116, opened 2016-10-21
Fix bug where PA bootstrap would not re-ingest any files

jakekeeys:
Prior to this change a bootstrap would only process unprocessed files
rather than the desired behavior of of re processing the most recent.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/PaBaseProgrammeUpdater.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/PaBaseProgrammeUpdater.java
@@ -103,7 +103,7 @@ public abstract class PaBaseProgrammeUpdater extends ScheduledTask {
         executor.shutdown();
     }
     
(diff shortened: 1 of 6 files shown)


src/main/java/org/atlasapi/remotesite/pa/PaBaseProgrammeUpdater.java
```diff
@@ -166,7 +166,14 @@ private void processBatch(Iterable<File> files) throws JAXBException,
 
 		            AtomicInteger jobsRemaining = new AtomicInteger();
 
-		            unmarshaller.setListener(channelDataProcessingListener(jobsRemaining, completion, submitted, fileToProcess, scheduleDay));
+		            unmarshaller.setListener(channelDataProcessingListener(
+		                    jobsRemaining,
```
emilssolmanis: Weird whitespace around here.

PR #1117, opened 2016-10-21
added title parsing to the RTE ingester

AdamDM:


src/main/java/org/atlasapi/remotesite/rte/RteParser.java
```diff
@@ -26,5 +26,24 @@ private static String buildCanonicalUriFromId(String id) {
         }
         return CANONICAL_URI_PREFIX + matcher.group(1);
     }
+
+    public static String titleParser(String originalTitle) {
+        String title = originalTitle.substring(6, (originalTitle.length() - 7));
```
Dalamar42: Magic numbers:
```
private static final String TITLE_PREFIX = "Watch ";
private static final String TITLE_POSTFIX = " online";

checkArgument(originalTitle.startsWith(TITLE_PREFIX);
checkArgument(orginalTitle.endsWith(TITLE_POSTFIX);

String title = originalTitle.substring(
        TITLE_PREFIX.length(), 
        originalTitle.length() - TITLE_POSTFIX.length()
);
```

src/main/java/org/atlasapi/remotesite/rte/RteParser.java
```diff
@@ -26,5 +26,24 @@ private static String buildCanonicalUriFromId(String id) {
...
+        if (!title.contains("Season")) {
+            return title;
+        }
+
+        String[] titleArray = title.split(" ");
```
Dalamar42: You can instead do:
```
private static final TITLE_SEASON = " Season";

return title.split(TITLE_SEASON)[0];
```

src/test/java/org/atlasapi/remotesite/rte/RteParserTest.java
```diff
@@ -35,5 +35,28 @@ public void canonicalUriGenerationShouldFailIfUriHasMultipleIdParams() {
...
-    
+
+    @Test
+    public void titleParsingWatchOnlineStrippingTest() {
+        Assert.assertEquals("Fair City Extras: Carol and Robbie",
+                RteParser.titleParser("Watch Fair City Extras: Carol and Robbie online"));
+    }
+
+    @Test
+    public void titleParsingWatchOnlineStrippingTest2() {
```
Dalamar42: Don't do test 1 and test 2, etc. Only make multiple tests if you have multiple behaviours to test in which case you should make each test method name describe what the different expected behaviour is

src/test/java/org/atlasapi/remotesite/rte/RteParserTest.java
```diff
@@ -35,5 +35,28 @@ public void canonicalUriGenerationShouldFailIfUriHasMultipleIdParams() {
     public void canonicalUriGenerationShouldFailIfUriHasEmptyIdParam() {
         RteParser.canonicalUriFrom("http://feedurl.com/?id=");
     }
-    
+
+    @Test
+    public void titleParsingWatchOnlineStrippingTest() {
```
Dalamar42: Write test method names like this: `doingXDoesY`

`titleParsingRemovesWatchPrefixAndOnlinePostfix`

src/main/java/org/atlasapi/remotesite/rte/RteParser.java
Dalamar42: Also confirm that the prefix and postfix are always the same. It might be safer to do something like the following unless you are sure we will always have the same prefix and postfix.

```
if (we have title prefix) {
    remove it;
}
if (we have online postfix) {
    remove it;
}

File: src/main/java/org/atlasapi/remotesite/rte/RteParser.java [hunk 5]
```diff
@@ -7,11 +7,15 @@
...
 import com.google.common.base.Strings;
 
-
 public class RteParser {
     
     private final static String CANONICAL_URI_PREFIX = "http://rte.ie/shows/";
     private final static Pattern ID_PATTERN = Pattern.compile(".*:(\\d+)$");
+    private static final String TITLE_PREFIX = "Watch ";
+    private static final String TITLE_POSTFIX = " online";
+    private static final String TITLE_SEASON = " Season";
```
Dalamar42: I think those two fields are unused now

PR #1128, opened 2016-10-27
Fix nitro ingest failing when failing to resolve channel

jamiesprax:
Unresolved channels are now filtered out and logged as errrors
instead of the ingest completely failing to initialise.

The hardcoded list should never have anything that isnt present
in the db so unresolved channels are a potential issue. This shouldn't
be ignored but also should not stop the ingest from initialising at all.


src/main/java/org/atlasapi/remotesite/bbc/nitro/BbcNitroModule.java
```diff
@@ -220,10 +223,18 @@ GlycerinNitroChannelAdapter nitroChannelAdapter(Glycerin glycerin) {
...
     private Supplier<ImmutableSet<Channel>> bbcChannelSupplier() {
-        return () -> ImmutableSet.copyOf(BbcIonServices.services.values()
-                .stream()
-                .map(channelResolver::fromUri)
-                .map(Maybe::requireValue)
-                .collect(Collectors.toList()));
+
+        ImmutableSet.Builder<Channel> resolvedChannelBuilder = ImmutableSet.builder();
+
+        for(String uri : BbcIonServices.services.values()) {
```
emilssolmanis: Ugh. No. :( You could've just added the logging by expanding the `.map()` a little bit, like
```
.map(uri -> {
    Maybe<Channel> chan = channelResolved.fromUri(uri);
    if (!chan.hasValue()) {
        log.error("boo boo");
    }
    return chan;
})
.filter()
...
```

PR #1133, opened 2016-10-28
Mbst 17216

jakekeeys:


src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -562,32 +564,56 @@ private Item getFilmWithoutBroadcast(ProgData progData) {
...
-            if (previous instanceof Film) {
-                film = (Film) previous;
+        Optional<String> rtFilmIdentifier = rtFilmIdentifierFor(progData);
+        rtFilmIdentifier.transform(rt -> uris.add(PaHelper.getFilmRtAlias(rt)));
+
+        uris.add(PaHelper.getAlias(progData.getProgId()));
+        uris.add(PaHelper.getFilmUri(identifierFor(progData)));
+        Map<String, Identified> resolvedContent = contentResolver.findByUris(
+                uris.build()
+        ).asResolvedMap();
```
JamieAP: gross

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
```diff
@@ -295,8 +295,11 @@ public void testGetTitleAndDescriptionForNewFilm() throws Exception {
 
         when(contentResolver.findByCanonicalUris(ImmutableList.of(brandUri)))
                 .thenReturn(ResolvedContent.builder().build());
-        when(contentResolver.findByUris(ImmutableList.of("http://pressassociation.com/1", expectedUri)))
-                .thenReturn(ResolvedContent.builder().build());
+        when(contentResolver.findByUris(ImmutableList.of(
+                "http://pressassociation.com/films/5",
+                "http://pressassociation.com/1",
+                expectedUri)
+        )).thenReturn(ResolvedContent.builder().build());
```
JamieAP: gross

src/test/java/org/atlasapi/remotesite/pa/PaProgrammeProcessorTest.java
```diff
@@ -381,8 +392,11 @@ public void testCreateBroadcastFromProgData() throws Exception {
...
```
JamieAP: gross

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -562,32 +564,56 @@ private Item getFilmWithoutBroadcast(ProgData progData) {
...
-        )).getFirstValue();
+        ImmutableList.Builder<String> uris = ImmutableList.builder();
 
-        Film film;
-        if (possiblePreviousData.hasValue()) {
-            Identified previous = possiblePreviousData.requireValue();
-            if (previous instanceof Film) {
-                film = (Film) previous;
+        Optional<String> rtFilmIdentifier = rtFilmIdentifierFor(progData);
+        rtFilmIdentifier.transform(rt -> uris.add(PaHelper.getFilmRtAlias(rt)));
```
Dalamar42: `getFilmRtAlias`? You mean `getLegacyFilmUri` no?

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -562,32 +564,56 @@ private Item getFilmWithoutBroadcast(ProgData progData) {
...
+            ProgData progData,
+            Optional<String> rtFilmIdentifier,
+            Map<String, Identified> resolvedContent
+    ) {
+        if (resolvedContent.isEmpty()) {
+            return getBasicFilm(progData);
+        }
+
+        Identified previous;
+        if (rtFilmIdentifier.isPresent()) {
```
Dalamar42: This could use a comment explaining why we give precedence to films found with the legacy format

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -562,32 +564,56 @@ private Item getFilmWithoutBroadcast(ProgData progData) {
...
+            String legacyFilmUri = PaHelper.getFilmRtAlias(rtFilmIdentifier.get());
```
Dalamar42: You are calling `getFilmRtAlias` here again. This should be `getLegacyFilmUri` I think

Dalamar42: Also you shouldn't be computing the `legacyFilmUri` twice. Pass it in in this method as an optional. If it doesn't exist then you know not to search for it. Computing it twice runs the risk of the two computations being different due to a bug

Dalamar42: To reduce the nesting you can do:

```
if (legacyFilmUri.isPresent() && resolvedContent.containsKey(legacyFilmUri.get()) {
    previous = resolvedContent.get(legacyFilmUri.get());
} else {
    previous = Iterables.getFirst(resolvedContent.values(), null);
}
```

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
jakekeeys: getFilmRtAlias is correct and is the original v3 namespace Alias... i have removed the old v4 alias as they were never used

jakekeeys:
I updated the existing tests to check for both the old and new Alias's

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
Dalamar42:
```
Optional<String> legacyFilmUri = rtFilmIdentifier.map(rt -> PaHelper.getFilmRtAlias(rt));
legacyFilmUri.ifPresent(uris::add);
```

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -562,32 +563,56 @@ private Item getFilmWithoutBroadcast(ProgData progData) {
...
         return film;
     }
-    
+
+    private Film getFilmFromResolvedContent(
+            ProgData progData,
+            Map<String, Identified> resolvedContent,
+            Optional<String> legacyFilmUri) {
+        // This method gives precedence to films found with the old version 3 alias URI's
+        // We were previously creating duplicates by creating content with the new version 3
```
Dalamar42: It would be better concept to explain that this was introduced because we wanted to change the format of Film uris.

PR #1145, opened 2016-11-08
Code style; Break up business logic in setPublishStatusOfItem

jakekeeys:


src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -196,7 +196,7 @@
     private @Autowired TaskStore taskStore;
     private @Autowired ContentHierarchyExpander hierarchyExpander;
     private @Autowired ChannelStore channelStore;
-    private @Autowired EquivalenceBreaker equivalenceBreaker;
+    @Autowired private EquivalenceBreaker equivalenceBreaker;
```
jamiesprax: If you're going to change one, you might as well change the others as you go

src/main/java/org/atlasapi/system/UnpublishContentController.java
```diff
@@ -78,6 +78,22 @@ private void setPublishStatusOfItem(String id, Optional<String> publisher, boole
...
+        // check publisher constraint is met
+        Described described = validatePublisher(publisher, contentId, identified);
+
+        // remove from equivset if un-publishing
+        removeItemFromEquivSet(status, lookupEntry);
+
+        // change publish status
+        described.setActivelyPublished(status);
+
+        // now write in the appropriate manner
```
jamiesprax: These comments are quite verbose and implied by the method names anyway so in my opinion I'd say they aren't needed

PR #1148, opened 2016-11-10
add flag to reset the score on content with only null results for ebs

AdamDM:
flag can be set to change logic in combiner to set score to 0 if all scorers returned null
ebs sport content registered to use this flag

### DIFF
--- a/src/main/java/org/atlasapi/equiv/EquivModule.java
+++ b/src/main/java/org/atlasapi/equiv/EquivModule.java
@@ -284,24 +284,36 @@ public class EquivModule {
             )));
     }
 
-    private ContentEquivalenceUpdater.Builder<Item> ebsItemUpdater(Set<Publisher> acceptablePublishers, Set<? extends EquivalenceScorer<Item>> scorers) {
+    private ContentEquivalenceUpdater.Builder<Item> ebsItemUpdater(
+            Set<Publisher> acceptablePublishers,
+            Set<? extends EquivalenceScorer<Item>> scorers
+    ) {
 
-        NullScoreAwareAveragingCombiner<Item> combiner = new NullScoreAwareAveragingCombiner<>();
-        combiner.setIgnoreNullScoringContent(true);
+        NullScoreAwareAveragingCombiner<Item> combiner =
+                new NullScoreAwareAveragingCombiner<>(true);
 
         return ContentEquivalenceUpdater.<Item> builder()
                 .withGenerators(ImmutableSet.<EquivalenceGenerator<Item>> of(
-                        new BroadcastMatchingItemEquivalenceGenerator(scheduleResolver,
-                                channelResolver, acceptablePublishers, Duration.standardMinutes(5), Predicates.alwaysTrue())
+                        new BroadcastMatchingItemEquivalenceGenerator(
+                                scheduleResolver,
+                                channelResolver,
+                                acceptablePublishers,
+                                Duration.standardMinutes(5),
+                                Predicates.alwaysTrue())
                 ))
                 .withExcludedUris(excludedUrisFromProperties())
                 .withScorers(scorers)
                 .withCombiner(combiner)
                 .withFilter(this.standardFilter())
-                .withExtractor(PercentThresholdAboveNextBestMatchEquivalenceExtractor.atLeastNTimesGreater(1.5))
+                .withExtractor(
+                        PercentThresholdAboveNextBestMatchEquivalenceExtractor
+                                .atLeastNTimesGreater(1.5)
+                )
                 .withHandler(new BroadcastingEquivalenceResultHandler<Item>(ImmutableList.of(
                         EpisodeFilteringEquivalenceResultHandler.relaxed(
-                                new LookupWritingEquivalenceHandler<Item>(lookupWriter, acceptablePublishers),
+                                new LookupWritingEquivalenceHandler<Item>(
+                                        lookupWriter,
+                                        acceptablePublishers),
                                 equivSummaryStore
                         ),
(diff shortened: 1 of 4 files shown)


src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -282,6 +283,34 @@ public boolean apply(Broadcast input) {
...
+                                new LookupWritingEquivalenceHandler<Item>(lookupWriter, acceptablePublishers),
+                                equivSummaryStore
+                        ),
+                        new ResultWritingEquivalenceHandler<Item>(equivalenceResultStore()),
+                        new EquivalenceSummaryWritingHandler<Item>(equivSummaryStore),
+                        MessageQueueingResultHandler.create(
+                                equivAssertDestination(), acceptablePublishers, lookupEntryStore
+                        )
+                )));
+    }
```
jamiesprax: lots of lines greater >100 characters in this method

src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -282,6 +283,34 @@ public boolean apply(Broadcast input) {
                 )
             )));
     }
+
+    private ContentEquivalenceUpdater.Builder<Item> ebsItemUpdater(Set<Publisher> acceptablePublishers, Set<? extends EquivalenceScorer<Item>> scorers) {
+
+        NullScoreAwareAveragingCombiner<Item> combiner = new NullScoreAwareAveragingCombiner<>();
+        combiner.setIgnoreNullScoringContent(true);
```
Dalamar42: Do not make a setter for this. This flag should be passed in the constructor and be `final`.

src/main/java/org/atlasapi/equiv/EquivModule.java
Dalamar42: That

src/main/java/org/atlasapi/equiv/results/combining/NullScoreAwareAveragingCombiner.java
```diff
@@ -22,11 +22,19 @@
...
     
+    private boolean ignoreNullScoringContent;
+
     public static final <T extends Described> NullScoreAwareAveragingCombiner<T> get() {
         return new NullScoreAwareAveragingCombiner<T>();
     }
     
-    public NullScoreAwareAveragingCombiner() {}
+    public NullScoreAwareAveragingCombiner() {
+        ignoreNullScoringContent = false;
```
Dalamar42: `ignoreNullScoringCandidate`

src/main/java/org/atlasapi/equiv/results/combining/NullScoreAwareAveragingCombiner.java
```diff
@@ -78,6 +86,11 @@ private void addCount(final Map<T, Integer> counts, Entry<T, Score> equivScore)
             public Score transformEntry(T key, Score value) {
                 if (value.isRealScore()) {
                     Integer count = publisherCounts.get(key.getPublisher());
+
+                    if (ignoreNullScoringContent && count == 1) {
```
Dalamar42: Add javadoc explaining that if `count == 1` then only the only score we have must be from the generator and no scorers have returned a non-null score.

src/main/java/org/atlasapi/equiv/update/ContentEquivalenceUpdater.java
```diff
@@ -25,30 +25,83 @@
...
 public class ContentEquivalenceUpdater<T extends Content> implements EquivalenceUpdater<T> {
 
-    public static final <T extends Content> Builder<T> builder() {
+    public static <T extends Content> Builder<T> builder() {
         return new Builder<T>();
     }
     
-    public static final class Builder<T extends Content> {
+    private final ScoredEquivalentsMerger merger = new ScoredEquivalentsMerger();
+    private final Function<ScoredCandidates<T>, Iterable<T>> extractCandidates = new Function<ScoredCandidates<T>, Iterable<T>>() {
```
Dalamar42: Line is too long

src/main/java/org/atlasapi/equiv/update/ContentEquivalenceUpdater.java
```diff
@@ -25,30 +25,83 @@
 
 public class ContentEquivalenceUpdater<T extends Content> implements EquivalenceUpdater<T> {
 
-    public static final <T extends Content> Builder<T> builder() {
```
Dalamar42: I presume all the changes in this file are just methods being moved around and there are no functional changes.

Refactoring-only changes like this should be done in a separate commit from functional changes.

src/test/java/org/atlasapi/equiv/results/combining/NullScoreAwareAveragingCombinerTest.java
```diff
@@ -115,4 +115,36 @@ public void testCombiningScoresForSamePublisherAveragesOverHighestNumberOfSource
         assertEquals(Score.valueOf(2.0), combined.candidates().get(equivalent5));
         
     }
+
+    @Test
+    public void testCombineWithIgnoreNullFlagOn() {
```
Dalamar42: Do not prefix tests with `test`. Format should be `ifWeDoXThenWeExpectY`

Dalamar42: This should be multiple tests:

`combineSingleNonNullScoreWithIgnoreNullScoringCandidatesFlagOffReturnsScore`
`combineSingleNonNullScoreWithIgnoreNullScoringCandidatesFlagOnReturnsNull`

etc.

PR #1153, opened 2016-11-11
Ensure equivalence updaters are only explicitly enabled

Dalamar42:
- This caused app servers to start running equivalence jobs because
  equiv updaters used to be enabled by default


lhopki01:
LGTM

PR #1154, opened 2016-11-11
Ensure updaters/feeds are only explicitly enabled

Dalamar42:
- They should be off by default and only enabled on the appropriate
  boxes


lhopki01:
You don't want to set them as false specifically rather than not set them?

Dalamar42:
@lhopki01 The default configuration file is already doing that. I am only changing the overrides

jakekeeys:
lgtm

PR #1156, opened 2016-11-14
Remove c4 press from acceptable publishers

jamiesprax:
This will mean that nothing can equivalate this publishers content but
content from this publisher can still equivalate to other things.


src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -398,7 +397,7 @@ public boolean apply(Publisher input) {
         Set<Publisher> nonStandardPublishers = ImmutableSet.copyOf(Sets.union(
             ImmutableSet.of(ITUNES, BBC_REDUX, RADIO_TIMES, FACEBOOK, LOVEFILM, NETFLIX,
                     RTE, YOUVIEW, YOUVIEW_STAGE, YOUVIEW_BT, YOUVIEW_BT_STAGE, TALK_TALK,
-                    PA, BT_VOD, BT_TVE_VOD, BETTY, AMC_EBS, BT_SPORT_EBS),
+                    PA, BT_VOD, BT_TVE_VOD, BETTY, AMC_EBS, BT_SPORT_EBS, C4_PRESS),
```
Dalamar42: This will remove C4 from the existing matchers completely. We want to be able to run equivalence on C4_PRESS itself for which you need to register a custom updater for it a couple of lines below this.

PR #1157, opened 2016-11-15
Add support for URI resolution to the debug un-publish and publish

jakekeeys:


src/main/java/org/atlasapi/system/UnpublishContentController.java
```diff
@@ -51,63 +50,99 @@ public UnpublishContentController(
...
+    private LookupEntry getLookupEntry(Optional<String> id, Optional<String> uri) {
+        LookupEntry lookupEntry;
+        if(id.isPresent()){
+            Long contentId = idCodec.decode(id.get()).longValue();
+            lookupEntry = lookupEntryStore
+                    .entriesForIds(Lists.newArrayList(contentId))
+                    .iterator()
+                    .next();
+        }
+        else if (uri.isPresent()){
```
emilssolmanis: Unnecessary newline.

src/main/java/org/atlasapi/system/UnpublishContentController.java
```diff
@@ -51,63 +50,99 @@ public UnpublishContentController(
...
+            lookupEntry = lookupEntryStore
+                    .entriesForCanonicalUris(Lists.newArrayList(uri.get()))
+                    .iterator()
+                    .next();
+        }
+        else {
```
emilssolmanis: Same.

src/main/java/org/atlasapi/system/UnpublishContentController.java
```diff
@@ -51,63 +50,99 @@ public UnpublishContentController(
...
+            throw new RuntimeException("id / uri parameter not specified");
```
emilssolmanis: `IllegalArgumentException` is more suitable.

PR #1158, opened 2016-11-17
Remove RT Upcoming from general equivalence

jamiesprax:
Intended behaviour is now:
* RT Upcoming to only equivalate to Amazon Unbox content
* Nothing can equivalate to RT Upcoming.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/EquivModule.java
+++ b/src/main/java/org/atlasapi/equiv/EquivModule.java
@@ -17,6 +17,7 @@ package org.atlasapi.equiv;
 import java.io.File;
 import java.util.Set;
 
+import com.google.common.base.*;
 import org.atlasapi.equiv.generators.BroadcastMatchingItemEquivalenceGenerator;
 import org.atlasapi.equiv.generators.ContainerCandidatesContainerEquivalenceGenerator;
 import org.atlasapi.equiv.generators.ContainerCandidatesItemEquivalenceGenerator;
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -409,6 +410,16 @@ public boolean apply(Publisher input) {
                 .build());
         }
 
+        updaters.register(RADIO_TIMES_UPCOMING, SourceSpecificEquivalenceUpdater.builder(RADIO_TIMES_UPCOMING)
+                .withItemUpdater(broadcastItemEquivalenceUpdater(
+                        ImmutableSet.of(AMAZON_UNBOX),
+                        Score.nullScore(),
+                        Predicates.alwaysTrue()
+                ))
+                .withTopLevelContainerUpdater(NullEquivalenceUpdater.get())
```
Dalamar42: We need to equivalate brands (i.e. topLevelContainers) here as well which will need a different kind of updater. Check one of the other jobs that equivalates brands.

src/main/java/org/atlasapi/remotesite/itunes/epf/ItunesEpfUpdateTask.java
```diff
@@ -261,7 +262,9 @@ public boolean process(EpfVideo row) {
         });
     }
 
-    private Multimap<Series, Episode> linkEpisodesAndSeries(EpfTable<EpfCollectionVideo> cvTable,
+    private Multimap<Series, Episode> linkEpisodesAndSeries(
```
Dalamar42: I think this is in the wrong commit :) :P

src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -409,6 +418,12 @@ public boolean apply(Publisher input) {
                 .build());
         }
 
+        updaters.register(RADIO_TIMES_UPCOMING, SourceSpecificEquivalenceUpdater.builder(RADIO_TIMES_UPCOMING)
+                .withItemUpdater(rtUpcomingItemUpdater)
+                .withTopLevelContainerUpdater(topLevelContainerUpdater(ImmutableSet.of(AMAZON_UNBOX)))
+                .withNonTopLevelContainerUpdater(standardSeriesUpdater(ImmutableSet.of(AMAZON_UNBOX)))
```
Dalamar42: Check if we actually get series from this source. If not it might be best to omit it

PR #1164, opened 2016-11-18
Add equivalence generator metadata

Dalamar42:
- This will show which sources we equivalate to
- Also add support in equivalence configuration endpoint to only show
  equivalence for a single publisher
- Add sorting by publisher to make the output more readable

### DIFF
--- a/src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
+++ b/src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
@@ -1,21 +1,26 @@
 package org.atlasapi.equiv.update.www;
 
 import java.io.IOException;
+import java.util.List;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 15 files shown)


src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
```diff
@@ -100,8 +105,22 @@ public void runUpdate(HttpServletResponse response,
...
+    ) throws IOException {
+        ImmutableSet<Publisher> requestedSources;
+
+        if (sources != null) {
+            requestedSources = sources.stream()
+                    .map(Publisher::fromKey)
+                    .map(Maybe::requireValue)
+                    .collect(MoreCollectors.toImmutableSet());
+        } else {
+            requestedSources = Publisher.all();
```
AdamDM: Is it worth printing somewhere that all publishers are accepted, rather than printing out every publisher?

PR #1170, opened 2016-11-24
Refactor equivalence configuration for readability

Dalamar42:
- The new structure is:
-- EquivalenceUpdaterProvider -> Class which provides a single updater.
   Unlike before  those classes do not share any logic between them
   to avoid coupling different updaters
-- ItemEquivalenceUpdaterType & ContainerEquivalenceUpdaterType ->
   Enums which provide the connection between the configuration and
   specific updaters
-- EquivalenceUpdaterProviderDependencies -> Holder for all the
   different dependencies the updaters may need to simplify
   constructing them
-- DefaultConfiguration -> Holder for hardcoded source groupings, e.g.
   MUSIC_SOURCES, etc.
-- UpdaterConfigurationRegistry -> Contains the updater
   configurations for each source
-- EquivModule -> Is only responsible for injecting the dependencies
   and iterating over the registry to construct and register each
   updater according to its configuration

### DIFF
--- a/src/main/java/org/atlasapi/equiv/EquivModule.java
+++ b/src/main/java/org/atlasapi/equiv/EquivModule.java
@@ -15,74 +15,19 @@ permissions and limitations under the License. */
 package org.atlasapi.equiv;
 
 import java.io.File;
-import java.util.Set;
 
-import com.google.common.base.*;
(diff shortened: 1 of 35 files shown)


Dalamar42:
This will be tested using the /equivalence/configuration endpoint to ensure nothing has changed so this RfC is for the overall structure rather than individual correctness.

PR #1171, opened 2016-11-24
Add logging to itunes series extractor

jamiesprax:
This will allow us to see the strings that fail and improve the regex if we think they should have passed.


AdamDM:
lgtm

PR #1174, opened 2016-11-25
Update regex for series extractor (iTunes)

jamiesprax:
The regex did not handle cases where there was a comma or
colon trailing the series number to continue the name string
such as "this series 2, blah" / "series 2: very good"


AdamDM:
lgtm

PR #1179, opened 2016-11-29
altering FIVE equivalence setup to ignore nullscoring candidates

AdamDM:
Change equiv setup so that the publisher FIVE will ignore candidates that don't have scores from scorers, so that equivalence cannot occur from generators alone


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -671,7 +692,7 @@ private UpdaterConfiguration makeRteConfiguration() {
                 Publisher.ITV,
                 Publisher.C4_PMLSD,
                 Publisher.C4_PMLSD_P06,
-                Publisher.FIVE
+                FIVE
```
emilssolmanis: Make these consistent. Either they're all static imports, or none.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/FiveItemUpdaterProvider.java
```diff
@@ -0,0 +1,120 @@
...
+    public EquivalenceUpdater<Item> getUpdater(
+            EquivalenceUpdaterProviderDependencies dependencies,
+            Set<Publisher> targetPublishers
+    ) {
+        return ContentEquivalenceUpdater.<Item>builder()
+                .withExcludedUris(
+                        dependencies.getExcludedUris()
+                )
+                .withGenerators(
+                        ImmutableSet.<EquivalenceGenerator<Item>>of(
```
emilssolmanis: Huh? Doesn't it infer the generic type?

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -661,6 +664,24 @@ private UpdaterConfiguration makeRteConfiguration() {
...
+                        FIVE_ITEM,
+                        MoreSets.add(TARGET_SOURCES, LOVEFILM)
+                )
+                .withTopLevelContainerEquivalenceUpdater(
+                        STANDARD_TOP_LEVEL_CONTAINER,
+                        MoreSets.add(TARGET_SOURCES, LOVEFILM)
+                )
+                .withNonTopLevelContainerEquivalenceUpdater(
+                        STANDARD_SERIES,
+                        TARGET_SOURCES
```
Dalamar42: Is there a reason we want the target sources for non-top-level-containers to be different than the rest? If not we should make sure they are consistent.

Dalamar42: Let's remove Lovefilm entirely from here actually.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/FiveItemUpdaterProvider.java
```diff
@@ -0,0 +1,120 @@
...
+import org.atlasapi.equiv.update.updaters.providers.EquivalenceUpdaterProviderDependencies;
+import org.atlasapi.media.entity.Item;
+import org.atlasapi.media.entity.Publisher;
+
+import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.joda.time.Duration;
+
+public class FiveItemUpdaterProvider implements EquivalenceUpdaterProvider<Item> {
```
Dalamar42: Let's check if there is an existing updater that does the same thing. If so we might just rename and reuse that one for consistency

PR #1181, opened 2016-12-01
Clean up PA channel and channel group ingester, bump it to Java 8 etc.

emilssolmanis:


src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelGroupsIngester.java
```diff
@@ -79,114 +111,137 @@ public void addChannelNumberings(List<EpgContent> epgContents, Map<String, Regio
             if (regionalisationList == null) {
                 // add to all regions
                 for (Region region : regions.values()) {
-                    clearNumberings(channel, region);
-                    channel.addChannelNumber(region, epgContent.getChannelNumber(), startDate, endDate);
+                    channel.addChannelNumber(
+                            region,
+                            epgContent.getChannelNumber(),
+                            startDate,
+                            endDate
```
AdamDM: Consider using Optional rather than occasionally passing null. Will require further changes to implement though.

PR #1183, opened 2016-12-06
Feature/rt model changes

AdamDM:


src/main/java/org/atlasapi/output/simple/DescribedModelSimplifier.java
```diff
@@ -42,25 +47,36 @@
...
     private final RatingModelSimplifier ratingModelSimplifier = new RatingModelSimplifier();
-    private final AudienceStatisticsModelSimplifier audienceStatsModelSimplifier = new AudienceStatisticsModelSimplifier();
+    private final AudienceStatisticsModelSimplifier audienceStatsModelSimplifier =
+            new AudienceStatisticsModelSimplifier();
+    private final SimplePersonModelSimplifier simplePersonModelSimplifier =
+            new SimplePersonModelSimplifier();
 
     protected DescribedModelSimplifier(ImageSimplifier imageSimplifier) {
         this.imageSimplifier = imageSimplifier;
         this.descriptionWatermarker = null;
```
jamiesprax: its already null

src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
```diff
@@ -69,7 +67,10 @@ public ContentWriteExecutor(
         this.eventResolver = checkNotNull(eventResolver);
     }
 
-    public InputContent parseInputStream(InputStream inputStream, Boolean strict) throws IOException,
+    public InputContent parseInputStream(
+            InputStream inputStream,
+            Boolean strict
+    ) throws IOException,
             ReadException {
```
jamiesprax: this doesnt need to be on a different line

src/main/java/org/atlasapi/output/simple/DescribedModelSimplifier.java
AdamDM: It's final so it complains if it isn't initialised

PR #1184, opened 2016-12-07
adding normalisation logic for titles that looks for variation on vs

AdamDM:
examples were found where titles separated by v were not matching the same titles separated by vs
this fixes that problem

### DIFF
--- a/src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
@@ -139,8 +139,8 @@ public class TitleMatchingItemScorer implements EquivalenceScorer<Item> {
 
     private Score compareTitles(final String subjectTitle, final String suggestionTitle) {
         boolean matches;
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
```diff
@@ -139,8 +139,8 @@ private String removeYearFromTitle(Item item) {
 
     private Score compareTitles(final String subjectTitle, final String suggestionTitle) {
         boolean matches;
-        String subjTitle = normalize(subjectTitle);
-        String suggTitle = normalize(suggestionTitle);
+        String subjTitle = normalize(normaliseForVs(subjectTitle));
+        String suggTitle = normalize(normaliseForVs(suggestionTitle));
```
jamiesprax: Is there a case where you would want to use normalise without normalising for Vs? If not you can just add it as a part of the normalise process instead of calling it explicitly.

Also note spellings z -> s, normaliszszse it while you're here ;)

PR #1185, opened 2016-12-12
Feature/content type updates

jakekeeys:


src/main/java/org/atlasapi/query/content/merge/BroadcastMerger.java
```diff
@@ -1,230 +1,58 @@
...
-import com.google.common.collect.ImmutableMultimap;
-import com.google.common.collect.ImmutableSet;
-import org.joda.time.DateTime;
-import org.joda.time.Interval;
-
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.gson.internal.$Gson$Preconditions.checkNotNull;
 
 public class BroadcastMerger {
```
Dalamar42: `BroadcastMerger`, `DefaultBroadcastMerger` and `VersionMerger` is a little confusing. Can we instead move the logic from the `BroadcastMerger` to the `VersionMerger` and rename `DefaultBroadcastMerger` back to `BroadcastMerger`?

This way we will only have two classes and `VersionMerger` will call `BroadcastMerger` which makes sense given versions contain broadcasts

src/main/java/org/atlasapi/query/content/merge/ContentMerger.java
```diff
@@ -47,14 +50,75 @@ public Content merge(
         if (possibleExisting.isNothing()) {
             return update;
         }
-        Identified existing = possibleExisting.requireValue();
-        if (existing instanceof Content) {
-            return merge((Content) existing, update, merge, defaultBroadcastMerger);
+        Identified identifiedContent = possibleExisting.requireValue();
+
+        if (identifiedContent instanceof Content) {
```
Dalamar42: You can replace this with a guard to remove the nesting since in the else case you only throw

src/main/java/org/atlasapi/query/content/merge/ContentMerger.java
```diff
@@ -47,14 +50,75 @@ public Content merge(
...
+        if (update instanceof Item){
+            return merge(copyToItem(existing), update, merge, defaultBroadcastMerger);
+        }
+
+        return merge(existing, update, merge, defaultBroadcastMerger);
+    }
+
+    private Content copyToBrand(Content existing){
+        Brand brand = new Brand();
+        Brand.copyTo(existing, brand);
```
Dalamar42: Overloading in Java is implemented using static binding so it's done at compile time. At this point as far as the compiler knows the class of `existing` is `Content` so amongst the overloaded `copyTo` methods it will pick the only one that matches which is `Content.copyTo`.

This means that all these methods will only ever copy `Content` fields to the new instance. Even if we are converting `Film` to `Episode` which both subclass `Item` the fields of `Item` will be lost.

Are we ok with this as a compromise here?
If we are we should add a comment to document that to make sure anyone else working on this knows it's deliberate.

PR #1187, opened 2016-12-20
fix creating brand refs for new episodes bug

dragospitica:


emilssolmanis:
I'm not 100% sure this is correct. What about `Episode`s, shouldn't their parent be the series, not the brand?

Dalamar42:
I'm also not sure this is correct for the reason @emilssolmanis mentioned and also because I'm not sure the content store after this will correctly update the parent ref if it sees it has changed.

A quick test around the mongo content store should confirm the second part.

PR #1195, opened 2017-01-11
Small fix to deal with empty string being passed in instead of null

Dalamar42:


src/main/java/org/atlasapi/AtlasMain.java
```diff
@@ -9,10 +23,10 @@
 import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
+import com.google.api.client.repackaged.com.google.common.base.Strings;
```
emilssolmanis: Huh?

Dalamar42: Right. Funny that should happen today. Fixed

PR #1196, opened 2017-01-17
added endpoint to unpublish RT content removed from the feed

dragospitica:


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -103,7 +110,9 @@ private ContentWriteController(
             LookupEntryStore lookupEntryStore,
             ContentResolver contentResolver,
             ContentWriter contentWriter,
-            EquivalenceBreaker equivalenceBreaker
+            EquivalenceBreaker equivalenceBreaker,
+            OldContentDeactivator oldContentDeactivator,
+            ApplicationConfigurationFetcher configurationFetcher
```
jamiesprax: this is already passed in

src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -389,7 +391,9 @@ private ContentWriteController contentWriteController() {
                         contentResolver,
                         entryStore,
                         lookupWriter
-                )
+                ),
+                oldContentDeactivator,
+                configFetcher
```
jamiesprax: already passed in

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -159,6 +174,44 @@ public WriteResponse unpublishContent(HttpServletRequest req, HttpServletRespons
...
+        }
+
+        String source = request.getParameter("source");
+        ImmutableList<String> valid_uris = ImmutableList.of(
+                request.getParameter("valid_uris")
+        );
+
+        Publisher publisher = Publisher.fromKey(source).requireValue();
+
+        oldContentDeactivator.deactivateOldContent(publisher, valid_uris, 50);
```
jamiesprax: Pretty major: I don't see any checking that the api key has write permissions for this publisher.

dragospitica: @jamiesprax the API key validation happens above when fetching the config from the request. This logic was already there.

jamiesprax: @dragospitica are you sure? I'm pretty sure that just resolves an application for an API key. Check the other methods in this class near the end. They do appConfig.canWrite(<publisher> checks.

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -159,6 +171,68 @@ public WriteResponse unpublishContent(HttpServletRequest req, HttpServletRespons
     }
 
     @Nullable
+    @RequestMapping(value = "/4/content.json", method = RequestMethod.DELETE)
```
jamiesprax: /3.0/content.json
You're in Owl

PR #1199, opened 2017-01-26
Shred homegrown HTTP client in Amazon ingester and replace with Apach…

emilssolmanis:
…e HC

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxHttpFeedSupplier.java
+++ b/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxHttpFeedSupplier.java
@@ -1,24 +1,26 @@
 package org.atlasapi.remotesite.amazonunbox;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
-import java.io.FileInputStream;
-import java.io.InputStream;
-import java.util.zip.ZipEntry;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


dragospitica:
LGTM

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxHttpFeedSupplier.java
```diff
@@ -31,43 +33,43 @@
...
     @Override
     public ImmutableList<AmazonUnboxItem> get() {
-        
-        try {
+        HttpGet get = new HttpGet(uri);
+        final ItemCollatingAmazonUnboxProcessor processor = new ItemCollatingAmazonUnboxProcessor();
+        final AmazonUnboxContentHandler handler = new AmazonUnboxContentHandler(processor);
+
+        try (
+                CloseableHttpClient client = HttpClients.createDefault();
```
Dalamar42: Do we need to make a new one every time?

emilssolmanis: Depends. Given the job only runs ever 12 hours, don't see a point in leaving it open consuming threads and whatnot.

Dalamar42: :+1:

PR #1202, opened 2017-01-30
Add logic to unpublish rt content

dragospitica:


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -74,6 +79,8 @@
     public static final String ASYNC_PARAMETER = "async";
     public static final String BROADCAST_ASSERTIONS_PARAMETER = "broadcastAssertions";
     private static final String STRICT_PARAMETER = "strict";
+    private static final String SOURCE = "source";
+    private static final String VALID_URIS = "valid_uris";
```
jamiesprax: I would append these with "_PARAMETER" for consistency

Dockerfile
```diff
@@ -0,0 +1,357 @@
+FROM 448613307115.dkr.ecr.eu-west-1.amazonaws.com/jvm-oracle:latest
```
Dalamar42: Why are we pushing the dockerfile as part of this change? Those should be unrelated changes.

Stage branch should already have a dockerfile and prod is not on docker yet.

src/main/java/org/atlasapi/query/QueryWebModule.java
```diff
@@ -389,6 +393,14 @@ private ContentWriteController contentWriteController() {
                         contentResolver,
                         entryStore,
                         lookupWriter
+                ),
+                new OldContentDeactivator(
```
Dalamar42: If any of these classes are part of the `atlas` repo let's change them to have static factory methods as per the code style

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -155,9 +167,75 @@ public WriteResponse putContent(HttpServletRequest req, HttpServletResponse resp
...
+
         return setPublishStatus(req, resp, false);
     }
 
+    private WriteResponse unpublishOldContent(
+            HttpServletRequest request,
+            HttpServletResponse response
+    ) {
+        String source = request.getParameter(SOURCE);
+        ImmutableList<String> validUris = ImmutableList.copyOf(
```
Dalamar42: You can just do 
```
ImmutableList.copyOf(
    request.getParameter(VALID_URIS)
        .split(",")
);
```

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -155,9 +167,75 @@ public WriteResponse putContent(HttpServletRequest req, HttpServletResponse resp
...
+            return writeResponse;
+        }
+
+        if (oldContentDeactivator.deactivateOldContent(publisher, validUris, 50)) {
+            response.setStatus(HttpServletResponse.SC_OK);
+        }
+        return null;
+    }
+
+    public WriteResponse validateApiKey(
```
Dalamar42: This is duplicating logic that already exists in the other methods in this class. Let's split this into two methods and reuse them. The first one will resolve the application configuration and ensure it exists. This one should happen in `#unpublishContent` before anything else.

The second one should validate the application has access to a particular source and should happen for each case where it already is.

src/test/java/org/atlasapi/query/v2/ContentWriteControllerTest.java
```diff
@@ -137,7 +139,7 @@ public void setUp() throws Exception {
         controller = ContentWriteController.create(
```
Dalamar42: Let's fix this for the code style. One line per parameter.

src/test/java/org/atlasapi/query/v2/ContentWriteControllerTest.java
```diff
@@ -93,6 +94,7 @@
     private @Mock Identified identified;
     private @Mock Described described;
     private @Mock SubstitutionTableNumberCodec substitutionTableNumberCodec;
+    private @Mock OldContentDeactivator oldContentDeactivator;
```
Dalamar42: We need tests for the new logic. At least one that ensures we call the deactivator with the right params from the call and one that ensure that we do not deactivate anything if the application does not have write permissions for the given source.

src/test/java/org/atlasapi/query/v2/ContentWriteControllerTest.java
```diff
@@ -251,4 +274,51 @@ public void postAsynchronously() throws Exception {
...
+        controller.unpublishContent(request, response);
+
+        ArgumentCaptor<AtlasErrorSummary> argument = ArgumentCaptor.forClass(
+                AtlasErrorSummary.class
+        );
+
+        verify(modelWriter, atLeast(2))
+                .writeError(eq(request), eq(response), argument.capture());
+
+        assertEquals("403", argument.getValue().statusCode().toString());
```
Dalamar42: Prefer `assertThat(something, is(something));`. It's more readable

src/test/java/org/atlasapi/query/v2/ContentWriteControllerTest.java
```diff
@@ -251,4 +274,51 @@ public void postAsynchronously() throws Exception {
...
+        when(request.getParameter("source")).thenReturn("upcoming.radiotimes.com");
+        when(request.getParameter("valid_uris")).thenReturn("uri1,uri2,uri3");
+
+        controller.unpublishContent(request, response);
+
+        ArgumentCaptor<AtlasErrorSummary> argument = ArgumentCaptor.forClass(
+                AtlasErrorSummary.class
+        );
+
+        verify(modelWriter, atLeast(2))
```
Dalamar42: Why are we calling this twice?

PR #1210, opened 2017-02-09
Update c5 ingester with new scheduler jobs

AdamDM:


src/main/java/org/atlasapi/remotesite/five/FiveModule.java
```diff
@@ -40,24 +40,41 @@
     
     @PostConstruct
     public void startBackgroundTasks() {
-        scheduler.schedule(fiveUpdater().withName("Five Updater"), DAILY);
+        scheduler.schedule(
+                fiveUpdater(FiveUpdater.TimeFrame.ALL)
+                        .withName("Five Updater")
+                , DAILY
```
Dalamar42: Let's change this one to 1/week on Friday night

src/main/java/org/atlasapi/remotesite/five/FiveModule.java
```diff
@@ -40,24 +40,41 @@
...
+        scheduler.schedule(
+                fiveUpdater(FiveUpdater.TimeFrame.TODAY)
+                        .withName("Five Today Updater")
+                , RepetitionRules.NEVER
+        );
+        log.info("Installed FiveToday updater");
+        scheduler.schedule(
+                fiveUpdater(FiveUpdater.TimeFrame.PLUS_MINUS_7_DAYS)
+                        .withName("Five +-7 Days Updater")
+                , RepetitionRules.NEVER
```
Dalamar42: And this one to 1/day

src/main/java/org/atlasapi/remotesite/five/FiveModule.java
```diff
@@ -40,24 +40,41 @@
...
-                fiveLocationPolicyIds(),
-                Configurer.get("five.timeout.socket", "180").toInt()
-        );
+    public FiveUpdater fiveUpdater(FiveUpdater.TimeFrame timeFrame) {
+        return FiveUpdater.builder()
+                .withContentWriter(contentWriter)
+                .withChannelResolver(channelResolver)
+                .withContentResolver(contentResolver)
+                .withLocationPolicyIds(fiveLocationPolicyIds())
+                .withSocketTimeout(Configurer.get("five.timeout.socket", "180").toInt())
```
Dalamar42: Let's move this parameter to the top of the class as a field to be consistent with how configuration works elsewhere

src/main/java/org/atlasapi/remotesite/five/FiveModule.java
jamiesprax: put these commas on the line above

src/main/java/org/atlasapi/remotesite/five/FiveModule.java
jamiesprax: Extract the Configurer.get to the top of the class with checkNotNull

src/test/java/org/atlasapi/remotesite/five/FiveUpdaterTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        assertEquals(expectedCall, foundCall);
+    }
+
+    private void getAndTestApiCall(FiveUpdater.TimeFrame timeFrame, String startDateTime, String endDateTime) {
+        FiveUpdater fiveUpdater = fiveUpdaterBuilder
+                .withTimeFrame(timeFrame)
+                .build();
+
+        DateTimeFormatter formatter = DateTimeFormat.forPattern("y-M-d'T'hh:mm:ss'Z'");
```
Dalamar42: This is ISO 8601. There is already a predefined formatter for this

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -153,11 +184,32 @@ private SimpleHttpClient buildFetcher() {
                 .build();
     }
 
-    private String getApiCall(int startingPoint) {
-        return String.format("%s/shows?offset=%d&limit=%d", BASE_API_URL, startingPoint, LIMIT);
+    @VisibleForTesting
```
Dalamar42: This is part of the class' internal logic. We should not be exposing it for testing. If the class does too many things to test we should break up the class

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -153,11 +184,32 @@ private SimpleHttpClient buildFetcher() {
...
+    public String getApiCall(int startingPoint, String channelId, DateTime dateTime) {
+        DateTimeFormatter formatter = DateTimeFormat.forPattern("y-M-d'T'hh:mm:ss'Z'");
+        String baseUri = "https://pdb.five.tv/internal/channels/%s/transmissions"
+                + "?expand=watchable&from_time=%s&until_time=%s";
+        if (timeFrame == TimeFrame.TODAY) {
+            DateTime todayStartDateTime = dateTime.withTimeAtStartOfDay();
+            DateTime todayEndDateTime = todayStartDateTime
+                    .plusDays(1)
+                    .withTimeAtStartOfDay()
+                    .minusMinutes(1);
```
Dalamar42: Why `.minusMinutes(1)` and not `.minusMillis(1)`?

Also if their API considers to the `to` date to be exclusive (needs checking) then you don't need to subtract anything. You can just set the `to` as the next midnight

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -132,13 +114,62 @@ public void runTask() {
         }
     }
 
+    private ArrayList<Element> combineElements(int startingPoint) throws Exception {
```
Dalamar42: Return immutable collections and if you absolutely need a mutable one which you don't need here then define them using the interface `List<>` not the concrete type `ArrayList<>`

Dalamar42: This method is hard to follow. Let's speak f2f.

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -132,13 +114,62 @@ public void runTask() {
...
+            }
+
+            return elementArrayList;
+        }
+
+        // get list of channel ids
+        Document channelIdsDocument = streamHttpClient.get(new SimpleHttpRequest<>(
+                "https://pdb.five.tv/internal/channels",
+                (responsePrologue, inputStream) -> parseResponse(inputStream)
+        ));
```
jamiesprax: a "combineElements" method should probably just combine elements, not make a http call

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -132,13 +114,62 @@ public void runTask() {
...
+        Elements channelIds = channelIdsDocument.getRootElement().getChildElements("channels");
+
+        for (int i = 0; i < channelIds.size(); i++) {
+            Element id = channelIds.get(i).getFirstChildElement("id");
+            Elements showElements = getElements(startingPoint, id.getValue());
+            for (int j = 0; j < showElements.size(); j++) {
+                Element element = showElements.get(i);
+                elementArrayList.add(element);
+            }
+        }
```
jamiesprax: Pretty sure this can be done cleaner using a lambda

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
```diff
@@ -153,11 +184,32 @@ private SimpleHttpClient buildFetcher() {
...
+            DateTime todayStartDateTime = dateTime.withTimeAtStartOfDay();
+            DateTime todayEndDateTime = todayStartDateTime
+                    .plusDays(1)
+                    .withTimeAtStartOfDay()
+                    .minusMinutes(1);
+            String todayStart = todayStartDateTime.toString(formatter);
+            String todayFinish = todayEndDateTime.toString(formatter);
+            return String.format(baseUri, channelId, todayStart, todayFinish);
+        } else if (timeFrame == TimeFrame.PLUS_MINUS_7_DAYS) {
+            DateTime weekAgo = dateTime.minusDays(7);
```
jamiesprax: I'd get rid of the assignment to a DateTime and just add .toString(formatter) on the end of this to just set a String seeing as you don't use the DateTime anywhere else

src/main/java/org/atlasapi/remotesite/five/FiveModule.java
```diff
@@ -37,27 +39,45 @@
...
-    
+    private final int socketTimeout = checkNotNull(Configurer.get("five.timeout.socket", "180").toInt());
+
     @PostConstruct
     public void startBackgroundTasks() {
-        scheduler.schedule(fiveUpdater().withName("Five Updater"), DAILY);
+        scheduler.schedule(
+                fiveUpdater(FiveUpdater.TimeFrame.ALL)
+                        .withName("Five Updater"),
+                DAILY
```
jamiesprax: Inconsistent with the other RepetitionRules. Having it as RepetitionRule.X makes it clearer what this is

src/main/java/org/atlasapi/remotesite/five/FiveUpdater.java
AdamDM: it was specified by Romain this way

src/test/java/org/atlasapi/remotesite/five/FiveUpdaterTest.java
AdamDM: The predefined formatter seems to give a +1.00 for the time zone

Dockerfile
```diff
@@ -10,6 +10,7 @@ ENV SERVER_PORT="8080" \
     EQUIV_STREAM_UPDATER_ENABLED="false" \
     EQUIV_UPDATER_ENABLED="false" \
     IDS_GENERATE="false" \
+    JSSE_ENABLESNIEXTENSION="false" \
```
jamiesprax: Keep the naming the same. "JSSE_SNI_EXTENSION_ENABLED" or something, I'm not quite sure where the different parts end

PR #1211, opened 2017-02-13
Mbst 18073/pa summaries

Dalamar42:


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -140,7 +143,26 @@ public static UpdaterConfigurationRegistry create() {
...
 
-        return configurations.build();
+        ImmutableList<UpdaterConfiguration> builtConfigurations = configurations.build();
+
+        validate(builtConfigurations);
+
+        return builtConfigurations;
+    }
+
+    private void validate(ImmutableList<UpdaterConfiguration> configurations) {
```
emilssolmanis: This can be static as well, no?

Dalamar42: oh yeah, missed it

PR #1212, opened 2017-02-14
Feature/migrate equiv tools to use ids

dragospitica:


AdamDM:
squash your commits!

src/main/java/org/atlasapi/equiv/results/www/EquivGraphController.java
```diff
@@ -1,60 +1,89 @@
...
+            throw new IllegalArgumentException("Must specify at least one of 'uri' or 'id'");
+        }
+
         int minimumEdgeCount = Integer.parseInt(minEdges);
+
+        model.put("id", id);
         model.put("uri", uri);
         model.put("min_edges", minimumEdgeCount);
+
+        String uriFromId = getUriFromId(id).get();
```
AdamDM: you're returning an Optional then immediately getting from it without checking it's present

src/main/java/org/atlasapi/equiv/generators/EquivalenceGenerators.java
```diff
@@ -19,17 +23,28 @@
...
     }
     
     public List<ScoredCandidates<T>> generate(T content, ResultDescription desc) {
         desc.startStage("Generating equivalences");
         Builder<ScoredCandidates<T>> generatedScores = ImmutableList.builder();
 
-        if (excludedUris.contains(content.getCanonicalUri())) {
+        ImmutableList<Long> excludedIds = excludedUris.stream()
+                .filter(id -> !id.contains("http"))
+                .map(id -> codec.decode(id).longValue())
```
jamiesprax: Wouldn't you need to do a bit of parsing to get the id from the uri?

src/main/java/org/atlasapi/equiv/results/filters/ExclusionListFilter.java
```diff
@@ -6,24 +6,39 @@
...
+        this.codec = SubstitutionTableNumberCodec.lowerCaseOnly();
     }
     
     @Override
     protected boolean doFilter(ScoredCandidate<T> candidate, T subject, ResultDescription desc) {
-        boolean result = !excludedUris.contains(candidate.candidate().getCanonicalUri());
+
+        ImmutableList<Long> excludedIds = excludedUris.stream()
+                .filter(id -> !id.contains("http"))
+                .map(id -> codec.decode(id).longValue())
```
jamiesprax: And here, "id" is actually a uri at this point

src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
```diff
@@ -104,6 +110,30 @@ public void runUpdate(HttpServletResponse response,
...
+        Iterable<Long> ids = StreamSupport.stream(commaSplitter.split(csvIds).spliterator(), false)
+                .map(input -> codec.decode(input).longValue())
+                .collect(Collectors.toList());
+
+        return StreamSupport.stream(lookupEntryStore.entriesForIds(ids).spliterator(), false)
+                .map(LookupEntry.TO_ID::apply)
+                .collect(Collectors.toList());
+    }
+
+    private Runnable updateFor(final Content content) {
```
jamiesprax: Does this need to be final?

src/main/java/org/atlasapi/equiv/results/www/EquivGraphController.java
dragospitica: @AdamDM good point

src/main/java/org/atlasapi/equiv/generators/EquivalenceGenerators.java
dragospitica: @jamiesprax the list would include IDs that get decoded into long values. e.g. the list would have `d4vqdx` that gets transformed into `40161410`

src/main/java/org/atlasapi/equiv/results/filters/ExclusionListFilter.java
dragospitica: @jamiesprax no, at this point I get rid of any string that has `http` (all URIs) and there would be only the IDs

src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
dragospitica: @jamiesprax that's the way it worked before, so I don't think it's going to change anything

src/main/java/org/atlasapi/equiv/EquivModule.java
```diff
@@ -142,10 +144,11 @@ public RecentEquivalenceResultStore equivalenceResultStore() {
...
     }
 
-    private ImmutableSet<String> excludedUrisFromProperties() {
-        if (Strings.isNullOrEmpty(excludedUris)) {
+    private ImmutableSet<String> excludedContentFromProperties(String excludedContent) {
+        if (Strings.isNullOrEmpty(excludedContent)) {
             return ImmutableSet.of();
         }
-        return ImmutableSet.copyOf(Splitter.on(',').split(excludedUris));
+        return ImmutableSet.copyOf(Splitter.on(',').split(excludedContent));
```
AdamDM: string.split() ?

src/main/java/org/atlasapi/equiv/results/www/EquivGraphController.java
```diff
@@ -1,60 +1,92 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
 
 @Controller
 public class EquivGraphController {
-    
+
     private final LookupEntryStore lookupStore;
+    private final SubstitutionTableNumberCodec codec;
 
     public EquivGraphController(LookupEntryStore lookupStore) {
```
jamiesprax: convert to private constructor and a static create method

src/main/java/org/atlasapi/equiv/results/filters/ExclusionListFilter.java
```diff
@@ -6,24 +6,40 @@
...
 
 
 public class ExclusionListFilter<T extends Identified> extends AbstractEquivalenceFilter<T> {
 
     private final Set<String> excludedUris;
+    private final Set<String> excludedIds;
+    private final SubstitutionTableNumberCodec codec;
 
-    public ExclusionListFilter(Iterable<String> excludedUris) {
+    public ExclusionListFilter(Iterable<String> excludedUris, Iterable<String> excludedIds) {
```
jamiesprax: convert to private constructor and a static create method

src/main/java/org/atlasapi/equiv/generators/EquivalenceGenerators.java
```diff
@@ -7,29 +7,50 @@
...
+        return new EquivalenceGenerators<T>(generators, excludedUris, excludedIds);
     }
 
     private final List<? extends EquivalenceGenerator<T>> generators;
     private final Set<String> excludedUris;
+    private final Set<String> excludedIds;
+    private final SubstitutionTableNumberCodec codec;
 
-    public EquivalenceGenerators(Iterable<? extends EquivalenceGenerator<T>> generators, Set<String> excludedUris) {
+    public EquivalenceGenerators(
```
jamiesprax: convert to private constructor and a static create method

src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
```diff
@@ -104,6 +110,30 @@ public void runUpdate(HttpServletResponse response,
...
+    private Iterable<String> urisFor(String csvIds) {
+        if (Strings.isNullOrEmpty((csvIds))) {
+            return ImmutableSet.of();
+        }
+        Iterable<Long> ids = StreamSupport.stream(commaSplitter.split(csvIds).spliterator(), false)
+                .map(input -> codec.decode(input).longValue())
+                .collect(Collectors.toList());
+
+        return StreamSupport.stream(lookupEntryStore.entriesForIds(ids).spliterator(), false)
+                .map(LookupEntry.TO_ID::apply)
```
jamiesprax: These constant functions are a bit confusing as it doesn't give you an ID back, it gives a uri. There are getters in the LookupEntry class so you can just use `.map(LookupEntry::uri)` and it's a lot clearer what's happening here

src/test/java/org/atlasapi/equiv/generators/EquivalenceGeneratorsTest.java
```diff
@@ -52,6 +61,7 @@ public void generatorCreatesNoCandidatesForExcludedUri() {
     @Test
     public void generatorCreatesCandidatesForNonExcludedUri() {
         Item item = new Item();
+        item.setId(12345678L);
```
jamiesprax: I'd write some explicit tests for your ID list instead of adding id's into the uri tests. It's quick and a lot clearer what you're testing. Also, decoding the string representation of an ID to a long and using that variable in the tests would be a lot clearer to show which ID matches.

PR #1216, opened 2017-02-21
added file upater for iTunes feed + updated configs

dragospitica:


src/main/java/org/atlasapi/remotesite/itunes/ItunesModule.java
```diff
@@ -13,67 +11,112 @@
...
+    //    @PostConstruct
+    //    public void startBackgroundTasks() {
+    //        //scheduler.schedule(itunesRssUpdater(), AT_NIGHT);
+    //        log.record(new AdapterLogEntry(Severity.INFO).withDescription("iTunes update scheduled task installed").withSource(ItunesRssUpdater.class));
+    //    }
+    //
+    //    public @Bean ItunesRssUpdater itunesRssUpdater() {
+    //        return new ItunesRssUpdater(feeds, HttpClients.webserviceClient(), contentWriter, itunesBrandAdapter, log);
+    //    }
+    //
```
dragospitica: I'm not sure if we still need this commented code as there's no notice why it's been left like this

Dalamar42: Dead code is dead code. Let's delete it

src/main/java/org/atlasapi/remotesite/itunes/ItunesModule.java
```diff
@@ -13,67 +11,112 @@
...
-//    public void startBackgroundTasks() {
-//        //scheduler.schedule(itunesRssUpdater(), AT_NIGHT);
-//        log.record(new AdapterLogEntry(Severity.INFO).withDescription("iTunes update scheduled task installed").withSource(ItunesRssUpdater.class));
-//    } 
-//    
-//    public @Bean ItunesRssUpdater itunesRssUpdater() {
-//        return new ItunesRssUpdater(feeds, HttpClients.webserviceClient(), contentWriter, itunesBrandAdapter, log);
-//    }
-//    
+    private final static Daily FOUR_AM = RepetitionRules.daily(new LocalTime(4, 0, 0));
```
Dalamar42: These two names are inconsistent with how we do things elsewhere. They should be the names of what the job that uses them is doing, not repeat the time.

Dalamar42: Also their type should be `RepetitionRule` not `Daily`

src/main/java/org/atlasapi/remotesite/itunes/ItunesModule.java
```diff
@@ -13,67 +11,112 @@
...
+    private @Value("${itunes.epf.username}") String epfUsername;
+    private @Value("${itunes.epf.password}") String epfPassword;
+    private @Value("${itunes.epf.feedPath}") String feedPath;
+    private @Value("${itunes.epf.localFilesPath}") String localFilesPath;
+    private @Value("${itunes.s3.bucket}") String s3bucket;
+    private @Value("${s3.access}") String s3access;
+    private @Value("${s3.secret}") String s3secret;
+
+
+    private @Autowired SimpleScheduler scheduler;
```
Dalamar42: This is inconsistent with our code style. What was there before was correct. Either that or
```
@Autowired private SimpleScheduler scheduler;
```

src/main/java/org/atlasapi/remotesite/itunes/ItunesModule.java
```diff
@@ -13,67 +11,112 @@
...
+    //
     @PostConstruct
     public void startBackgroundTasks() {
         try {
+            scheduler.schedule(
+                    itunesFileUpdater().withName("iTunes File Updater"),
+                    FOUR_AM
+            );
+
             if (!Strings.isNullOrEmpty(localFilesPath)) {
```
Dalamar42: Is there any case that it would be legitimate for us to start Atlas with a null `localFilesPath`? If there is no good reason for this let's remove it and `checkArgument(!Strings.isNullOrEmpty(localFilesPath));` it instead

Dalamar42: If we are using this value being null as a flag to prevent the iTunes ingester from running in processing hosts where we don't want it to run which I suspect is what's happening here then that's terrible and we should fix by adding another configuration param which will be an explicit enabled flag similar to how we do it elsewhere.

If this flag is false neither the existing jobs nor the file updater should be added to the scheduler.

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+            String password,
+            String feedPath,
+            String localFilesPath,
+            S3Client s3Client
+    ) {
+        this.username = checkNotNull(username);
+        this.password = checkNotNull(password);
+        this.feedPath = checkNotNull(feedPath);
+        this.localFilesPath = checkNotNull(localFilesPath);
+        this.s3client = s3Client;
```
Dalamar42: `checkNotNull`

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+        List<String> latestDirectoryNames = getLatestDirectoryNames();
+        itunesDirectory = latestDirectoryNames.get(0);
+        pricingDirectory = latestDirectoryNames.get(1);
+
+        File localDirectory = new File(localFilesPath);
+        File existingItunesDirectory = getExistingItunesDirectory(localDirectory.listFiles());
+
+        if (existingItunesDirectory == null) {
+            createNewItunesDirectory();
+        } else if (existingItunesDirectory.isDirectory() &&
```
Dalamar42: Why do we need to check this here? Isn't it enough to check if the directory exists, if it doesn't make it, if it does and it's not a directory throw an exception to fail-fast.

Why do we need to check the directory's name and/or delete it?

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+            Matcher pricingMatcher = pricingPattern.matcher(inputLine);
+
+            if (itunesMatcher.find()) {
+                currentDirectories.add(itunesMatcher.group(0));
+            } else if (pricingMatcher.find()) {
+                currentDirectories.add(pricingMatcher.group(0));
+            }
+        }
+
+        in.close();
```
Dalamar42: `BufferedReader` implements `AutoClosable`. Use the [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) pattern instead

Dalamar42: If it hadn't been implementing `AutoClosable` then `in.close();` should have been in a finally block instead

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+
+        return currentDirectories;
+    }
+
+    @Nullable
+    private File getExistingItunesDirectory(File[] files) {
```
Dalamar42: If you are legitimately expecting this to return no file then have it return `Optional<File>` instead of null.

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+        this.feedPath = checkNotNull(feedPath);
+        this.localFilesPath = checkNotNull(localFilesPath);
+        this.s3client = s3Client;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public void handleItunesDirectory() throws IOException {
```
Dalamar42: Why do we need this method to be public? It looks like an internal implementation detail of the updater

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+    private void createNewItunesDirectory() {
+        File newItunesDirectory = new File(
+                String.format("%s%s", localFilesPath, itunesDirectory)
+        );
+
+        newItunesDirectory.mkdir();
+    }
+
+    private void saveFeedFile(String directory, String fileName) throws IOException {
+        URL fileUrl = new URL(String.format("%s%s/%s.tbz", feedPath, directory, fileName));
```
Dalamar42: If I understand correctly this downloads the latest file from them every time, i.e. they don't give us historical files. Just the latest one every time.

If that's right then this is fine.

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
```diff
@@ -0,0 +1,247 @@
...
+        File itunesFeedFile = new File(String.format(
+                "%s%s/%s",
+                localFilesPath,
+                itunesDirectory,
+                fileName
+        ));
+
+        FileUtils.copyURLToFile(
+                fileUrl,
+                itunesFeedFile
```
Dalamar42: Are we downloading the file over the existing one? If so then this can break if e.g.

- we start downloading the new file
- we partially overwrite the existing one
- the server gets restarted and the download is interrupted
- the ingester starts later and tries to read the partial file

It's safer to download the file under a temporary name and then after the download is done execute a move to replace the existing one with it. Move is very fast because it doesn't have to move the actual data.

src/main/java/org/atlasapi/remotesite/itunes/ItunesModule.java
dragospitica: @Dalamar42 the `localFilesPath` is coming from `environment.properties`, so I don't see a case where it would be null. Also, the directory `/data/itunes/` is always created when the processing host start. I don't think we need this validation at all.

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
dragospitica: @Dalamar42 Because there should be only one itunes directory under `/data/itunes/`, therefore if the existing one is `itunes12345678` and the new one is `itunes12345679`, we should replace the existing directory with the new one.

Actually, if the existing directory name and the directory name we get from the feed are similar, we should stop the updating process as there won't be any new files, right? I don't see why we would want to overwrite existing files unless they change them in the meantime.

src/main/java/org/atlasapi/remotesite/itunes/ItunesEpfFileUpdater.java
dragospitica: @Dalamar42 this refers to the comment above

PR #1224, opened 2017-03-06
Feature/allow query channel by pa

dragospitica:


src/main/java/org/atlasapi/query/v2/ChannelController.java
```diff
@@ -54,37 +54,49 @@
...
+            .add(Annotation.VARIATIONS)
+            .add(Annotation.RELATED_LINKS)
+            .build();
 
     private static final AtlasErrorSummary NOT_FOUND = new AtlasErrorSummary(new NullPointerException())
-        .withMessage("No such Channel exists")
-        .withMessage("Channel not found")
-        .withStatusCode(HttpStatusCode.NOT_FOUND);
+            .withMessage("No such Channel exists")
+            .withMessage("Channel not found")
```
jamiesprax: This looks like an error and should likely follow the same format as the others:
.withMessage(...)
.withErrorCode(...)
.withStatusCode(...)

In it's current state the first .withMessage is ignored.

PR #1227, opened 2017-03-13
change UpdaterConfigurationRegistry to turn on Amazon Unbox equivalen…

AdamDM:
…ce for non top level containers


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -583,8 +583,8 @@ private static UpdaterConfiguration makeAmazonUnboxConfiguration() {
                         targetSources
                 )
                 .withNonTopLevelContainerEquivalenceUpdater(
-                        NOP_CONTAINER,
-                        ImmutableSet.of()
+                        STANDARD_SERIES,
+                        targetSources
```
Dalamar42: We need to add RT to `targetSources` as well.

PR #1228, opened 2017-03-14
fix two edge cases in equivalence scorer

AdamDM:
Edge case 1 - title starts with more than one ':' throwing OOB
Edge case 2 - title starts with a number followed by a colon and matches a regex throwing OOB

Due to these changes, titles with a leading ':' may be less likely to get a partial match, but these should be rare anyway. These changes are reflected in the test alterations

### DIFF
--- a/src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
@@ -160,31 +160,37 @@ public class TitleMatchingItemScorer implements EquivalenceScorer<Item> {
     }
 
     private Score partialTitleScore(String subjectTitle, String suggestionTitle) {
-        if (subjectTitle.indexOf(':') == 0) {
-            subjectTitle = subjectTitle.substring(1);
-        }
-        if (suggestionTitle.indexOf(':') == 0) {
-            suggestionTitle = suggestionTitle.substring(1);
-        }
-        if (subjectTitle.contains(":") && suggestionTitle.contains(":")) {
-            String subjTitle = normalizeWithoutReplacing(subjectTitle);
+
+        String subjectTitleWithoutLeadingColons = removeLeadingColons(subjectTitle);
+        String suggestionTitleWithoutLeadingColons = removeLeadingColons(suggestionTitle);
+
+        String subjTitle = normalizeWithoutReplacing(subjectTitleWithoutLeadingColons);
+        String suggTitle = normalizeWithoutReplacing(suggestionTitleWithoutLeadingColons);
+
+        if (subjTitle.contains(":") && suggTitle.contains(":")) {
+
             subjTitle = subjTitle.substring(0, subjTitle.indexOf(":"));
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
```diff
@@ -160,33 +160,39 @@ private Score compareTitles(final String subjectTitle, final String suggestionTi
...
 
         return scoreOnMismatch;
     }
 
+    private String removeLeadingColons(String title) {
+        Pattern colonsAtStart = Pattern.compile(":+(.*)");
+
+        Matcher titleMatcher = colonsAtStart.matcher(title);
+
+        return titleMatcher.matches() ? titleMatcher.group(1) : title;
```
jamiesprax: This looks like what `String.replace` does. Assuming the string isn't null/empty:
`if (title.charAt(0) == ':') {
    return title.replaceFirst(":", "");
}`

src/test/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorerTest.java
```diff
@@ -199,9 +199,20 @@ public void testPartialMatchAfterSemicolon() {
...
+    public void testForMultipleColonsToStart() {
+        DefaultDescription desc = new DefaultDescription();
+        score(0, scorer.score(itemWithTitle("Storage Hunters"), of(itemWithTitle("::::Storage Hunters: UK")), desc));
     }
 
+    @Test
+    public void testForPrefixRemovalBug() {
+        DefaultDescription desc = new DefaultDescription();
+        score(0, scorer.score(itemWithTitle("Storage: Bunters"), of(itemWithTitle(" 5 : Storage Hunters")), desc));
+    }
```
jamiesprax: What would happen if `title = "::::"` ?

src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
AdamDM: It's meant to prevent the edge case of something starting with multiple ':'s from throwing exceptions later in the code

src/test/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorerTest.java
```diff
@@ -199,9 +199,20 @@ public void testPartialMatchAfterSemicolon() {
     @Test
     public void testForOutOfBoundsException() {
         DefaultDescription desc = new DefaultDescription();
-        score(1, scorer.score(itemWithTitle("Storage Hunters"), of(itemWithTitle(":Storage Hunters: UK")), desc));
+        score(0, scorer.score(itemWithTitle("Storage Hunters"), of(itemWithTitle(":Storage Hunters: UK")), desc));
```
Dalamar42: All these tests are returning a score of 0. You need some tests that return a good score for these cases as well

PR #1231, opened 2017-03-16
Expose graphite params to configuration

Dalamar42:


src/main/java/org/atlasapi/AtlasMain.java
```diff
@@ -249,14 +255,19 @@ private GraphiteReporter startGraphiteReporter() {
...
                     .convertDurationsTo(TimeUnit.MILLISECONDS)
                     .filter(MetricFilter.ALL)
-                    .build(new Graphite(GRAPHITE_ADDRESS));
-            if (!IS_PROCESSING) {
+                    .build(new Graphite(new InetSocketAddress(
+                            graphiteHost,
+                            graphitePort
+                    )));
+
+            if (!IS_PROCESSING && graphiteReportingEnabled) {
```
emilssolmanis: Wait, why do we not report metrics for processing...? What am I missing...?

Dalamar42: I don't think there are any useful metrics for processing in there

lhopki01:
lgtm

PR #1232, opened 2017-03-24
Metrics in nitro

jamiesprax:
We don't know whats going on.
Maybe now we can find out whats going on.


src/main/java/org/atlasapi/AtlasMain.java
```diff
@@ -71,7 +71,7 @@
             System.getProperty("metrics.graphite.enabled", "true")
     );
 
-    public final MetricRegistry metrics = new MetricRegistry();
+    public static final MetricRegistry metrics = new MetricRegistry();
```
Dalamar42: You need to also change the code here that prevents reporting of metrics for processing hosts

PR #1233, opened 2017-03-27
Add logging & set explicitly

jamiesprax:
Add logging around graphite reporting status and startup
Explicitly set config variable


src/main/resources/config/environment.properties
```diff
@@ -466,4 +466,6 @@ sitemaps.c4.flashplayerversion.cacheInMinutes=30
 
 unbox.s3.bucket=
 unbox.missingContent.percentage=100
-unbox.url=  
+unbox.url=
+
+metrics.graphite.enabled=true
```
Dalamar42: This is unnecessary. At this point the configuration file is not read yet. This is why we are only reading this value as an environment variable instead of using `Configurer`

dockerfiles/atlas-owl-api/Dockerfile
```diff
@@ -79,6 +79,7 @@ CMD java \
     -Droot.log.level="$ROOT_LOG_LEVEL" \
     -Dapplications.client.host="$APPLICATIONS_CLIENT_HOST" \
     -Dapplications.client.env="$APPLICATIONS_CLIENT_ENV" \
+    -Dmetrics.graphite.enabled="$METRICS_GRAPHITE_ENABLED" \
```
Dalamar42: Are we using this Dockerfile? Check what new jenkins is doing. There is another Dockerfile at the root of the project and that one already has this option

PR #1234, opened 2017-03-28
Add ISO 8601 duration to feed stats endpoint

dragospitica:


AdamDM:
lgtm

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +54,24 @@ public FeedStatsController(
...
 
     @RequestMapping(value="/3.0/feeds/youview/{publisher}/statistics.json", method = RequestMethod.GET)
-    public void statistics(HttpServletRequest request, HttpServletResponse response,
-            @PathVariable("publisher") String publisherStr) throws IOException {
+    public void statistics(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @PathVariable("publisher") String publisherStr
+    ) throws IOException {
+        String isoDuration = request.getParameter("duration");
```
emilssolmanis: This is not necessary, add it as an annotated `@RequestParam`.

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +54,24 @@ public FeedStatsController(
...
+
+        if (Strings.isNullOrEmpty(isoDuration)) {
```
emilssolmanis: This is going to break the endpoint for existing clients of the API, you probably need to figure out  a default value. If not, just use `@RequestParam(required=true)`.

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +54,24 @@ public FeedStatsController(
...
+            errorViewFor(request,
+                    response,
+                    AtlasErrorSummary.forException(new IllegalArgumentException(
+                            "The fromTime request parameter must be specified"
+                    ))
+            );
+        }
+
+        java.time.Duration duration = java.time.Duration.parse(isoDuration);
```
emilssolmanis: Joda isn't used in this class, no need for absolute qualifier. Just import it.

emilssolmanis: The name is off, `duration` has no semantic value, it's literally the type name of the variable.

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +55,15 @@ public FeedStatsController(
     }
 
     @RequestMapping(value="/3.0/feeds/youview/{publisher}/statistics.json", method = RequestMethod.GET)
-    public void statistics(HttpServletRequest request, HttpServletResponse response,
-            @PathVariable("publisher") String publisherStr) throws IOException {
+    public void statistics(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @PathVariable("publisher") String publisherStr,
+            @RequestParam(value = "isoDuration") String isoDuration
```
emilssolmanis: `value` is the default and is optional, you can just write `@RequestParam("isoDuration")`, but again, that is a very non descriptive argument name.

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +55,15 @@ public FeedStatsController(
...
+    ) throws IOException {
+        // we parse the ISO 8601 duration as a period because hours are inexact in terms of milliseconds
+        Period requiredPeriod = Period.parse(isoDuration);
```
emilssolmanis: More bad naming.

PR #1236, opened 2017-04-03
Feature/platform editor be shaman

dragospitica:


src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+public class ChannelGroupTransformer implements
+        ModelTransformer<org.atlasapi.media.entity.simple.ChannelGroup, ChannelGroup> {
+
+    private SubstitutionTableNumberCodec idCodec = new SubstitutionTableNumberCodec();
+
+    @Override
+    public ChannelGroup transform(org.atlasapi.media.entity.simple.ChannelGroup simple) {
+        Platform complex = new Platform();
+
+        if (Strings.isNullOrEmpty(simple.getId())) {
```
AdamDM: I think this should be negated
if (!Strings.isNullOrEmpty(simple.getId())) {

Otherwise it only adds null and ignores it when it's not null

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        complex.setPublisher(getPublisher(simple.getPublisherDetails()));
+        complex.addTitle(simple.getTitle());
+        complex.setAvailableCountries(Countries.fromCodes(simple.getAvailableCountries()));
+
+        return complex;
+    }
+
+    private Publisher getPublisher(PublisherDetails publisherDetails) {
```
AdamDM: avoid naming things 'getSomething' if they're not actual getters. This is possibly extracting it so maybe call it extract publisher.

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+        if (Strings.isNullOrEmpty(publisherDetails.getKey())) {
+            throw new IllegalArgumentException("Missing publisher key.");
```
AdamDM: This might be better off higher in the pipeline so it doesn't get this far and then fail, unless it's there too and this is for completeness

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -75,18 +80,28 @@
...
-    public ChannelGroupController(ApplicationFetcher configFetcher, AdapterLog log, AtlasModelWriter<Iterable<ChannelGroup>> outputter,
-            ChannelGroupResolver channelGroupResolver, ChannelResolver channelResolver, NumberToShortStringCodec idCodec) {
+    public ChannelGroupController(
+            ApplicationFetcher configFetcher,
+            AdapterLog log,
+            AtlasModelWriter<Iterable<ChannelGroup>> outputter,
+            ChannelGroupResolver channelGroupResolver,
+            ChannelGroupWriteController channelGroupWriteController,
+            ChannelResolver channelResolver,
+            NumberToShortStringCodec idCodec
```
AdamDM: should use a builder with this many parameters

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    private WriteResponse deserializeAndUpdateChannelGroup(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            boolean createNewPlatform
+    ) {
+        Boolean strict = Boolean.valueOf(request.getParameter(STRICT));
+
+        // authentication
```
AdamDM: This method is massive, instead of partitioning sections off with comments about what they do simply split into smaller well named methods

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+                    ),
+                    e
+            );
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+            return error(request, response, AtlasErrorSummary.forException(e));
+        }
+
+        response.setStatus(HttpServletResponse.SC_OK);
+
+        return null;
```
AdamDM: Try not to return null

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
```
AdamDM: again, massive method that returns null. Either split it up into smaller methods or even considering (with a class this size) refactoring the whole class into several manageable classes that do specific things

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -52,8 +55,15 @@ public FeedStatsController(
...
-    public void statistics(HttpServletRequest request, HttpServletResponse response,
-            @PathVariable("publisher") String publisherStr) throws IOException {
+    public void statistics(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @PathVariable("publisher") String publisherStr,
+            @RequestParam("timespan") String timespan
+    ) throws IOException {
+        // we parse the ISO 8601 duration as a period because hours are inexact in terms of milliseconds
+        Period timeBeforeNow = Period.parse(timespan);
```
AdamDM: timeBeforeNow is a strange name, what does this time period represent?

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
Dalamar42: That ^
A test should have caught this. Let's also add a test that would fail due to this.

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
Dalamar42: This is fine

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,49 @@
...
+    public ChannelGroup transform(org.atlasapi.media.entity.simple.ChannelGroup simple) {
+        Platform complex = new Platform();
+
+        if (Strings.isNullOrEmpty(simple.getId())) {
+            complex.setId(idCodec.decode(simple.getId()).longValue());
+        }
+
+        complex.setPublisher(getPublisher(simple.getPublisherDetails()));
+        complex.addTitle(simple.getTitle());
+        complex.setAvailableCountries(Countries.fromCodes(simple.getAvailableCountries()));
```
Dalamar42: This can't be every field of the ChannelGroup accounted for. What about channel numberings?

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
Dalamar42: That ^

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -180,24 +200,59 @@ public void listChannel(HttpServletRequest request, HttpServletResponse response
             errorViewFor(request, response, AtlasErrorSummary.forException(e));
         }
     }
+
+    @RequestMapping(value = { "/3.0/channel_groups.*" }, method = RequestMethod.POST)
+    public WriteResponse createChannelGroup(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @RequestParam("newPlatform") String newPlatform
```
Dalamar42: We should not have a `newPlatform` parameter on this endpoint. The usual behaviours for POST/PUT is either:
- POST fails if the entity already exists
- POST acts as PUT if the entity already exists

I believe in our APIs we do the second so if that's correct we should be doing the same here.

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -180,24 +200,59 @@ public void listChannel(HttpServletRequest request, HttpServletResponse response
...
+            ) {
+        boolean createNewPlatform = Boolean.parseBoolean(newPlatform);
+        return channelGroupWriteController.createPlatform(request, response, createNewPlatform);
+    }
+
+    @RequestMapping(value = { "/3.0/channel_groups.*" }, method = RequestMethod.PUT)
+    public WriteResponse updateChannelGroup(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @RequestParam("newPlatform") String newPlatform
```
Dalamar42: Ditto

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -180,24 +200,59 @@ public void listChannel(HttpServletRequest request, HttpServletResponse response
...
+    ) {
+        boolean createNewPlatform = Boolean.valueOf(newPlatform);
+        return channelGroupWriteController.updatePlatform(request, response, createNewPlatform);
+    }
+
+    @RequestMapping(value = { "/3.0/channel_groups/{id}.*" }, method = RequestMethod.DELETE)
+    public WriteResponse deleteChannelGroup(
```
Dalamar42: Is this for unpublishing channel groups? Do we need this for the platform editor? If not it would be best not to add it until we need it.

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ChannelGroupWriteController {
+
+    private static final Logger log = LoggerFactory.getLogger(ChannelGroupWriteController.class);
+    private static final String STRICT = "strict";
```
Dalamar42: The `strict` flag was only a temporary measure to allow us to make validation stricter without breaking existing clients. This is a new endpoint so it should be strict by default and not have the flag

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+    private static final AtlasErrorSummary UNAUTHORIZED = AtlasErrorSummary.forException(new UnauthorizedException(
+            "API key is unauthorised"
```
Dalamar42: Don't we have these error summaries somewhere centralised already? This smells of duplication

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+        this.channelGroupResolver = checkNotNull(builder.channelGroupResolver);
+        this.outputWriter = checkNotNull(builder.outputWriter);
+        this.transformer = checkNotNull(builder.transformer);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public WriteResponse createPlatform(
```
Dalamar42: We should have one method for both creating and updating to allow the POST endpoint to act as PUT if the channel group already exists

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import com.google.api.client.util.Lists;
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ChannelGroupWriteController {
```
Dalamar42: As we discussed the whole `XWriteController` thing is an anti-pattern and we should not be continuing it. Instead the logic that belongs in the controller should just be in the `ChannelGroupController` and the rest in a `ChannelGroupExecutor` class.

I've put a note below about what belongs in the executor

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+        if (!possibleApplication.get()
+                .getConfiguration()
+                .isWriteEnabled(complexChannelGroup.getPublisher())) {
+            response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+            return error(request, response, FORBIDDEN);
+        }
+
+        // write to DB
+        try {
+            long channelGroupId = store.createOrUpdate(complexChannelGroup).getId();
```
Dalamar42: This functionality belongs in the executor

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+            org.atlasapi.media.entity.simple.ChannelGroup simpleChannelGroup
+    ) {
+        return transformer.transform(simpleChannelGroup);
+    }
+
+    private ChannelGroup updateChannelGroupNumberings(
+            ChannelGroup channelGroupToBeUpdated,
+            org.atlasapi.media.entity.simple.ChannelGroup simple
+    ) {
+        List<ChannelNumbering> channelNumberingList = Lists.newArrayList();
```
Dalamar42: This logic belongs in the `ChannelGroupTransformer`

src/main/java/org/atlasapi/query/v2/ChannelGroupWriteController.java
```diff
@@ -0,0 +1,342 @@
...
+                                .longValue())
+                        .withChannelGroup(channelGroupToBeUpdated.getId())
+                        .build()
+        ));
+        channelGroupToBeUpdated.setChannelNumberings(channelNumberingList);
+
+        return channelGroupToBeUpdated;
+    }
+
+    private WriteResponse deletePlatformById(
```
Dalamar42: Ditto as before

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -17,19 +14,25 @@
...
+import org.springframework.web.bind.annotation.RequestParam;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableSet;
-import com.metabroadcast.common.http.HttpStatusCode;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 
 @Controller
 public class FeedStatsController extends BaseController<Iterable<FeedStatistics>> {
```
Dalamar42: Why do you need to change this controller at all? This has nothing to do with channel groups. You should not need to make any changes here. It is good to refactor, but we should also try to keep the footprint of a change as small as possible by not changing classes that are not otherwise part of the change

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        complex.setPublisher(extractPublisher(simple.getPublisherDetails()));
+        complex.addTitle(simple.getTitle());
+        complex.setAvailableCountries(Countries.fromCodes(simple.getAvailableCountries()));
+
+        return complex;
+    }
+
+    public void updateChannelGroupNumberings(
```
Dalamar42: This isn't called by `#transform`. Also why is it public?

src/main/java/org/atlasapi/input/ChannelGroupTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+                        .withChannel(idCodec.decode(
+                                channelNumbering.getChannel().getId()).longValue()
+                        )
+                        .withChannelGroup(channelGroupToBeUpdated.getId())
+                        .build()
+        ));
+        channelGroupToBeUpdated.setChannelNumberings(channelNumberingList);
+    }
+
+    public long decodeId(String id) {
```
Dalamar42: Why is this public?

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException;
+import com.google.common.base.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ChannelGroupExecutor {
```
Dalamar42: `ChannelGroupWriteExecutor`

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+                    .withStatusCode(HttpStatusCode.BAD_REQUEST);
+            return error(request, response, errorSummary);
+        }
+
+        if (!possibleApplication.get()
+                .getConfiguration()
+                .isWriteEnabled(complexChannelGroup.getPublisher())) {
+            response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+            return error(request, response, AtlasErrorSummary.forException(new ForbiddenException()));
+        }
```
Dalamar42: The logic up to here should be in the controller, not in the executor. The executor should only be given a content group and return a write result of some description. It should not be given the http servlet request and response and it should not be setting http status codes or do anything else that has to do with the http layer. That stuff is the responsibility of the controller. The executor should just worry about writing the content group.

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+
+    private Optional<AtlasErrorSummary> createOrUpdateChannelGroup(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            ChannelGroup complexChannelGroup,
+            org.atlasapi.media.entity.simple.ChannelGroup simpleChannelGroup
+    ) {
+        try {
+            if (complexChannelGroup.getId() != null) {
+                store.createOrUpdate(complexChannelGroup).getId();
```
Dalamar42: You are calling `.gitId()` here, but doing nothing with the result. What's the purpose of this?

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    private Optional<AtlasErrorSummary> createOrUpdateChannelGroup(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            ChannelGroup complexChannelGroup,
+            org.atlasapi.media.entity.simple.ChannelGroup simpleChannelGroup
+    ) {
+        try {
+            if (complexChannelGroup.getId() != null) {
```
Dalamar42: I'm not following the logic here. If there is an ID you just call `.createOrUpdate` on the store, but if there isn't one you call the same method with a null ID then get the new ID, then you update the content group you just wrote with the channel numberings?

The logic here should be:
```
if (id == null) {
  create ID
  write content group with new ID
} else {
  get content group with this ID
  if (that content group exists) {
    update existing content group by merging with the new one
  }
  write it
}
```

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+                store.createOrUpdate(complexChannelGroup).getId();
+            } else {
+                long channelGroupId = store.createOrUpdate(complexChannelGroup).getId();
+                com.google.common.base.Optional<ChannelGroup> channelGroupToUpdate = store.channelGroupFor(
+                        channelGroupId
+                );
+                if (channelGroupToUpdate.isPresent()) {
+                    transformer.updateChannelGroupNumberings(
```
Dalamar42: The transformer is responsible for converting the complex model into the simple and vv.

Merging a new content group with an existing one is a different responsibility that should be handled by the executor or a delegate of the executor.

Also, this logic will only update channel group numberings. That is not correct. It's valid to e.g. use this logic to update a channel group's title, etc.

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
```diff
@@ -0,0 +1,295 @@
...
+                    String.format("No such platform exists with ID %s", id)
+            )));
+        }
+
+        if (!possibleApplication.get()
+                .getConfiguration()
+                .isWriteEnabled(possibleChannelGroup.get().getPublisher())) {
+            response.setStatus(HttpServletResponse.SC_FORBIDDEN);
+            return error(request, response, AtlasErrorSummary.forException(new ForbiddenException()));
+        }
```
Dalamar42: Ditto, everything up to here should be in the controller. The executor shouldn't have anything to do with the HTTP request or response

src/test/java/org/atlasapi/input/ChannelGroupTransformerTest.java
```diff
@@ -0,0 +1,54 @@
...
+    @Before
+    public void setUp() {
+        simpleChannelGroup = mock(ChannelGroup.class);
+
+        transformer = new ChannelGroupTransformer();
+        publisherDetails = new PublisherDetails(Publisher.BBC.key());
+    }
+
+    @Test
+    public void testChannelGroupIsTransformedCorrectly() throws Exception {
```
Dalamar42: This test is not testing the numberings

src/test/java/org/atlasapi/query/v2/ChannelGroupExecutorTest.java
```diff
@@ -0,0 +1,87 @@
...
+    private ChannelGroupStore store;
+    private com.google.common.base.Optional<ChannelGroup> channelGroup;
+    private ApplicationFetcher applicationFetcher;
+    private Application application;
+    private ApplicationConfiguration configuration;
+    private HttpServletRequest request;
+    private HttpServletResponse response;
+
+    @Before
+    public void setUp() {
```
Dalamar42: The permissions test should be on the controller test. The executor should test that it calls the store appropriately, that it merges an existing content group with a new one correctly, the it unpublishes content as expected, etc.

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
dragospitica: @Dalamar42 I'll still need to set the status of the HTTP response or return `#error` if any of the create/update/delete process is failing no?

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
dragospitica: @Dalamar42 I moved the update logic from the transformer.

However, if a channelGroup changes its, for example, title, it will be handled by the current process: get channelGroup JSON -> transform simple model to complex model -> find the existing channelGroup -> update the channelGroup

src/main/java/org/atlasapi/query/v2/ChannelGroupExecutor.java
dragospitica: @Dalamar42 the logic of creating a contentGroup ID and updating the existing one (if the ID is not null) is already in `MongoChannelGroupStore`

PR #1238, opened 2017-04-07
handled new PA flags

dragospitica:


src/main/java/org/atlasapi/input/BroadcastModelTransformer.java
```diff
@@ -51,6 +57,8 @@ public Broadcast transform(org.atlasapi.media.entity.simple.Broadcast simple) {
             complex.setBlackoutRestriction(new BlackoutRestriction(simple.getBlackoutRestriction().getAll()));
         }
         complex.setRevisedRepeat(simple.getRevisedRepeat());
+        complex.setContinuation(simple.getContinuation());
```
Dalamar42: This needs to be tested

src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
```diff
@@ -405,6 +408,8 @@ private void addTo(org.atlasapi.media.entity.simple.Item simpleItem, Version ver
         simpleModel.setNewEpisode(broadcast.getNewEpisode());
         simpleModel.setAliases(broadcast.getAliasUrls());
         simpleModel.setRevisedRepeat(broadcast.getRevisedRepeat());
+        simpleModel.setContinuation(broadcast.getContinuation());
```
Dalamar42: Ditto, this needs a test

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -777,7 +777,7 @@ private Broadcast addBroadcast(
     }
 
     private void setDescription(ProgData progData, Item item, boolean isWelshChannel) {
-        if (progData.getBillings() != null) {
+        if (progData.getBillings() != null && progData.getAttr().getContinued().equals("no")) {
```
Dalamar42: I'm not sure how easy it is to test this class, but if we have a testing harness around it this also needs a test

src/main/java/org/atlasapi/remotesite/pa/PaProgrammeProcessor.java
```diff
@@ -1000,6 +1000,8 @@ private Broadcast broadcast(
 
             broadcast.setNewSeries(newSeries);
             broadcast.setNewEpisode(newEpisode);
+            broadcast.setNewOneOff(getBooleanValue(progData.getAttr().getNewOneOff()));
```
Dalamar42: Ditto

PR #1242, opened 2017-04-12
fix bug where null uri gets passed to images which then does not resolve

AdamDM:
We found a series with an image in images with a null uri and it caused issues
This should prevent this from happening in the future

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractor.java
+++ b/src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractor.java
@@ -384,7 +384,7 @@ public class AmazonUnboxContentExtractor implements ContentExtractor<AmazonUnbox
     }
 
     private List<Image> generateImages(AmazonUnboxItem item) {
-        if (item.getLargeImageUrl() == null) {
+        if (Strings.isNullOrEmpty(item.getLargeImageUrl())) {
             return ImmutableList.of();
         }
         Image image = new Image(item.getLargeImageUrl());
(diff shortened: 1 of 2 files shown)


src/test/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractorTest.java
```diff
@@ -374,6 +374,18 @@ public void testExtractionOfTopLevelSeries() {
         assertNull(series.getParent());
     }
 
+    @Test
+    public void testImageExtractionHandlesNullImageUris() {
```
dragospitica: add a test where the imageUrl is ""

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractor.java
```diff
@@ -384,6 +384,9 @@ public String apply(AmazonUnboxGenre input) {
     }
 
     private List<Image> generateImages(AmazonUnboxItem item) {
+        if (item.getLargeImageUrl() == null) {
```
dragospitica: better use `Strings.isNullOrEmpty(item.getLargeImageUrl())`

src/test/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractorTest.java
AdamDM: good point - dealt with

src/test/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentExtractorTest.java
```diff
@@ -374,6 +374,27 @@ public void testExtractionOfTopLevelSeries() {
...
+        AmazonUnboxItem amazonUnboxItem =
+                createAmazonUnboxItem("seasonAsin", ContentType.TVSEASON)
+                        .withLargeImageUrl(null)
+                        .build();
+
+        Series series = (Series) Iterables.getOnlyElement(extractor.extract(amazonUnboxItem));
+
+        assertEquals(0, series.getImages().size());
+
+        AmazonUnboxItem amazonUnboxItem2 =
```
Dalamar42: This should be a separate test. Each test should only be testing one thing.

PR #1243, opened 2017-04-21
stop person from being written if they have no name

AdamDM:
PA are sending us people with no names, this is being set as null in the database and being thrown out by search, therefore not indexed, therefore not discoverable in search, therefore can't be equivalated.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/people/PaPeopleProcessor.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/people/PaPeopleProcessor.java
@@ -38,7 +38,13 @@ public class PaPeopleProcessor {
     }
     
     public void process(org.atlasapi.remotesite.pa.profiles.bindings.Person paPerson) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/remotesite/pa/people/PaPeopleProcessor.java
```diff
@@ -38,7 +38,13 @@ public PaPeopleProcessor(PeopleResolver personResolver, PersonWriter personWrite
     }
     
     public void process(org.atlasapi.remotesite.pa.profiles.bindings.Person paPerson) {
-        Person person = ingestPerson(paPerson);
+        Optional<Person> optionalPerson = ingestPerson(paPerson);
+        Person person;
+        if (!optionalPerson.isPresent()) {
+            return;
+        } else {
```
Dalamar42: The else here is unnecessary since the if branch has a return

PR #1245, opened 2017-05-11
Update channels logo

dragospitica:


src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+                        response,
+                        AtlasErrorSummary.forException(
+                                new IllegalArgumentException(
+                                        "You must specify the height, width and mimeType in order to create/update an image."
+                                )
+                        )
+                );
+            }
+
+            createOrUpdateImage(existingChannel, imageUri, imageTheme, imageMimeType, imageHeight, imageWidth);
```
jamiesprax: It'd be a lot cleaner just to pass the "incomingJson" in here

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+                    || Strings.isNullOrEmpty(imageWidth)
+                    || Strings.isNullOrEmpty(imageMimeType)) {
+                response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
+                return error(
+                        request,
+                        response,
+                        AtlasErrorSummary.forException(
+                                new IllegalArgumentException(
+                                        "You must specify the height, width and mimeType in order to create/update an image."
+                                )
```
jamiesprax: All this is just validation that you can extract to an input validation method. This would also allow you to write tests to cover this much easier

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+            AtlasErrorSummary summary
+    ) {
+        try {
+            outputter.writeError(request, response, summary);
+        } catch (IOException e) {
+        }
+        return null;
+    }
+
+    public static class ImageDetails {
```
jamiesprax: I don't think this is the right place for this. It's only used for deserialisation so it could be a wrapper around an image. I would argue that it belongs in atlas-model with a name like ImageWrapper

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+
+    public Void updateChannelImage(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            NumberToShortStringCodec codec
+    ) throws IOException {
+        ObjectMapper mapper = new ObjectMapper();
+        ImageDetails incomingJson = mapper.readValue(request.getInputStream(), ImageDetails.class);
+
+        Optional<Application> possibleApplication;
```
jamiesprax: This application checking logic can be extracted for both large methods in this class to reduce duplication

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+                    response,
+                    AtlasErrorSummary.forException(
+                            new UnauthorizedException(
+                                    "API key is unauthorised"
+                            )
+                    )
+            );
+        }
+
+        String channelId = incomingJson.getChannelId();
```
jamiesprax: From here down ~50 lines is all validation that can be extracted and testable

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+                            String.format("No channel has been found for ID %s.", channelId)
+                    ))
+            );
+        }
+
+        Channel existingChannel = possibleChannel.requireValue();
+
+        String imageUri = incomingJson.getImageUri();
+        if (Strings.isNullOrEmpty(imageUri)) {
+            deleteImage(imageTheme, existingChannel);
```
jamiesprax: Was there a reason you went for parsing logic to determine whether or not to delete an image? I can't immediately see a downside to a DELETE endpoint which accepts a Theme/URI.

This would avoid the large "do it all" method you have here

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
jamiesprax: Which is actually ImageDetails. Incoming Json at first glance suggests raw json string to me

src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -0,0 +1,461 @@
...
+    private void setImageDetails(
+            String mimeType,
+            String imageTheme,
+            String imageHeight,
+            String imageWidth,
+            Image existingImage
+    ) {
+        existingImage.setMimeType(MimeType.valueOf(mimeType.toUpperCase()));
+        existingImage.setType(ImageType.LOGO);
+        existingImage.setColor(ImageColor.MONOCHROME);
```
jamiesprax: Why is this hardcoded? What if it's not?

PR #1252, opened 2017-05-22
added subtitle comparison for encodings

dragospitica:


src/main/java/org/atlasapi/remotesite/bbc/nitro/LastUpdatedSettingContentWriter.java
```diff
@@ -226,7 +226,8 @@ private boolean equal(Encoding encoding, Encoding prevEncoding) {
                 && Objects.equal(encoding.getVideoCoding(), prevEncoding.getVideoCoding())
                 && Objects.equal(encoding.getVideoFrameRate(), prevEncoding.getVideoFrameRate())
                 && Objects.equal(encoding.getVideoProgressiveScan(), prevEncoding.getVideoProgressiveScan())
-                && Objects.equal(encoding.getVideoVerticalSize(), prevEncoding.getVideoVerticalSize());
+                && Objects.equal(encoding.getVideoVerticalSize(), prevEncoding.getVideoVerticalSize())
+                && encoding.getSubtitled().equals(prevEncoding.getSubtitled());
```
jamiesprax: Why didn't you follow the format of Objects.equal(x.y)?
This will still work fine but just a bit strange to not follow what's already here.

jamiesprax: Also these .equals methods are in alphabetical order

PR #1255, opened 2017-05-24
Fix start and end dates not being ingested

jamiesprax:
New logic was added in to update images on channels which caused the start
and end dates for those images to be lost as they were being set on the wrong
level. This retains the dates and keeps the existing specification of not
overwriting existing images if not found in new images.


src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -161,11 +167,12 @@ private Channel createOrMerge(Channel newChannel) {
...
-                existingChannel.setImages(newChannel.getAllImages());
-            } else {
-                updateExistingChannelImages(newChannel, existingChannel);
-            }
+            existingChannel.setImages(
+                    Iterables.isEmpty(existingChannel.getAllImages())
+                            ? newChannel.getAllImages()
+                            : updateImages(newChannel, existingChannel)
+            );
+            existingChannel.setImages(updateImages(newChannel, existingChannel));
```
TjlHope: Doesn't this change mean you'll end up doing the update twice?

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -203,41 +210,37 @@ private Channel createOrMerge(Channel newChannel) {
...
+                        .collect(MoreCollectors.toImmutableMap(
+                                temporalImage -> temporalImage.getValue().getTheme(),
+                                temporalImage -> temporalImage
+                        ));
+
+        return ImmutableSet.<TemporalField<Image>>builder()
+                .addAll(existingImages.values())
+                .addAll(
+                        newImageMap.entrySet().stream()
+                                .filter(entry -> !existingImages.keySet().contains(entry.getKey()))
```
TjlHope: Just
!existingImages.containsKey(entry.getKey())

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -203,41 +210,37 @@ private Channel createOrMerge(Channel newChannel) {
...
-        existingImageValue.setAvailabilityStart(newImageValue.getAvailabilityStart());
-        existingImageValue.setAvailabilityEnd(newImageValue.getAvailabilityEnd());
-        existingImageValue.setWidth(newImageValue.getWidth());
-        existingImageValue.setHeight(newImageValue.getHeight());
-        existingImageValue.setHasTitleArt(newImageValue.hasTitleArt());
+    // We need to update the existing channel images to avoid overwriting all existing images every
+    // time we ingest PA channels. This should go away once we implement channel equivalence.
+    Iterable<TemporalField<Image>> updateImages(Channel newChannel, Channel existingChannel) {
+        Map<ImageTheme, TemporalField<Image>> newImageMap = StreamSupport.stream(
+                newChannel.getAllImages().spliterator(),
```
dragospitica: this will throw an NPE if the new channel doesn't have any images no?

PR #1256, opened 2017-05-25
Sanitise existing channel images

jamiesprax:
Existing images that were ingested without dates persisted and caused issues
as there were duplicate themes for valid dates. This should clean those up
while preserving the logic.


src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -212,38 +205,47 @@ private Channel createOrMerge(Channel newChannel) {
         }
     }
 
+    Iterable<TemporalField<Image>> updateImages(Channel newChannel, Channel existingChannel) {
```
dragospitica: I don't really see the point of this method. Also there'd be two methods with the same name which is not good practice

PR #1260, opened 2017-06-01
Add an exact match check to equiv TitleSubset scorer.

TjlHope:
If there is a name consisting of purely common words, or stripped
characters then titles won't match. In this case if there's an exact
match it would make sense to return a positive score.
This scorer also originally replaced an exact match scorer, so it
simply makes it a complete replacement/improvement of the old scorer.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorer.java
+++ b/src/main/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorer.java
@@ -13,7 +13,6 @@ import org.atlasapi.media.entity.Item;
 import org.atlasapi.persistence.content.ContentResolver;
 
 import com.google.common.base.CharMatcher;
-import com.google.common.base.Function;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


src/test/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorerTest.java
```diff
@@ -45,6 +45,10 @@ public void testMatches() {
                 itemWithTitle("Power - Rangers: R.P.M.!!"),
                 itemWithTitle("Power Rangers RPM")
         ));
+        assertEquals(Score.ONE, score(
```
AdamDM: As this tests for a unique thing it should be it's own test (as should the others originally)

src/main/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorer.java
```diff
@@ -83,52 +82,62 @@ protected String getName() {
...
+    private boolean titleMissing(Content subject) {
+        return Strings.isNullOrEmpty(subject.getTitle());
+    }
+
+    private String sanitize(String title) {
+        return titleTransformer.expand(title).toLowerCase();
+    }
+
+    private boolean titlesEqual(String subjectTitle, String candidateTitle) {
+        return false && subjectTitle.equals(candidateTitle);
```
AdamDM: wouldn't false and something always be false?

TjlHope: Oh bollocks - added that in to test the test something else - completely forgot to remove it - test won't actually pass at the moment - doh!

src/test/java/org/atlasapi/equiv/scorers/TitleSubsetBroadcastItemScorerTest.java
TjlHope: Fair point, I'll split them up

PR #1262, opened 2017-06-07
Update bt channel data

jamiesprax:
This creates a separate channel for each BT environment with a URI dependant on the publisher in the format "http://:key/:id" (where key is publisher key and id is the encoded atlas ID of the base channel.) This will allow channel equivalence to match them to the base channel from the URI.

In this current state, the base channels are updated in the same way as before but this can be removed later once it's confirmed working so the ingester only updates the channels for the BT environment it's running in.


src/main/java/org/atlasapi/remotesite/bt/channels/BtMpxChannelDataIngester.java
```diff
@@ -116,12 +116,14 @@ private void removeOldChannelGroupChannels(Set<String> allCurrentChannelGroups)
...
+        Predicate<ChannelNumbering> shouldKeepPredicate
             = shouldKeepChannelGroupMembership(channelGroupIdsToRemove);
         
         for (Channel channel : channelResolver.all()) {
             int existingSize = channel.getChannelNumbers().size();
-            Iterable<ChannelNumbering> filtered = Iterables.filter(channel.getChannelNumbers(), shouldKeepPredicate);
+            Iterable<ChannelNumbering> filtered = channel.getChannelNumbers().stream()
+                    .filter(shouldKeepPredicate)
+                    .collect(Collectors.toList());
             if (Iterables.size(filtered) != existingSize) {
```
TjlHope: You could always assign it as a Collection (as you're collecting it to a collection, not an iterable, then the size call can be direct, but doesn't really matter...

src/test/java/org/atlasapi/equiv/channel/matchers/BtChannelMatcherTest.java
```diff
@@ -1,58 +1,54 @@
...
-    private BtChannelMatcher btChannelMatcher = BtChannelMatcher.create();
-    private Channel masterChannel;
+    private SubstitutionTableNumberCodec codec = SubstitutionTableNumberCodec.lowerCaseOnly();
+    private BtChannelMatcher btChannelMatcher = BtChannelMatcher.create(Publisher.BT_TV_CHANNELS);
 
-    @Before
-    public void setUp() {
-        masterChannel = channelFor(Publisher.METABROADCAST, "123");
-    }
+    private Channel existingChannel =  channelFor(
```
TjlHope: As Channel isn't immutable, you should probably be keeping it in a @Before method - so you definitely get a clean copy for each test.

src/test/java/org/atlasapi/equiv/channel/matchers/BtChannelMatcherTest.java
```diff
@@ -1,58 +1,54 @@
...
-        Channel nonMatchingAliasChannel = channelFor(Publisher.BT_TV_CHANNELS, "321");
-
-        assertFalse(btChannelMatcher.isAMatch(masterChannel, nonMatchingAliasChannel));
-    }
-
-    @Test
-    public void setBtChannelMatcherDoesNotMatchOnMissingPaIds() throws Exception {
-        Channel missingAliasChannel = channelFor(Publisher.BT_TV_CHANNELS, "");
+        Channel nonMatchingCandidate = channelFor(
+                Publisher.METABROADCAST,
```
TjlHope: should you test the publisher and id separately?

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -79,80 +84,114 @@ public void addAvailableDateToChannel(PaginatedEntries paginatedEntries) {
...
 
-        for(Channel channel : channels) {
+        for (Channel channel : channels) {
             if(!channelIdsThatHaveAliasesAdded.contains(channel.getId())) {
                 channel.setAliases(
-                        Iterables.filter(channel.getAliases(),
-                                not(isAliasWithNamespace(aliasNamespace)))
+                        channel.getAliases()
+                                .stream()
+                                .filter(alias -> !alias.getNamespace().equals(aliasNamespace))
```
jamiesprax: An alias can't exist without a namespace and aliasNamespace has a checkNotNull

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -79,80 +84,114 @@ public void addAvailableDateToChannel(PaginatedEntries paginatedEntries) {
...
 
+        if (advertiseToDate != null && advertiseToDate.getMillis() > 0) {
+            channel.setAdvertiseTo(advertiseToDate);
+        } else {
+            channel.setAdvertiseTo(null);
+        }
+
         channelWriter.createOrUpdate(channel);
-        return Maybe.just(channel);
+        return Optional.of(channel);
```
jamiesprax: Because it's returned immediately again from the method above where an optional is needed

TjlHope: Yes I saw that. I can see why you might want it from a simplistic view, but it just doesn't feel right to me.
Looking at the method signature in isolation it seems like you give it a concrete channel, and you might or might not get one back, which just seems weird. As it's not modifying the channel I would have probably expected it to be a void (indicates it's working through side effects), or return the channel itself (because it wants to chain).
But yes, very minor point, and it's a private method, so don't really care :-)

src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
TjlHope: Fair enough

PR #1263, opened 2017-06-13
Fix/channel image upload endpoint

dragospitica:


src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -47,20 +48,23 @@
...
     private final AtlasModelWriter<Iterable<Channel>> outputter;
+    private final NumberToShortStringCodec codec;
 
     private ChannelWriteExecutor(Builder builder) {
         this.appConfigFetcher = checkNotNull(builder.appConfigFetcher);
         this.store = checkNotNull(builder.store);
         this.reader = checkNotNull(builder.reader);
         this.channelTransformer = checkNotNull(builder.channelTransformer);
         this.outputter = checkNotNull(builder.outputter);
+        this.codec = SubstitutionTableNumberCodec.lowerCaseOnly();
```
jamiesprax: where is this used?

dragospitica: @jamiesprax line 105 is one example. I'm creating a new one in here because before I was passing an existing one from the controller, but that was handling owl IDs rather than Deer IDs.

PR #1273, opened 2017-07-04
Alias equiv for barb

AdamDM:


src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -119,6 +120,10 @@ public BarbAliasEquivalenceGenerator(
                 desc
         );
 
+        if (!(subject.getAliasUrls().size() == 0)) {
```
TjlHope: Use `.isEmpty()` instead of `.size == 0`

jamiesprax: .isEmpty() ?

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -145,6 +150,50 @@ public BarbAliasEquivalenceGenerator(
...
+        desc.startStage("Resolving Barb Aliases:");
+        subject.getAliasUrls().forEach(alias -> desc.appendText(alias));
+        desc.finishStage();
+
+        Set<Iterable<LookupEntry>> entriesSet = subject.getAliases().stream().map(alias -> {
+            Iterable<LookupEntry> entries = lookupEntryStore.entriesForAliases(
+                    Optional.of(alias.getNamespace()),
+                    ImmutableSet.of(alias.getValue())
+            );
+            return entries;
```
TjlHope: you can just do a straight return rather than assigning to an intermediate variable (and then you don't need the curlys)

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -145,6 +150,50 @@ public BarbAliasEquivalenceGenerator(
...
+        }).collect(Collectors.toSet());
+
+        entriesSet.forEach(iterableLookupEntry -> {
+            iterableLookupEntry.forEach(entry -> {
+                Identified identified = resolver.findByCanonicalUris(
+                        ImmutableSet.of(entry.uri())
+                ).getFirstValue().requireValue();
+
+                if (identified.getAliasUrls().size() !=0) {
```
jamiesprax: isEmpty()

TjlHope: `isEmpty()`

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -145,6 +150,50 @@ public BarbAliasEquivalenceGenerator(
...
+                    boolean match = false;
+
+                    for (String alias : identified.getAliasUrls()) {
+                        if (subject.getAliasUrls().contains(alias)) {
+                            match = true;
```
TjlHope: `break;`

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorTest.java
```diff
@@ -2,38 +2,113 @@
...
+    private BarbAliasEquivalenceGenerator aliasGenerator;
+    private final ContentResolver aliasResolver = mock(ContentResolver.class);
+    private final MongoLookupEntryStore aliasLookupEntryStore = mock(MongoLookupEntryStore.class);
+
+    Content aliasIdentified1;
+    Content aliasIdentified2;
+
     public BarbAliasEquivalenceGeneratorTest() {
 
+        setupForceEquivalenceTests();
```
TjlHope: setup should always be done in a `@Before public void setUp()` method instead of the constructor to ensure you have a clean slate for each test

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorTest.java
```diff
@@ -59,6 +134,10 @@ public BarbAliasEquivalenceGeneratorTest() {
         generator = new BarbAliasEquivalenceGenerator(lookupEntryStore, resolver);
     }
 
+    private ImmutableSet getIdentifiedImmutableSet(LookupEntry lookupEntry) {
```
TjlHope: Is it really worth it when you can just use `.of` ?

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -145,6 +150,50 @@ public BarbAliasEquivalenceGenerator(
...
+            DefaultScoredCandidates.Builder equivalents,
+            ResultDescription desc
+    ) {
+
+        desc.startStage("Resolving Barb Aliases:");
+        subject.getAliasUrls().forEach(alias -> desc.appendText(alias));
+        desc.finishStage();
+
+        Set<Iterable<LookupEntry>> entriesSet = subject.getAliases().stream().map(alias -> {
+            Iterable<LookupEntry> entries = lookupEntryStore.entriesForAliases(
```
jamiesprax: You shouldn't need to assign and return a value, just ".map(alias -> method call)"

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorTest.java
```diff
@@ -104,4 +183,38 @@ public ResultDescription finishStage() {
 
         assertEquals(scoredCandidates, equivalents.build());
     }
-}
\ No newline at end of file
+
+    @Test
+    public void aliasGeneratorFindsByAlias() {
+
+        ResultDescription desc = new ResultDescription() {
```
TjlHope: Is it worth defineing this as a separate NullResultDescription class that can be used by other tests?

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -145,6 +150,50 @@ public BarbAliasEquivalenceGenerator(
...
+
+                if (identified.getAliasUrls().size() !=0) {
+                    boolean match = false;
+
+                    for (String alias : identified.getAliasUrls()) {
+                        if (subject.getAliasUrls().contains(alias)) {
+                            match = true;
+                        }
+                    }
+
```
jamiesprax: It looks like you could stream the lookupEntries, map them to identified, filter by aliasUrls.isEmpty, and then do the bit at the end to make this cleaner

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorTest.java
```diff
@@ -104,4 +183,38 @@ public ResultDescription finishStage() {
...
+            public ResultDescription finishStage() {
+                return null;
+            }
+        };
+
+        ScoredCandidates scoredCandidates = aliasGenerator.generate(aliasIdentified2, desc);
+
+        for (Object scoredCandidate : scoredCandidates.candidates().keySet()) {
+            Object content = scoredCandidates.candidates().get(scoredCandidate);
+            if (content instanceof Content) {
```
jamiesprax: just make this an assertion to clean up the else part

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorTest.java
AdamDM: I found this online - it gets past a weird thing where mockito doesn't like returning generics

PR #1280, opened 2017-07-12
Bbc nitro to telescope

alianos-:
The structure of this change is fairly simple.
There is a new class that Proxies with telescopeClient. There is a factory that makes proxies based on the name of the ingester. Whoever wants to report to telescope needs to:

1. Update the factory to make proxies for the new ingester (i.e. add a name).
2. Gets the right proxy through the factory
3. proxy.startReporting(), proxy.reportEvent(), proxy.endReporting().
4. victory dance.

=============== READ BELOW PLEASE ===========
Please note that ATLAS now lists as a dependency a copy of the telescope client that shades the Jackson library (so it will not conflict with ATLAS's versions of Jackson). This is 1.0-SNAPSHOT-SHADED-0.3 and will have to be manually updated if we update telescope.

Also at the moment these libs come from a stripped down version of telescope. That way we did not have to mess with the telescope server, but this means every time we update telescope, we have to manually rebuild for the sole purpose of recreating the versions that shade Jackson.

I put this here to bring it to everyone's attention. Ask me for more details.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
@@ -126,19 +126,24 @@ public class ContentUpdatingNitroBroadcastHandler implements NitroBroadcastHandl
 
                 addBroadcast(item, versionUri(nitroBroadcast), broadcast.get());
 
+                //To report to telescope, we need atlasIds (which are created with the below codec.
+                NumberToShortStringCodec idCodec = SubstitutionTableNumberCodec.lowerCaseOnly();
+
                 Brand brand = getBrand(item, brandIndex);
                 if (brand != null) {
                     writer.createOrUpdate(brand);
+                    String atlasId = idCodec.encode(BigInteger.valueOf(brand.getId()));
+                    telescope.reportSuccessfulEvent(atlasId, TelescopeHelperMethods.getAliases(brand.getAliases()), nitroBroadcast);
                 }
 
                 Series sery = getSeries(item, seriesIndex);
                 if (sery != null) {
                     writer.createOrUpdate(sery);
+                    String atlasId = idCodec.encode(BigInteger.valueOf(sery.getId()));
+                    telescope.reportSuccessfulEvent(atlasId, TelescopeHelperMethods.getAliases(sery.getAliases()), nitroBroadcast);
                 }
-                item  = writer.createOrUpdate(item);
 
-                //Report to telescope
-                NumberToShortStringCodec idCodec = new SubstitutionTableNumberCodec();
+                writer.createOrUpdate(item);
                 String atlasId = idCodec.encode(BigInteger.valueOf(item.getId()));
                 telescope.reportSuccessfulEvent(atlasId, TelescopeHelperMethods.getAliases(item.getAliases()), nitroBroadcast);
 
--- a/src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
+++ b/src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
@@ -4,19 +4,32 @@ import com.google.common.base.Function;
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
 import com.google.common.base.Predicates;
-import com.google.common.collect.*;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.metabroadcast.atlas.glycerin.model.PidReference;
 import com.metabroadcast.common.ids.NumberToShortStringCodec;
 import com.metabroadcast.common.ids.SubstitutionTableNumberCodec;
-import org.atlasapi.media.entity.*;
+import org.atlasapi.media.entity.Brand;
+import org.atlasapi.media.entity.Broadcast;
+import org.atlasapi.media.entity.Container;
+import org.atlasapi.media.entity.Episode;
+import org.atlasapi.media.entity.Identified;
+import org.atlasapi.media.entity.Item;
+import org.atlasapi.media.entity.ParentRef;
 import org.atlasapi.media.entity.ScheduleEntry.ItemRefAndBroadcast;
+import org.atlasapi.media.entity.Series;
+import org.atlasapi.media.entity.Version;
 import org.atlasapi.persistence.content.ContentResolver;
 import org.atlasapi.persistence.content.ContentWriter;
 import org.atlasapi.remotesite.bbc.BbcFeeds;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroBroadcastExtractor;
 import org.atlasapi.remotesite.bbc.nitro.extract.NitroUtil;
-import org.atlasapi.telescope.TelescopeHelperMethods;
 import org.atlasapi.telescope.TelescopeProxy;
+import org.atlasapi.telescope.TelescopeUtilityMethods;
 import org.atlasapi.util.GroupLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -24,7 +37,9 @@ import org.slf4j.LoggerFactory;
 import java.math.BigInteger;
 import java.util.Set;
 
-import static com.google.common.base.Preconditions.*;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 
 /**
  * {@link NitroBroadcastHandler} which fetches, updates and writes relevant
@@ -38,8 +53,7 @@ public class ContentUpdatingNitroBroadcastHandler implements NitroBroadcastHandl
     private final LocalOrRemoteNitroFetcher localOrRemoteFetcher;
     private final GroupLock<String> lock;
 
-    private final NitroBroadcastExtractor broadcastExtractor
-            = new NitroBroadcastExtractor();
+    private final NitroBroadcastExtractor broadcastExtractor = new NitroBroadcastExtractor();
 
 
     public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWriter writer,
@@ -135,7 +149,7 @@ public class ContentUpdatingNitroBroadcastHandler implements NitroBroadcastHandl
                     //report to telescope
                     if (brand.getId() != null) {
                         String atlasId = idCodec.encode(BigInteger.valueOf(brand.getId()));
-                        telescope.reportSuccessfulEvent(atlasId, TelescopeHelperMethods.getAliases(brand.getAliases()), nitroBroadcast);
+                        telescope.reportSuccessfulEvent(atlasId, TelescopeUtilityMethods.getAliases(brand.getAliases()), nitroBroadcast);
                     } else {
                         telescope.reportFailedEvent("BBC Nitro ingester failed to write brand in the DB", nitroBroadcast);
                     }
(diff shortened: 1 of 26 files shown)


src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,153 @@
...
+
+    /**
+     * The client always reports to {@link TelescopeConfiguration#TELESCOPE_HOST}
+     */
+    TelescopeProxy(Process process) {
+        this.process = process;
+        //get a client
+        TelescopeClientImpl client = TelescopeClientImpl.create(TelescopeConfiguration.TELESCOPE_HOST);
+        if (client == null) {
+            throw new NullPointerException("Could now get a TelescopeClientImpl object with the given TELESCOPE_HOST:" + TelescopeConfiguration.TELESCOPE_HOST);
```
alianos-: That is certainly not best practice I am not sure what to do. I doesn't make sense to return the object if it cannot report. Also I don't know or if this call can return null at all.

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,153 @@
...
+                            .withRawMime(MimeType.APPLICATION_JSON.toString())
+                            .build()
+                    )
+                    .withTaskId(taskId)
+                    .withTimestamp(LocalDateTime.now())
+                    .build();
+
+            telescopeClient.createEvents(ImmutableList.of(reportEvent));
+
+            log.info("Reported successfully event with taskId {}", reportEvent.getTaskId().get());
```
alianos-: These log line is meaningless. Add atlasId and leave it here or move it to log.debug()?

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,153 @@
...
+                            .withRaw(objectMapper.writeValueAsString(objectToSerialise))
+                            .withRawMime(MimeType.APPLICATION_JSON.toString())
+                            .build()
+                    )
+                    .withTaskId(taskId)
+                    .withTimestamp(LocalDateTime.now())
+                    .build();
+            telescopeClient.createEvents(ImmutableList.of(reportEvent));
+
+            log.info("Reported succeffully FAILED event with taskId {}", reportEvent.getTaskId().get());
```
alianos-: This log line is also meaningless?

alianos-: and I just noticed (@dragos) that you use .get() in the optional without checking it first.

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
alianos-: I will replace that with logging and a make the object do nothing.

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
alianos-: @dragospitica also you used .get() on optional directly. Should be .orElse() imho.

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -45,42 +33,42 @@
...
-    
+
     private final NitroBroadcastExtractor broadcastExtractor
-        = new NitroBroadcastExtractor();
+            = new NitroBroadcastExtractor();
 
 
     public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWriter writer,
-            LocalOrRemoteNitroFetcher localOrRemoteNitroFetcher, GroupLock<String> lock) {
+                                                LocalOrRemoteNitroFetcher localOrRemoteNitroFetcher, GroupLock<String> lock) {
```
AdamDM: not sure this amount of space is correct - you may want to check your intellij settings as it seems to be reformatting everything with new spaces, which is also making the PR difficult to read

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -45,42 +33,42 @@
...
             ImmutableSet<Container> resolvedSeries = localOrRemoteFetcher.resolveOrFetchSeries(items.getAll());
             ImmutableSet<Container> resolvedBrands = localOrRemoteFetcher.resolveOrFetchBrand(items.getAll());
-            
+
             Iterable<Series> series = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Series.class);
             Iterable<Brand> brands = Iterables.filter(Iterables.concat(resolvedSeries, resolvedBrands), Brand.class);
-            
-            return writeContent(nitroBroadcasts, items, series, brands);
+
+            return writeContent(nitroBroadcasts, items, series, brands, telescope);
```
AdamDM: are you sure it's necessary to send telescope to this and then return it without using it?

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -92,65 +80,92 @@ public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWri
     private Set<String> itemIds(
             Iterable<com.metabroadcast.atlas.glycerin.model.Broadcast> nitroBroadcasts) {
         return ImmutableSet.copyOf(Iterables.transform(nitroBroadcasts,
-            new Function<com.metabroadcast.atlas.glycerin.model.Broadcast, String>() {
-                @Override
-                public String apply(com.metabroadcast.atlas.glycerin.model.Broadcast input) {
-                    return NitroUtil.programmePid(input).getPid();
+                new Function<com.metabroadcast.atlas.glycerin.model.Broadcast, String>() {
```
AdamDM: you should be able to use a stream style lambda here rather than a Function

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -92,65 +80,92 @@ public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWri
...
+                //To report to telescope, we need atlasIds (which are created with the below codec.
+                NumberToShortStringCodec idCodec = SubstitutionTableNumberCodec.lowerCaseOnly();
+
                 Brand brand = getBrand(item, brandIndex);
                 if (brand != null) {
                     writer.createOrUpdate(brand);
+                    //report to telescope
+                    if (brand.getId() != null) {
+                        String atlasId = idCodec.encode(BigInteger.valueOf(brand.getId()));
+                        telescope.reportSuccessfulEvent(atlasId, TelescopeHelperMethods.getAliases(brand.getAliases()), nitroBroadcast);
```
AdamDM: possible line length problem - keep it under 100 characters
On that note have you imported the intellij settings from the engineering repo (I think they're there) - that'll flag a lot of things for our code style, and give you a ruler for 100 characters to remind you not to go over it

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -92,65 +80,92 @@ public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWri
...
+                    } else {
+                        telescope.reportFailedEvent("BBC Nitro ingester failed to write brand in the DB", nitroBroadcast);
```
AdamDM: you're using telescope in a try block - so you'll also want a catch block to report to telescope that something failed - otherwise it won't get reported

AdamDM: ignore this comment ^ I found it

src/main/java/org/atlasapi/telescope/TelescopeConfiguration.java
```diff
@@ -0,0 +1,17 @@
+package org.atlasapi.telescope;
+
+import com.metabroadcast.common.properties.Configurer;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * TODO: This file is probably generic for atlas, and does not need to be in BBC nitro.
```
AdamDM: add jira tickets for TODOs (if you have a jira account setup yet) and put a link to the ticket in the comment

src/main/java/org/atlasapi/telescope/TelescopeFactory.java
```diff
@@ -0,0 +1,74 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Creates proxies to telescopeClients that can be used for reporting to telescope.
+ * <p>
+ * If you need to extend this class to accommodate more Processes (i.e. add more owl ingesters),
+ * extend the {@link IngesterName} enum accordingly.
+ * <p>
+ * Created by andreas on 07/07/2017.
```
AdamDM: remove created by - we use git blame

src/main/java/org/atlasapi/telescope/TelescopeFactory.java
```diff
@@ -0,0 +1,74 @@
...
+
+
+    //create and return a telescope.api.Process.
+    private static Process getProcess(IngesterName name) {
+        Environment environment;
+        try {
+            environment = Environment.valueOf(TelescopeConfiguration.ENVIRONMENT);
+        } catch (IllegalArgumentException e) {
+            //add stage as the default environment, which is better than crashing
+            log.error("Could not find a telescope environment with the given name, name={}. Falling back to STAGE.", TelescopeConfiguration.ENVIRONMENT);
```
AdamDM: this shouldn't be necessary and if people aren't watching the logs carefully enough this could cause an ingest to stage without someone realising

src/main/java/org/atlasapi/telescope/TelescopeFactory.java
```diff
@@ -0,0 +1,74 @@
...
+            e.printStackTrace();
```
jamiesprax: add the error to the end of the log.error(). We don't print stack traces

src/main/java/org/atlasapi/telescope/TelescopeHelperMethods.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @author andreas
+ */
+public class TelescopeHelperMethods {
+     public static ImmutableList<Alias> getAliases(Set<org.atlasapi.media.entity.Alias> aliases) {
+        return aliases.stream()
+                .map(alias -> Alias.create(alias.getNamespace(), alias.getValue()))
+                .collect(MoreCollectors.toImmutableList());
+    }
+}
```
jamiesprax: It's in our style guide (#29) to avoid using static helper methods

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,175 @@
+package org.atlasapi.telescope;
+
+import com.google.common.collect.ImmutableList;
+import com.metabroadcast.columbus.telescope.api.*;
```
jamiesprax: We don't use * imports either

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * The client always reports to {@link TelescopeConfiguration#TELESCOPE_HOST}
+     */
+    TelescopeProxy(Process process) {
+        this.process = process;
+
+        //get a client
+        try {
+            //might throw RuntimeException: java.net.UnknownHostException
```
jamiesprax: We don't really need all these comments in here. The method names should be self describing. We generally only use comments when some weird workaround has been used and needs documenting

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -1,42 +1,30 @@
...
-import org.atlasapi.media.entity.ParentRef;
+import com.google.common.base.Function;
+import com.google.common.base.Objects;
+import com.google.common.base.Optional;
+import com.google.common.base.Predicates;
+import com.google.common.collect.*;
+import com.metabroadcast.atlas.glycerin.model.PidReference;
+import com.metabroadcast.common.ids.NumberToShortStringCodec;
+import com.metabroadcast.common.ids.SubstitutionTableNumberCodec;
+import org.atlasapi.media.entity.*;
```
jamiesprax: remove * imports

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -1,42 +1,30 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
+import java.math.BigInteger;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.*;
```
jamiesprax: *

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
alianos-: Good find. That was intelliJ optimising imports, I don't like wildcard imports either.

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,231 @@
...
+    private String taskId;
+    private Process process;
+    private ObjectMapper objectMapper;
+    private boolean startedReporting = false; //safeguard flags
+    private boolean stoppedReporting = false;
+
+    /**
+     * The client always reports to {@link TelescopeConfiguration#TELESCOPE_HOST}
+     */
+    TelescopeProxy(Process process) {
```
dragospitica: Read this if you didn't already `https://docs.metabroadcast.com/display/mbst/Java+Style+Guide` #38

src/main/java/org/atlasapi/telescope/TelescopeProxy.java
```diff
@@ -0,0 +1,231 @@
...
+     * useless.
+     */
+    public void endReporting() {
+        if (!initialized()) {
+            return;
+        }
+        if (startedReporting) {
+            telescopeClient.endIngest(taskId);
+            stoppedReporting = true;
+            log.debug("Finished reporting to Telescope (taskId:)", taskId);
```
dragospitica: I think you missed a `{}`

src/main/resources/config/environment.properties
```diff
@@ -22,6 +22,9 @@ mongo.db.tag.fallback=
 #string whose valid values are the enums listed http://api.mongodb.org/java/2.6/com/mongodb/WriteConcern.html
 processing.mongo.writeConcern=
 
+# TELESCOPE SERVER
+telescope.host=columbus-telescope.stage.svc.cluster.local
+telescope.environment=STAGE
```
dragospitica: You need to leave these 2 ^ empty and override them in the k8 deployment. Don't forget to add them in the Dockerfile as well

PR #1286, opened 2017-08-10
Equiv to telescope

AdamDM:


src/main/java/org/atlasapi/equiv/update/MultipleSourceEquivalenceUpdater.java
```diff
@@ -34,8 +35,8 @@ public MultipleSourceEquivalenceUpdater register(
     }
 
     @Override
-    public boolean updateEquivalences(Content subject) {
-        return updaters.get(subject.getPublisher()).updateEquivalences(subject);
+    public boolean updateEquivalences(Content subject, OwlTelescopeProxy telescopeProxy) {
+        return updaters.get(subject.getPublisher()).updateEquivalences(subject, telescopeProxy);
```
TjlHope: Not your bug, but this is asking for a NPE

src/main/java/org/atlasapi/equiv/update/SourceSpecificEquivalenceUpdater.java
```diff
@@ -85,9 +86,9 @@ private boolean topLevelSeries(Content content) {
     }
 
     private <T> boolean update(EquivalenceUpdater<T> updater,
-            T content) {
+            T content, OwlTelescopeProxy telescopeProxy) {
```
TjlHope: method param formatting

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -44,6 +48,10 @@ public ContentEquivalenceUpdateTask(ContentLister contentLister, ContentResolver
         this.progressStore = progressStore;
         this.updater = RootEquivalenceUpdater.create(contentResolver, updater);
         this.ignored = ignored;
+        this.telescopeProxy = OwlTelescopeProxy.create(
+                OwlTelescopeReporters.EQUIVALENCE,
```
TjlHope: equivalence for publishers? ideally we'd have a list of publishers in the name, but not sure how much it matters...

src/main/java/org/atlasapi/equiv/update/www/ChannelEquivalenceUpdateController.java
```diff
@@ -90,18 +95,31 @@ public void runUpdate(
...
-    private Runnable updateFor(Channel channel) {
+    private Runnable updateFor(Channel channel, OwlTelescopeProxy telescopeProxy) {
         return () -> {
             try {
-                channelUpdater.updateEquivalences(channel);
+                channelUpdater.updateEquivalences(channel, telescopeProxy);
                 log.info("Finished updating {}", channel);
             } catch (Exception e) {
                 log.error("Error updating equivalence for channel: {}", channel, e);
+                telescopeProxy.reportFailedEventWithError(
```
TjlHope: I don't really like the fact that the successful event is reported within the method, but the failed even is reported here.
That said I don't know how clean it would be doing the failed event within the method as you'd probably need a try {...} catch (Exception e) { reportFailed(); throw e; } which is quite ugly...

src/main/java/org/atlasapi/equiv/update/www/ContentEquivalenceUpdateController.java
```diff
@@ -123,13 +136,17 @@ public void runUpdate(
...
-    private Runnable updateFor(final Content content) {
+    private Runnable updateFor(final Content content, OwlTelescopeProxy telescopeProxy) {
         return () -> {
             try {
-                contentUpdater.updateEquivalences(content);
+                contentUpdater.updateEquivalences(content, telescopeProxy);
                 log.info("Finished updating {}", content);
             } catch (Exception e) {
                 log.error(content.toString(), e);
+                telescopeProxy.reportFailedEventWithError(
```
TjlHope: same with success inside, failed outside

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporters.java
```diff
@@ -8,7 +8,10 @@
 public enum OwlTelescopeReporters implements TelescopeReporterName {
     BBC_NITRO_INGEST("bbc-nitro-scheduled-ingester", "BBC Nitro Scheduled Ingester"),
     BBC_NITRO_INGEST_API("bbc-nitro-ingester-api", "BBC Nitro API Ingester"),
-    MOCK_REPORTER("mock-reporter", "Mock reporter") //used by OwlTelescopeProxyMock
+    MOCK_REPORTER("mock-reporter", "Mock reporter"), //used by OwlTelescopeProxyMock
```
TjlHope: TBH I think this should be in the test code, not the main code.

src/test/java/org/atlasapi/equiv/EquivalenceUpdatingWorkerTest.java
```diff
@@ -98,7 +103,7 @@ public void testWorkerThatOnlyUpdatesItemsHandlesResolvingNothing() {
         EntityUpdatedMessage msg = new EntityUpdatedMessage("1", Timestamp.of(1L), eid, "brand", "bbc.co.uk");
         workerThatOnlyUpdatesItems.process(msg);
         
-        verify(updater, never()).updateEquivalences(null);
+        verify(updater, never()).updateEquivalences(null, telescopeProxy);
```
TjlHope: surely this test should be updateEquivalences(any(...), any(...))

src/test/java/org/atlasapi/equiv/channel/updaters/MultipleSourceChannelEquivalenceUpdaterTest.java
```diff
@@ -26,12 +27,14 @@
...
     private EquivalenceUpdater<Channel> equivUpdater3 = mock(EquivalenceUpdater.class);
+    @Mock
+    private OwlTelescopeProxy telescopeProxy = mock(OwlTelescopeProxy.class);
 
     @Before
     public void setUp() {
-        when(equivUpdater1.updateEquivalences(any(Channel.class))).thenReturn(true);
-        when(equivUpdater2.updateEquivalences(any(Channel.class))).thenReturn(true);
-        when(equivUpdater3.updateEquivalences(any(Channel.class))).thenReturn(true);
+        when(equivUpdater1.updateEquivalences(any(Channel.class), any(OwlTelescopeProxy.class))).thenReturn(true);
```
TjlHope: this should use the mocked telescopeProxy rather than any if that's actually what you're expecting to be called

PR #1287, opened 2017-08-15
Telescope support

alianos-:


src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -0,0 +1,140 @@
...
+            log.debug("Reported successfully a FAILED event with taskId={}, error={}", getTaskId(), errorMsg);
+            if (isFinished()) {
+                log.warn("An error was reported to telescope after the telescope client={} has finished reporting.", getTaskId());
+            }
+        } catch (JsonProcessingException e) {
+            log.error("Couldn't convert the given object to a JSON string.", e);
+        }
+    }
+
+}
```
jamiesprax: Looks like a bit of code duplication in this class but I understand it may get quite messy if the small parts are extracted

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporters.java
```diff
@@ -0,0 +1,32 @@
...
+    BBC_NITRO_INGEST_PICKS("bbc-nitro-ingester-picks","Mbst Picks Content Group Updater"),
+    BBC_NITRO_INGEST_API("bbc-nitro-ingester-api", "BBC Nitro API Ingester"),
+    BBC_NITRO_INGEST_CHANNELS("bbc-nitro-ingester-channels", "BBC Nitro Channel Ingester"),
+    ;
+
+    final String reporterKey;
+    final String reporterName;
+
+    OwlTelescopeReporters(String reporterKey, String reporterName) {
+        this.reporterKey = reporterKey;
```
jamiesprax: checkNotNull these?

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
alianos-: @jamiesprax Yes it is annoying, but the error messages are different as is the Event itself of course. Maybe I could ditch the try/catch and do the serialization in a method.

PR #1288, opened 2017-08-15
changed duplicate path

alianos-:


src/main/java/org/atlasapi/remotesite/bbc/nitro/ScheduleDayUpdateController.java
```diff
@@ -43,7 +43,7 @@ public ScheduleDayUpdateController(ChannelResolver resolver, ChannelDayProcessor
         this.processor = processor;
     }
 
-    @RequestMapping(value="/system/bbc/nitro/update/{service}/{date}", method=RequestMethod.POST)
+    @RequestMapping(value="/system/bbc/nitro/update/service/{service}/date/{date}", method=RequestMethod.POST)
```
TjlHope: I get the impression having the `/date/` part before the date is not how a lot of the other endpoints work, but I'm certainly not against it 🙂

PR #1292, opened 2017-08-21
Support new pa flags

jamiesprax:


src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelsIngester.java
```diff
@@ -297,10 +295,7 @@ private Channel processStandaloneChannel(
     }
 
     private Boolean getHighDefinition(@Nullable String format) {
-        if (format != null) {
-            return format.equals(FORMAT_HD);
-        }
-        return false;
+        return format != null && format.equals(FORMAT_HD);
```
TjlHope: `return FORMAT_HD.equals(format);` is fewer bytecode instructions, but now I'm really splitting hairs 😄

PR #1293, opened 2017-08-21
Nitro reporting numerous additions and fixes

alianos-:


src/main/java/org/atlasapi/remotesite/bbc/nitro/channels/ChannelIngestTask.java
```diff
@@ -91,10 +95,12 @@ protected void runTask() {
             writeAndMergeChannels(withUris.build(), telescope);
 
             reportStatus(progress.toString());
-            telescope.endReporting();
         } catch (GlycerinException e) {
             throw Throwables.propagate(e);
         }
+        finally{
```
TjlHope: `    } finally {`

src/main/java/org/atlasapi/remotesite/metabroadcast/picks/PicksDayUpdater.java
```diff
@@ -134,12 +136,12 @@ private void addPicksToContentGroup(Iterable<Item> items, OwlTelescopeReporter t
                 telescope.reportSuccessfulEvent(
                         contentGroup.getId(),
                         contentGroup.getAliases(),
-                        contentGroup
+                        contentGroup, items
```
TjlHope: new line?

src/main/java/org/atlasapi/remotesite/metabroadcast/picks/PicksDayUpdater.java
```diff
@@ -134,12 +136,12 @@ private void addPicksToContentGroup(Iterable<Item> items, OwlTelescopeReporter t
...
                 );
             } else {
                 telescope.reportFailedEvent(
-                        "Atlas did not return an id after attempting to create or update this Brand",
-                        contentGroup
+                        "There was an error while trying to write this ContentGroup to Atlas.",
+                        contentGroup, items
```
TjlHope: new line

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -24,7 +26,12 @@
 public class OwlTelescopeReporter extends TelescopeReporter {
 
     private static final Logger log = LoggerFactory.getLogger(OwlTelescopeReporter.class);
-    private static final ObjectMapper objectMapper = new ObjectMapper();
+
+    private static final ObjectMapper objectMapper = new ObjectMapper()
+            .setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY)
+            .setVisibility(PropertyAccessor.CREATOR, JsonAutoDetect.Visibility.NONE)
```
TjlHope: It's probably worth setting `disable(AUTO_DETECT_GETTER)` as well

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -37,23 +44,15 @@ public static OwlTelescopeReporter create(TelescopeReporterName reporterName, Ev
        return new OwlTelescopeReporter(reporterName, eventType);
     }
 
-    private void reportSuccessfulEventGeneric(String atlasItemId, List<Alias> aliases, String warningMsg, Object objectToSerialise){
+    private void reportSuccessfulEventGeneric(String atlasItemId, List<Alias> aliases, String warningMsg, String payload){
```
TjlHope: for consistency this should probably be split onto a line per param

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -37,23 +44,15 @@ public static OwlTelescopeReporter create(TelescopeReporterName reporterName, Ev
...
-            serialized = objectMapper.writeValueAsString(objectToSerialise);
-        } catch (JsonProcessingException e) {
-            warningMsg = "Object could not be serialized";
-        }
-
         EntityState.Builder entityState = EntityState.builder()
                 .withAtlasId(atlasItemId)
-                .withRaw(serialized)
+                .withRaw(payload)
                 .withRawMime(MimeType.APPLICATION_JSON.toString());
```
TjlHope: Is there any scope for making this method more generic? Other owl ingesters use, for example, xml.
Or Maybe someone does that change when they report on that ingester I guess...

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -113,12 +112,22 @@ public void reportFailedEvent(String errorMsg, Object objectToSerialise) {
...
-        try {
-            return objectMapper.writeValueAsString(objectToSerialise);
-        } catch (JsonProcessingException e) {
-            return "Couldn't convert the given object to a JSON string. " + e.getMessage();
+    public static String serialize(Object... objectToSerialise) {
+
+        StringBuilder sb = new StringBuilder().append("{");
+        boolean first = true; //to comma separate more than 1 fields.
+        for (Object o : objectToSerialise) { //one by one, so what can be serialized will be serialized.
+            sb.append(first ? "" : ",");
```
TjlHope: better to do `if (!first) sb.append(',');`

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
```diff
@@ -113,12 +112,22 @@ public void reportFailedEvent(String errorMsg, Object objectToSerialise) {
...
+            sb.append("\"").append(o.getClass().getSimpleName()).append("\":");
```
TjlHope: If you have multiple objects of the same class then this won't be valid JSON (it's a JSON object, so fields must have unique names)

src/test/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroEpisodeExtractorTest.java
```diff
@@ -12,6 +12,7 @@
 import org.atlasapi.media.entity.MediaType;
 import org.atlasapi.media.entity.Restriction;
 import org.atlasapi.persistence.content.people.QueuingPersonWriter;
+import org.atlasapi.reporting.telescope.OwlTelescopeReporter;
```
TjlHope: import but not used?

src/main/java/org/atlasapi/reporting/telescope/OwlTelescopeReporter.java
alianos-: @TjlHope YV also reports in XML but YV has a different implementation classes in atlas-feeds so they are written to support XML. There are multiple things we can do, but in any case you can always construct the event on the spot straight from the builder. We could also write convenience methods. We could give it a default field like with the eventType (ingest, upload, etc). I'll leave this pending atm though.

PR #1307, opened 2017-09-12
Get remotesite.youview working with multiple aliases/channels per service ID

TjlHope:
This changes things to use multimaps instead of plain maps, and passes
the channel around, instead of relying on resolving it from the service
ID.


src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -71,150 +72,105 @@ public static DefaultYouViewChannelResolver create(
...
+        return resolvedChannels.build();
     }
 
     private void buildEntriesForPrefix(
             String prefix,
-            Map<Integer, Channel> channelMap,
-            Map<Integer, String> aliasMap
+            ResolvedChannels.Builder resolvedChannels
     ) {
         Pattern pattern = Pattern.compile("^" + prefix + "(\\d+)$");
```
AdamDM: while you're in here, add a comment above the pattern explaining what the regex is trying to match in case someone has to work it out next time they come in here

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -71,150 +72,105 @@ public static DefaultYouViewChannelResolver create(
...
-
-    private void addService(
-            Pattern pattern,
-            String alias,
-            Channel channel,
-            Map<Integer, Channel> channelMap,
-            Map<Integer, String> aliasMap
-    ) {
+    @Nullable
+    private Integer extractServiceId(Pattern pattern, String alias) {
```
AdamDM: is returning an optional better than a nullable Integer?

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -71,150 +72,105 @@ public static DefaultYouViewChannelResolver create(
...
+            if (resolvedChannels.contains(override.getKey())) {
                 continue;
             }
-            addService(
-                    pattern,
-                    normaliseOverrideAlias(override.getValue()),
-                    override.getKey(),
-                    channelMap,
-                    aliasMap
+            // TODO - should never happen? pass serviceId in to check?
```
AdamDM: don't leave questions in comments, be clearer and more explicit about what's going on
Also don't leave TODO's in code without creating a jira ticket to look into the potential issue, and putting the link to the ticket in the TODO

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewElementProcessor.java
```diff
@@ -43,19 +43,12 @@ public DefaultYouViewElementProcessor(YouViewContentExtractor extractor, Content
...
         final String scheduleEventAliasNamespace = extractor.getScheduleEventAliasNamespace(); 
-        this.scheduleEventAliasPredicate = new Predicate<Alias>() {
-
-            @Override
-            public boolean apply(Alias input) {
-                return input.getNamespace().equals(scheduleEventAliasNamespace);
-            };
-            
-        };
+        this.scheduleEventAliasPredicate = input -> input.getNamespace().equals(scheduleEventAliasNamespace);
```
AdamDM: does this need to be defined or can it be used inline? I can only see it being used once

TjlHope: No idea :-) I was fixing some of the intellij warnings to try and make sure I didn't break anything...
Will check and change if possible.

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: Yeh I meant to get back to all these, missed this one. Will be fixing.

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: questionable as it's private not api - I'll certainly think about it.

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: I'm so used to reading regexs now I don't even think about it :-) Will do

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewElementProcessor.java
TjlHope: No, it is used twice, so think I should probably leave it

PR #1308, opened 2017-09-13
add equivalence jobs and hardcode in bbc1/2 channels

AdamDM:


src/main/java/org/atlasapi/remotesite/bbc/nitro/channels/NitroChannelMap.java
```diff
@@ -0,0 +1,58 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+
+public class NitroChannelMap extends ForwardingMap<String, Channel> {
+
+    private ImmutableBiMap<String, Channel> delegate;
+
+    public NitroChannelMap(ChannelResolver resolver) {
+        this.delegate = ImmutableBiMap.<String, Channel>builder()
+                .put("BBC One Yorkshire", resolve(resolver, "http://www.bbc.co.uk/services/bbcone/yorkshire"))
```
AdamDM: There was no easy way to maintain line length here - it would have looked even more messy
Since they're hardcoded values anyway I'll leave it like this

Other examples I found also used hardcoded channel values, so I believe this is the correct way of doing it

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -683,8 +695,19 @@ private EquivalenceUpdatingWorker equivUpdatingWorker() {
...
+                                        ITV,
+                                        FIVE,
+                                        BBC_REDUX,
+                                        ITUNES,
+                                        RADIO_TIMES,
+                                        LOVEFILM,
+                                        TALK_TALK,
+                                        YOUVIEW,
+                                        NETFLIX,
+                                        BBC_NITRO
```
TjlHope: Why have you only added it to the container part of the or?

src/main/java/org/atlasapi/remotesite/bbc/nitro/channels/NitroChannelMap.java
```diff
@@ -0,0 +1,58 @@
...
+                .put("BBC One HD", resolve(resolver, "http://www.bbc.co.uk/services/bbcone/hd"))
+                .put("BBC One Wales HD", resolve(resolver, "http://ref.atlasapi.org/channels/pressassociation.com/1778"))
+                .put("BBC One Wales", resolve(resolver, "http://www.bbc.co.uk/services/bbcone/wales"))
+                .put("BBC One Yorkshire & Lincolnshire", resolve(resolver, "http://www.bbc.co.uk/services/bbcone/east_yorkshire"))
+                .put("BBC Two Northern Ireland", resolve(resolver, "http://www.bbc.co.uk/services/bbctwo/ni"))
+                .put("BBC Two Scotland", resolve(resolver, "http://www.bbc.co.uk/services/bbctwo/scotland"))
+                .put("BBC Two Wales", resolve(resolver, "http://www.bbc.co.uk/services/bbctwo/wales"))
+                .put("BBC Two HD", resolve(resolver, "http://ref.atlasapi.org/channels/pressassociation.com/1782"))
+                .put("BBC Two England", resolve(resolver, "http://www.bbc.co.uk/services/bbctwo/england"))
+                .build();
```
TjlHope: This seems a strange way of getting the channels. Is there no way of resolving them from the DB, or is there an enum or similar that defines them?
Also, why are some uris `ref.atlasapi.org` whereas others are `www.bbc.co.uk` ?

PR #1309, opened 2017-09-14
handle missing equivalence summaries better

AdamDM:


src/main/java/org/atlasapi/equiv/handlers/ContainerSummaryRequiredException.java
```diff
@@ -0,0 +1,15 @@
+package org.atlasapi.equiv.handlers;
+
+import org.atlasapi.media.entity.Item;
+
+public class ContainerSummaryRequiredException extends RuntimeException {
+
+    public ContainerSummaryRequiredException(String message) {
```
TjlHope: I don't think you want a simple String constructor,
Probably want an ID for the container that's causing the problem.
Or just have the Item one as below.

src/main/java/org/atlasapi/equiv/handlers/EpisodeFilteringEquivalenceResultHandler.java
```diff
@@ -76,7 +76,7 @@ public boolean handle(EquivalenceResult<Item> result) {
         
         if (!possibleSummary.isPresent()) {
             desc.appendText("Item Container summary not found").finishStage();
-            return false;
+            throw new ContainerSummaryRequiredException(result.subject());
```
TjlHope: You might want to be catching and re-throwing this somewhere up the chain so the equiv `desc` can be written?
Not sure if that's even possible though...

src/main/java/org/atlasapi/equiv/EquivalenceUpdatingWorker.java
```diff
@@ -72,7 +73,11 @@ public void process(EntityUpdatedMessage message) {
             log.debug("{} updating equivalence: {} {} {}", 
                 new Object[]{message.getMessageId(), 
                     message.getEntitySource(), message.getEntityType(), eid});
-            equivUpdater.updateEquivalences(content, telescope);
+            try {
+                equivUpdater.updateEquivalences(content, telescope);
+            } catch (ContainerSummaryRequiredException e) {
+                log.error(e.getMessage());
```
TjlHope: obv the next step would be to actually rerun equiv on the container, then re-try the item...

AdamDM: how?!

src/main/java/org/atlasapi/equiv/handlers/ContainerSummaryRequiredException.java
```diff
@@ -0,0 +1,11 @@
+package org.atlasapi.equiv.handlers;
+
+import org.atlasapi.media.entity.Item;
+
+public class ContainerSummaryRequiredException extends RuntimeException {
+
```
TjlHope: You should probably store the item as a field, and give it a getter.
Either that or just store the container Id as a field and give that a getter.
Because when we come to re-run equiv on the container then the item, it's good to know exactly what's needed, rather than having to work it out from the Exception class/message

src/main/java/org/atlasapi/equiv/EquivalenceUpdatingWorker.java
TjlHope: Something like:
```
} catch (ContainerSummaryRequiredExcpetion e) {
  equivUpdater.updateEquivalences(resolve(content.getContainer().getId()), telescope);
  equivUpdater.updateEquivalences(content, telescope);
}
```
Though I'd probably prefer to get the "things to re-run" from the exception, rather than working it out based on the type of exception.

src/test/java/org/atlasapi/equiv/handlers/EpisodeFilteringEquivalenceResultHandlerTest.java
```diff
@@ -323,30 +323,6 @@ public void whenContainerIsNullReturnFalseWhenDelegateReturnsFalse() {
...
-
-        EquivalenceResultHandler<Item> handler = EpisodeFilteringEquivalenceResultHandler.relaxed(
-                delegate,
-                summaryStore
-        );
-
-        assertThat(
-                handler.handle(result),
-                is(false)
-        );
```
TjlHope: Instead of getting rid of the test:
```
try {
  handler.handle(result);
  fail("more descriptive message saying it failed because it didn't throw");
} catch (Container...Exception e) {
  assertThat(e.getItem(), is(result.getSubject()));
}
```

src/main/java/org/atlasapi/equiv/EquivalenceUpdatingWorker.java
```diff
@@ -72,7 +73,24 @@ public void process(EntityUpdatedMessage message) {
             log.debug("{} updating equivalence: {} {} {}", 
                 new Object[]{message.getMessageId(), 
                     message.getEntitySource(), message.getEntityType(), eid});
-            equivUpdater.updateEquivalences(content, telescope);
+            try {
+                equivUpdater.updateEquivalences(content, telescope);
+            } catch (ContainerSummaryRequiredException e) {
+                log.trace(e.getMessage());
```
TjlHope: It's probably worth having something like: `log.trace("ContainerSummaryRequired, attempting to run", e)`
So it's a bit more descriptive, and you get the full stack trace when debugging.

src/main/java/org/atlasapi/equiv/EquivalenceUpdatingWorker.java
```diff
@@ -72,7 +73,24 @@ public void process(EntityUpdatedMessage message) {
...
+                Maybe<Identified> maybeContainer = contentResolver.findByCanonicalUris(
+                        ImmutableSet.of(e.getItem()
+                                .getContainer()
+                                .getUri())).getFirstValue();
+                if (maybeContainer.hasValue()) {
+                    Identified container = maybeContainer.requireValue();
+                    equivUpdater.updateEquivalences((Content) container, telescope);
```
TjlHope: Probably worth running both of these in a try-catch as they could potentially throw said exception as well.
Don't necessarily need to do anything about it. other than log and potentially report an error to telescope, but that's probably better than just letting it propagate.

src/main/java/org/atlasapi/equiv/handlers/ContainerSummaryRequiredException.java
```diff
@@ -0,0 +1,20 @@
...
+import org.atlasapi.media.entity.Container;
+import org.atlasapi.media.entity.Item;
+
+public class ContainerSummaryRequiredException extends RuntimeException {
+
+    private Item item;
+
+    public ContainerSummaryRequiredException(Item item) {
+        super("Container summary not found so equivalence is being rerun on container with id: "
```
TjlHope: It's not *being* re-run, it needs to be rerun

PR #1311, opened 2017-09-19
Rework remotesite.youview multi alias/channel to lock things together

TjlHope:
Changes multiple serviceId->Channel, alias->Channel, serviceId->alias
multimaps to a single ServiceId<int,Channel>->Channel multimap so that
the service ID (int), alias, and alias prefix can all be held together
to prevent things ending up on the wrong channel.


src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -239,89 +213,140 @@ private String normaliseOverrideAlias(String alias) {
...
+        private final String prefix;
+        private final String alias;
+        private ImmutableServiceId(int id, String prefix, String alias) {
+            this.id = id;
+            this.prefix = checkNotNull(prefix);
+            this.alias = checkNotNull(alias);
+        }
+        @Nullable
+        public static ImmutableServiceId create(@Nullable String prefix, @Nullable String alias) {
+            if (prefix == null || alias == null) return null;
```
AdamDM: is it better to return Optional rather than nullable?

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -239,89 +213,140 @@ private String normaliseOverrideAlias(String alias) {
...
+        }
+
+        @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")   // checked by ServiceId.equals
+        @Override public boolean equals(@Nullable Object obj) {
+            return ServiceId.equals(this, obj);
+        }
+        @Override public int hashCode() {
+            return ServiceId.hashCode(this);
+        }
+        private transient String str;
```
AdamDM: is there a reason for this to be being declared midway through the class?

src/main/java/org/atlasapi/remotesite/youview/YouViewChannelResolver.java
```diff
@@ -1,25 +1,59 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.Objects;
+
 public interface YouViewChannelResolver {
 
     @Deprecated
-    Collection<String> getChannelUris(int channelId);
+    Collection<String> getChannelUris(int serviceId);
```
AdamDM: you've now got a class called ServiceId but you're using an int variable with the same name here, that could get confusing, can you think of a better way of naming either?

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -239,89 +213,140 @@ private String normaliseOverrideAlias(String alias) {
...
         public Set<Channel> getAllChannels() {
-            return channel2alias.keySet();
+            return channelMap.inverse().keySet();
         }
 
-        public Multimap<Integer, Channel> getAllServiceIdsToChannels() {
-            return serviceId2channel;
+        public Multimap<ServiceId, Channel> getAllServiceIdsToChannels() {
+            return channelMap;
         }
```
alianos-: You are exposing a field here. I know it is Immutable, but technically if someone changes the internal map to mutable in the future he will have to think to check that (and sonar will mark this an a vulnerability btw).

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
```diff
@@ -239,89 +213,140 @@ private String normaliseOverrideAlias(String alias) {
...
             }
 
-            public boolean contains(Channel channel) {
-                return channel2alias.containsKey(channel);
+            public boolean contains(@Nullable Channel channel) {
+                return channel != null && channelMap.containsKey(channel);
+            }
+            public boolean contains(@Nullable ServiceId serviceId) {
+                return serviceId != null && serviceId.equals(serviceIdMap.get(serviceId.getId()));
             }
```
alianos-: There are different ServiceId objects with the same serviceId field? e.g. that contain different aliases?

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: Yes, it's purely a placeholder to lazy-cache the result of the toString operation.
It's a fairly common pattern in immutable types.
And generally you group the 'cache' field with the method that uses it.

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: TBH I'm tempted to use exceptions instead :-)

src/main/java/org/atlasapi/remotesite/youview/YouViewChannelResolver.java
TjlHope: The only thing I can think of is serviceIdId which isn't very good :-)
Have you got any ideas?

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: Bugger - I knew there was a reason for those being multimaps to start with - doh!

src/main/java/org/atlasapi/remotesite/youview/DefaultYouViewChannelResolver.java
TjlHope: Thing is - this whole object is an immutable one - here I disagree with sonar - the whole point of using immutable types is to prevent the need for defensive copying.
That said, I could just wrap it in an ImmutableSetMultimap.copyOf which should be a no-op unless someone changes the code...

PR #1315, opened 2017-09-25
Equivalence results improvements

AdamDM:
Yes it’s quite large but the second to last commit is pretty big and very samey
There is a reason the last commit is like it is - read the commit message it’ll explain why I’ve done that
Looking for feedback still though, as it might not be the best way of doing things


src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -225,11 +236,18 @@ public BarbAliasEquivalenceGenerator(
                     if (match) {
                         equivalents.addEquivalent((T) identified, Score.ONE);
                         desc.appendText("Resolved %s", identified.getCanonicalUri());
+
+                        // this if statement keeps lots of old tests happy
+                        if (identified.getId() != null) {
```
alianos-: If this approach is retained after the HH, maybe you can combine the ifs, i.e. if(match && identified.getId()!=null) . Unless you need the intermediate stuff to happen for the tests not to fail.

AdamDM: yeah I need the intermediate stuff to happen for the tests to not fail

PR #1316, opened 2017-09-29
Amazon telescope

Cottonglow:
Amazon Ingest to Telescope


src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -359,17 +409,52 @@ private void checkForDeletedContent() {
...
+                                notSeenContent.getId(),
+                                notSeenContent.getAliases(),
+                                EntityType.ITEM,
+                                UNPUBLISH_NO_PAYLOAD_STRING
+                        );
+                    } else {
+                        telescope.reportFailedEvent(
+                                "Atlas did not return an id after attempting to create or update this Item",
+                                EntityType.ITEM,
+                                UNPUBLISH_NO_PAYLOAD_STRING
```
alianos-: I think in the erroneous cases you should pass in the whole 'notSeenContent' as payload. In addition to the string.

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -359,17 +409,52 @@ private void checkForDeletedContent() {
...
                 if (notSeenContent instanceof Item) {
                     writer.createOrUpdate((Item) notSeenContent);
+                    //report to telescope
+                    if(notSeenContent.getId() != null){
+                        telescope.reportSuccessfulEvent(
+                                notSeenContent.getId(),
+                                notSeenContent.getAliases(),
+                                EntityType.ITEM,
+                                UNPUBLISH_NO_PAYLOAD_STRING
+                        );
```
alianos-: I cant tell what the successful event is to be honest. @AdamDM @dragospitica any ideas? It seems like we are loading resolved content and then we are rewritting it into the database. Presumably this is the content that we have in the db, but not in seen in the ingest, but why we are writting it to atlas it eludes me.

Maybe the payload should be "resolved content has no payload. This content was missing from the last ingest."

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -387,115 +472,190 @@ private void checkForDeletedContent() {
...
-            writer.createOrUpdate((Item) content);
+            writer.createOrUpdate(content.asModelType(Item.class).getModel());
+            //report to telescope
+            if(content.getModel().getId() != null){
+                telescope.reportSuccessfulEvent(
+                        content.getModel().getId(),
+                        content.getModel().getAliases(),
+                        EntityType.CONTENT,
+                        content.getPayload()
+                );
```
alianos-: You can potentially make use of the report successful event with autoTypeDiscovery which might be able to pinpoint a bit better what type of Content this is.

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxHttpFeedSupplier.java
```diff
@@ -18,6 +18,7 @@
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
+import org.atlasapi.reporting.telescope.OwlTelescopeReporter;
 import org.xml.sax.SAXException;
```
alianos-: unused import?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
+    public void finish(OwlTelescopeReporter telescope) {
+        processSeenContent(telescope);
+        processTopLevelSeries(telescope);
+        processStandAloneEpisodes(telescope);
         
         if (cached.values().size() > 0) {
             log.warn("{} extracted but unwritten", cached.values().size());
-            for (Entry<String, Collection<Content>> mapping : cached.asMap().entrySet()) {
+            for (Entry<String, Collection<ModelWithPayload<? extends Content>>> mapping : cached.asMap().entrySet()) {
                 log.warn(mapping.toString());
```
TjlHope: If we're now reporting the error to telescope, do we want the warn line?
or should it at least be more descriptive?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
-                seenContent.values()
-                        .stream()
-                        .filter(and(not(IS_BRAND::test), not(IS_SERIES::test))::apply)
-                        .collect(Collectors.toList());
-        for (Content notContainer : notContainers) {
-            checkAndWriteSeenContent(notContainer);
-        }
+        brands.stream().forEach(c -> checkAndWriteSeenContent(c, telescope));
+        series.stream().forEach(c -> checkAndWriteSeenContent(c, telescope));
+        notContainers.stream().forEach(c -> checkAndWriteSeenContent(c, telescope));
```
TjlHope: What's the point of splitting them into separate arrays, if you're just calling the same method on individual elements?
Could you not just call the methods directly in the for loop?

TjlHope: Actually the method seems to be the same regardless.
Why are they being split even?
Oh - if this is so that all brands are done first, then all series, then everything else, could you put a comment explaining

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
+            } else if (content.getModel() instanceof Container) {
+                write(mergeItemsWithPayload(ContentMerger.asContainer(identified), content.asModelType(Container.class)),
+                        telescope);
             }
         }
     }
 
-    private void processStandAloneEpisodes() {
+    /**
+     * Method can only merge instances of Items and instances of Containers, otherwise a nullpointer exception occurs.
```
TjlHope: If that's the case I'd probably have 2 separate methods, one for each.
Or at least check early, and throw an IllegalArgumentException rather than an NPE
Also, if the method name is mergeItems, I wouldn't expect it to actually work with containers...

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
-                    Iterables.filter(cached.get(entry.getKey()), IS_ITEM::test)
-            );
+
+            List<ModelWithPayload<Item>> cachedItems =
+                    cached.get(entry.getKey()).stream()
+                            .filter(c -> c.getModel() instanceof Item)
+                            .map(c -> c.asModelType(Item.class))
+                            .collect(Collectors.toList());
+
+            ModelWithPayload<Item> item = Iterables.getOnlyElement(cachedItems);
```
TjlHope: Rather than collecting to a list, and then using `getOnlyElement`, couldn't you use the `findAny` method on the stream - or if you want the explicit checking given by the `getOnlyElement` method, then use `Stream.reduce` with a method that throws?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
-                    Iterables.filter(cached.get(entry.getKey()), IS_SERIES::test)
-            );
+
+            List<ModelWithPayload<Series>> cachedSeries =
+                    cached.get(entry.getKey()).stream()
+                            .filter(c -> c.getModel() instanceof Series)
+                            .map(c -> c.asModelType(Series.class))
+                            .collect(Collectors.toList());
+
+            ModelWithPayload<Series> series = Iterables.getOnlyElement(cachedSeries);
```
TjlHope: ditto with the findAny

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -196,180 +201,266 @@ private void checkForDuplicatesSaveForProcessing(Content content) {
...
             }
         }
     }
      
-    private void checkForDeletedContent() {
-        Set<Content> allLoveFilmContent = ImmutableSet.copyOf(resolveAllLoveFilmContent());
-        Set<Content> notSeen = Sets.difference(allLoveFilmContent, seenContent.values());
+    private void checkForDeletedContent(OwlTelescopeReporter telescope) {
+        Set<Content> allAmazonContent = ImmutableSet.copyOf(resolveAllAmazonUnboxContent());
+        Set<Content> contentWithPayload = seenContent.values().stream()
```
TjlHope: slightly misleading name, as it's now content without the payload :-)

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -387,115 +478,190 @@ private void checkForDeletedContent() {
...
+                    EntityType.SERIES,
+                    series.getPayload()
+            );
+        } else {
+            telescope.reportFailedEvent(
+                    "Atlas did not return an id after attempting to create or updare this Series",
+                    EntityType.SERIES,
+                    series.getPayload()
+            );
+        }
```
TjlHope: I've seen this if statement in several different forms now.
Could you not abstract it out - maybe pass in the ModelWithPayload, and the EntityType, then modify the FailedEvent string to include the EntityType as a placeholder?

src/main/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessor.java
```diff
@@ -387,115 +478,190 @@ private void checkForDeletedContent() {
...
+                telescope.reportSuccessfulEvent(
+                        content.getModel().getId(),
+                        content.getModel().getAliases(),
+                        content.getModel(),
+                        content.getPayload()
+                );
+            } else {
+                telescope.reportFailedEvent(
+                        "Atlas did not return an id after attempting to create or update this Content",
+                        EntityType.CONTENT,
```
TjlHope: could you be more specific about the EntityType here?
I think andreas put a method somewhere to allow converting the object to a type (maybe OwlTelescopeReporter)

src/test/java/org/atlasapi/remotesite/amazonunbox/AmazonUnboxContentWritingItemProcessorTest.java
```diff
@@ -57,6 +56,8 @@
     private ContentLister lister;
     @Mock
     private AmazonUnboxBrandProcessor brandProcessor;
+    @Mock
+    private OwlTelescopeReporter telescope = mock(OwlTelescopeReporter.class);
```
TjlHope: you don't need both the `@Mock` annotation and the `mock(Type.class)` (that's the point of the annotation)

PR #1317, opened 2017-10-03
initial changes

Cottonglow:
Added a ContentAliasScorer class and an empty Test class.


src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private static final String NAME = "Content-Alias-Scorer";
+
+    private final Score mismatchScore;
+
+    public ContentAliasScorer(Score mismatchScore) {
+        this.mismatchScore = checkNotNull(mismatchScore);
+    }
+
+    @Override public ScoredCandidates<Item> score(Item subject, Set<? extends Item> candidates,
```
AdamDM: style: put @override on the line above for method declarations, and if your parameters make the line too long list all parameters line by line

e.g.
public ScoredCandidates<Item> score(
                                Item subject,
                                Set<? extends Item> candidates,
                                ResultDescription desc
) {
         ....

AdamDM: (but github isn't liking my spacing - Intellij normally makes it clear for you, if not ask someone)

AdamDM: Also I just noticed that you're missing something that hasn't been added yet - it's in my equiv results improvements that hasn't been deployed yet - you couldn't have known, but do remind me to add it later (can't add until I deploy as it relies on a lot more code)
Let me know if you end up deploying this before next Tuesday as it'll make my deploy not compile as I've changed the EquivalenceScorer interface

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,63 @@
...
+                                                  ResultDescription desc) {
+        DefaultScoredCandidates.Builder<Item> equivalents = DefaultScoredCandidates.fromSource(NAME);
+
+        for (Item candidate : candidates) {
```
AdamDM: You've probably got this from the other scorer you were going on - but for each loops are pretty much not used in Java any more since Java 8 - ask someone to explain how to turn this into a stream instead

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,63 @@
...
+            equivalents.addEquivalent(candidate, score(subject, candidate));
+        }
+
+        return equivalents.build();
+    }
+
+    @Override public String toString(){
```
AdamDM: style: stick this at the bottom of the class for neatness, and as above don't forget to move the @override to the line above

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,63 @@
...
+        return NAME;
+    }
+
+    private Score score(Item subject, Item candidate) {
+        Set<String> aliasStringsOfCandidate = candidate.getAliasUrls();
+
+        Set<String> aliasStringsOfSubject = subject.getAliasUrls();
+
+        if (!Sets.intersection(aliasStringsOfCandidate, aliasStringsOfSubject)
```
AdamDM: similar to the new line formatting of the parameters for the method header mentioned above, put a new line after the !Sets.Intersection( and then another before the last ) as it looks a bit neater and is easier to read

Cottonglow:
I have added Tests now as well.

src/test/java/org/atlasapi/equiv/scorers/ContentAliasScorerTest.java
```diff
@@ -0,0 +1,91 @@
...
+                new DefaultDescription());
+
+        assertThat(results.candidates().get(candidate), is(Score.ONE));
+    }
+
+    @Test
+    public void testScoresMismatchIfNoContentAliasMatch(){
+
+        Item subject = new Item("subj", "subj", Publisher.YOUVIEW);
+        Alias subjAlias = new Alias("subjAlias", "subjAlias");
```
jamiesprax: You should ideally make these values different in case you're reading the wrong one accidentally somewhere

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,69 @@
...
+
+        candidates.forEach(candidate ->
+                equivalents.addEquivalent(candidate, score(subject, candidate)));
+
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item candidate) {
+        Set<String> aliasStringsOfCandidate = candidate.getAliasUrls();
+
```
jamiesprax: un-needed whitespace

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,69 @@
...
+        if (!Sets.intersection(
+                aliasStringsOfCandidate,
+                aliasStringsOfSubject)
+                .isEmpty()) {
+            return Score.ONE;
+        }
+
+        Set<Alias> aliasesOfCandidate = candidate.getAliases();
+
```
jamiesprax: whitespace

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,69 @@
...
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item candidate) {
+        Set<String> aliasStringsOfCandidate = candidate.getAliasUrls();
+
+        Set<String> aliasStringsOfSubject = subject.getAliasUrls();
+
+        if (!Sets.intersection(
```
jamiesprax: This is <100 chars so could be single-lined

src/main/java/org/atlasapi/equiv/scorers/ContentAliasScorer.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public ContentAliasScorer(Score mismatchScore) {
+        this.mismatchScore = checkNotNull(mismatchScore);
+    }
+
+    @Override
+    public ScoredCandidates<Item> score(
+            Item subject,
+            Set<? extends Item> candidates,
+            ResultDescription desc) {
```
jamiesprax: `) {` would usually go on the next line down when doing this

PR #1320, opened 2017-10-08
add logic to give content with duration < 10m a 2m flexibility

AdamDM:


src/main/java/org/atlasapi/equiv/generators/BroadcastMatchingItemEquivalenceGenerator.java
```diff
@@ -218,11 +219,29 @@ private boolean flexibleAroundEndTime(DateTime transmissionTime, DateTime transm
...
+                broadcast.getTransmissionTime(),
+                broadcast.getTransmissionEndTime()
+        );
+
         DateTime start = broadcast.getTransmissionTime().minus(flexibility);
         DateTime end = broadcast.getTransmissionEndTime().plus(flexibility);
+
+        // 600000 millis is 10 minutes
+        // if the broadcast is less than 10 minutes long, reduce the flexibility
+        if (broadcastPeriod.compareTo(new Duration(600000)) < 0) {
```
jamiesprax: You can avoid the explanation comment by naming by naming a constant TEN_MINUTES or something

jamiesprax: or Duration.standardMinutes(10) like you've used above

src/main/java/org/atlasapi/equiv/generators/BroadcastMatchingItemEquivalenceGenerator.java
```diff
@@ -218,11 +219,29 @@ private boolean flexibleAroundEndTime(DateTime transmissionTime, DateTime transm
     }
 
     private Schedule scheduleAround(Broadcast broadcast, Set<Publisher> publishers) {
+        Duration shortBroadcastFlexibility = Duration.standardMinutes(2);
```
TjlHope: The original `flexibility` is passed into the constructor. Is it worth making this a param as well?
Or should it be a percentage of the original?
Or should we be using a percentage linear to the duration?

src/main/java/org/atlasapi/equiv/generators/BroadcastMatchingItemEquivalenceGenerator.java
AdamDM: good point, done

src/main/java/org/atlasapi/equiv/generators/BroadcastMatchingItemEquivalenceGenerator.java
AdamDM: I made it a final field, but if it's made a percentage of the original then very short content will never match, and could be other issues

PR #1330, opened 2017-12-12
Added reporting to status service

Cottonglow:


src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -189,14 +192,26 @@ public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWri
...
                     if (sery.getModel().getId() != null) {
-                        telescope.reportSuccessfulEvent(
+                        owlReporter.getTelescopeReporter().reportSuccessfulEvent(
                                 sery.getModel().getId(),
                                 sery.getModel().getAliases(),
                                 EntityType.SERIES,
                                 nitroBroadcast
                         );
+
+                        if (sery.getModel().getTitle() == null || sery.getModel().getTitle().isEmpty()){
```
dragospitica: Could use `Strings.isNullOrEmpty` instead

src/main/java/org/atlasapi/remotesite/bbc/nitro/ContentUpdatingNitroBroadcastHandler.java
```diff
@@ -169,14 +160,26 @@ public ContentUpdatingNitroBroadcastHandler(ContentResolver resolver, ContentWri
                     writer.createOrUpdate(brand.getModel());
                     //report to telescope
                     if (brand.getModel().getId() != null) {
-                        telescope.reportSuccessfulEvent(
+                        owlReporter.getTelescopeReporter().reportSuccessfulEvent(
```
dragospitica: Since you access it multiple times, it may be worth extracting it to a variable

src/main/java/org/atlasapi/AtlasMain.java
```diff
@@ -27,18 +13,27 @@
...
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMap.Builder;
+import org.atlasapi.util.jetty.InstrumentedQueuedThreadPool;
 import org.eclipse.jetty.io.ByteBufferPool;
-import org.eclipse.jetty.server.Connector;
-import org.eclipse.jetty.server.HttpConfiguration;
-import org.eclipse.jetty.server.HttpConnectionFactory;
-import org.eclipse.jetty.server.Server;
-import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.server.*;
```
dragospitica: It looks like you organized imports in the whole package. I assume that not all changed files are related to your work, therefore they should not be changed. Also, as per Java Style Guide #15, * imports are forbidden, too

PR #1336, opened 2018-02-12
Tjlh/bt channel removal

TjlHope:
Support removing channels from channel groups, as well as adding them.


src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
```diff
@@ -138,6 +149,39 @@ public AbstractBtChannelGroupSaver(
         return currentChannels;
     }
 
+    private boolean removeChannelGroupFromChannel(
```
AdamDM: to be clear, this is a method that tries to do something, and returns a boolean, that's false if it failed? It was a bit unclear, consider adding 'attempt' somewhere in the method name

src/main/java/org/atlasapi/remotesite/bt/channels/WatchableChannelGroupSaver.java
```diff
@@ -41,8 +40,7 @@ public WatchableChannelGroupSaver(
 
     @Override
     protected List<String> keysFor(Entry channel) {
-        // We're getting a problem with hm27 appearing, so hack this in for now...
-        if (channel.isStreamable() && !channel.getGuid().trim().equals("hm27")) {
+        if (channel.isStreamable() && !"hm27".equals(channel.getGuid().trim())) {
```
AdamDM: perhaps a comment about what hm27 is so people who are working on it later know
also you removed the comment saying this was a temporary hack, does this still need to be here?

TjlHope: Sorry - ignore hm27 - that's test code

src/main/java/org/atlasapi/remotesite/bt/channels/AbstractBtChannelGroupSaver.java
```diff
@@ -138,6 +149,39 @@ public AbstractBtChannelGroupSaver(
...
+        for (ChannelNumbering channelNumbering : channel.getChannelNumbers()) {
+            if (channelGroup.getId().equals(channelNumbering.getChannelGroup())) {
+                // don't add it to the new list
+                modified = true;    // and we'll need to do the write
+            } else {
+                newNumberings.add(channelNumbering);
+            }
+        }
+        if (!modified) {
+            return false;
```
jamiesprax: This section seems slightly confusing. Why does one case of the id's matching mean we don't update anything?

PR #1344, opened 2018-05-14
Synchronous equiv endpoint

tristanbeardmore-gray:


tristanbeardmore-gray:
Made the equiv update endpoint synchronous by default

PR #1347, opened 2018-06-18
Barb channels

jamiesprax:
A publisher was added to atlas-model and is live already so in order to resolve channels ingested this way, owl & deer api & processing need to be rebuilt and redeployed to pick up the atlas model changes even though this is just an owl processing change.


src/test/java/org/atlasapi/remotesite/barb/channels/BarbStringToChannelTransformerTest.java
```diff
@@ -0,0 +1,64 @@
...
+        String channelStrings = " 1111 , Test 1 |2222,Test2|3333, Test 3";
+
+        String[] splitChannels = channelStrings.split("\\|");
+        assertThat(splitChannels.length, is(3));
+
+        Channel[] channels = new Channel[3];
+
+        for (int i = 0; i < splitChannels.length; i++) {
+            channels[i] = barbTransformer.transform(splitChannels[i]);
+        }
```
alianos-: That test contains the split logic in itself so it doesn't test it. If someone changes how the endpoint works this test would still succeed, so if I am not mistaken it will succeed and fail along with the single channel test above. I would suggest you split the splitting logic in a method at the endpoint, and use it in the test to do the splitting.

jamiesprax: It was mainly to test the transformation logic but I see your point

PR #1351, opened 2018-06-28
Parent title scorer + amazon equiv

alianos-:


alianos-:
In the meantime I forgot what was the point of this, so I am closing the PR.

PR #1353, opened 2018-07-10
barb alias generator will now match content between ITV and STV. Also…

alianos-:
…, refactoring.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
+++ b/src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
@@ -1,11 +1,9 @@
 package org.atlasapi.equiv.generators;
 
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.metabroadcast.common.stream.MoreCollectors;
-import com.metabroadcast.common.stream.MoreStreams;
 import org.atlasapi.equiv.results.description.ResultDescription;
 import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
 import org.atlasapi.equiv.results.scores.Score;
@@ -19,8 +17,11 @@ import org.atlasapi.persistence.content.ResolvedContent;
 import org.atlasapi.persistence.lookup.entry.LookupEntry;
 import org.atlasapi.persistence.lookup.mongo.MongoLookupEntryStore;
 
-import java.util.Objects;
-import java.util.Set;
+import com.metabroadcast.common.stream.MoreCollectors;
+import com.metabroadcast.common.stream.MoreStreams;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableSet;
 
 public class BarbAliasEquivalenceGenerator<T extends Content> implements EquivalenceGenerator<T> {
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -19,8 +17,11 @@
 import org.atlasapi.persistence.lookup.entry.LookupEntry;
 import org.atlasapi.persistence.lookup.mongo.MongoLookupEntryStore;
 
-import java.util.Objects;
-import java.util.Set;
+import com.metabroadcast.common.stream.MoreCollectors;
+import com.metabroadcast.common.stream.MoreStreams;
+
+import com.google.common.base.Optional;
```
AdamDM: is there a specific need for this or can we use java util optional?

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -124,68 +129,74 @@ public BarbAliasEquivalenceGenerator(
...
+    }
+
+    /**
+     * Expand the list of aliases to include things that don't have the same namespace, but should
+     * still match. (such as originatingOwnerBcid and Bcid, or ITV and STV.
+     * We can then use the expanded list to fetch candidates that have either alias.
+     */
+    private Set<Alias> getExpandedAliases(Set<Alias> aliases) {
+        Set<Alias> expandedAliases = new HashSet<>();
+        for (Alias alias : aliases) {
```
AdamDM: consider streaming

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
```diff
@@ -124,68 +129,74 @@ public BarbAliasEquivalenceGenerator(
...
-                replaceNamespace(oldAlias.getNamespace(), oldNamespacePrefix, newNamespacePrefix),
-                oldAlias.getValue()
-        );
+                getNewNamespace(alias.getNamespace(), oldPrefix, newPrefix),
+                alias.getValue());
     }
-
-    private String replaceNamespace(String namespace, String oldNamespacePrefix, String newNamespacePrefix) {
-        return newNamespacePrefix.concat(namespace.substring(oldNamespacePrefix.length()));
+    //replaces the oldPrefix with the newPrefix in the given namespace.
```
AdamDM: comment and variables all talk about prefix, maybe `getNewNamespacePrefix(..` ?

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
alianos-: Y the lookupEntryStore requires guava. I guess it was made pre java8

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
alianos-: meh using a foreach here would be way less readable.

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGenerator.java
alianos-: It doesn't get the prefix, it replaces the oldPrefix with the newPrefix in the given namespace, and returns the namespace.

PR #1357, opened 2018-08-02
changed how rt film feed gets and transforms via http

AdamDM:
the transform logic was less tangled than I thought, this should do it.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
+++ b/src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
@@ -3,15 +3,26 @@ package org.atlasapi.remotesite.rt;
 import static org.atlasapi.persistence.logging.AdapterLogEntry.errorEntry;
 
 import java.io.InputStreamReader;
+import java.net.URI;
 
 import nu.xom.Builder;
 import nu.xom.Element;
 import nu.xom.NodeFactory;
 import nu.xom.Nodes;
 
+import org.apache.http.*;
 import org.apache.http.client.fluent.Request;
 
 import org.apache.http.client.fluent.Response;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicHttpRequest;
+import org.apache.http.message.BasicRequestLine;
+import org.apache.http.params.HttpParams;
 import org.atlasapi.persistence.content.ContentResolver;
 import org.atlasapi.persistence.content.ContentWriter;
 import org.atlasapi.persistence.logging.AdapterLog;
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
```diff
@@ -118,6 +115,24 @@ protected void runTask() {
         }
     }
 
+    private void getAndTransform(String feedUrl) throws Exception {
+        Response response = Request.Get("http://somehost/")
```
alianos-: somehost?

jamiesprax: is this supposed to be 'somehost' ?

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
```diff
@@ -118,6 +115,24 @@ protected void runTask() {
...
+                .connectTimeout(1000)
+                .socketTimeout(1000)
+                .execute();
+
+        InputStreamReader responseStream = new InputStreamReader(response.returnContent().asStream());
+
+        FilmProcessingNodeFactory filmProcessingNodeFactory = new FilmProcessingNodeFactory();
```
alianos-: Not sure what that factory does, but are you sure you need a new factory for http request?

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
```diff
@@ -3,20 +3,20 @@
...
 import java.io.InputStreamReader;
-import java.util.concurrent.TimeUnit;
 
 import nu.xom.Builder;
 import nu.xom.Element;
 import nu.xom.NodeFactory;
 import nu.xom.Nodes;
 
-import org.atlasapi.http.AbstractHttpResponseTransformer;
+import org.apache.http.client.fluent.Request;
```
alianos-: You tested that on stage I assume and these work with SSL? cause I can't comment on that.

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
```diff
@@ -100,14 +92,19 @@ protected void runTask() {
...
         } else {
-            requestUri += "/since?lastUpdated=" + UrlEncoding.encode(dateFormat.print(new DateTime(DateTimeZone.UTC).minusDays(3)));
+            requestUri += "/since?lastUpdated=" + UrlEncoding
+                    .encode(dateFormat.print(new DateTime(DateTimeZone.UTC).minusDays(3)));
         }
         
         try {
             reportStatus("Started...");
-            client.get(new SimpleHttpRequest<Void>(requestUri, TRANSFORMER));
+            getAndTransform(feedUrl);
```
jamiesprax: shouldn't this be requestUri?

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
```diff
@@ -118,6 +115,24 @@ protected void runTask() {
         }
     }
 
+    private void getAndTransform(String feedUrl) throws Exception {
+        Response response = Request.Get("http://somehost/")
+                .connectTimeout(1000)
+                .socketTimeout(1000)
```
jamiesprax: These don't match the previous settings that it was configured to of 5 min socket timeout, 1 min connect timeout

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
alianos-: That should be requestUri no?

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
alianos-: The original values were some orders of magnitude bigger, maybe there was a reason for that?

 .withSocketTimeout(5, TimeUnit.MINUTES)	
 .withConnectionTimeout(1, TimeUnit.MINUTES)

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
alianos-: Also I am uncertain how this feed works, but it's possible that there's a perfomance hit for not having a client. Maybe leave a comment somewhere (e.g. expected deploys) that this might be the case.

src/main/java/org/atlasapi/remotesite/rt/RtFilmFeedUpdater.java
jamiesprax: Also is it possible here to specify a user agent? Best to keep it as similar as possible

PR #1362, opened 2018-09-18
Wiki film equiv

jamiesprax:


src/main/java/org/atlasapi/equiv/scorers/TitleMatchingItemScorer.java
```diff
@@ -62,9 +62,9 @@ private boolean matches(String title) {
     }
 
     private final Score scoreOnMismatch;
-    
+
     public TitleMatchingItemScorer() {
-        this(Score.NULL_SCORE);
+        this(Score.nullScore());
```
jamiesprax: NULL_SCORE was deprecated for some reason, dw about this

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -196,6 +200,30 @@ private static UpdaterConfiguration makeDefaultConfiguration(Publisher publisher
...
     }
 
+    private static UpdaterConfiguration makeWikipediaConfiguration() {
+        return UpdaterConfiguration.builder()
+                .withSource(WIKIPEDIA)
+                .withItemEquivalenceUpdater(
+                        WIKIPEDIA_ITEM,
+                        ImmutableSet.of(
+                                BBC_NITRO, ITV_CPS, UKTV, C4_PMLSD, C5_DATA_SUBMISSION
+                        )
```
alianos-: We don't really want to do items though right? We probably need to find a way to indicate "don't do anything"

alianos-: oh there is no separate TOP LVL ITEMS, ignore me.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/WikipediaContainerUpdateProvider.java
```diff
@@ -0,0 +1,129 @@
...
+                )
+                .withScorers(
+                        ImmutableSet.of(
+                                new TitleMatchingContainerScorer(2),
+                                new ContainerHierarchyMatchingScorer(
+                                        dependencies.getContentResolver(),
+                                        Score.negativeOne(),
+                                        new SubscriptionCatchupBrandDetector(
+                                                dependencies.getContentResolver()
+                                        )
```
alianos-: This probably requires testing, but I doubt the hierarchy matching scorer is appropriate (as I doubt we can rely on wikipedia to have a good hierarchy).

I think we need a "same year" scorer (if it doesn't exist already).

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/WikipediaContainerUpdateProvider.java
```diff
@@ -0,0 +1,129 @@
...
+                                )
+                        )
+                )
+                .withCombiner(
+                        new RequiredScoreFilteringCombiner<>(
+                                new NullScoreAwareAveragingCombiner<>(),
+                                TitleMatchingContainerScorer.NAME,
+                                ScoreThreshold.greaterThanOrEqual(2)
```
alianos-: this can maybe be simplified to a sum.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/WikipediaContainerUpdateProvider.java
```diff
@@ -0,0 +1,129 @@
...
+                        )
+                )
+                .withFilter(
+                        ConjunctiveFilter.valueOf(ImmutableList.of(
+                                new MinimumScoreFilter<>(0.25),
```
alianos-: and that can probably increase to something reasonable (say if you get a max of 2 from title, and 0 or 2 if the years match), then this can be a maybe 3?

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/WikipediaContainerUpdateProvider.java
```diff
@@ -0,0 +1,129 @@
...
+                                        dependencies.getExcludedUris(),
+                                        dependencies.getExcludedIds()
+                                )
+                        ))
+                )
+                .withExtractors(
+                        ImmutableList.of(
+                                MultipleCandidateExtractor.create(),
+                                PercentThresholdAboveNextBestMatchEquivalenceExtractor
+                                        .atLeastNTimesGreater(1.5)
```
alianos-: And I would probably change that to plain greater than 3, or 3.8 or something (unsure how fuzzy title matching is).

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/WikipediaItemUpdateProvider.java
```diff
@@ -0,0 +1,128 @@
...
+    public EquivalenceUpdater<Item> getUpdater(
+            EquivalenceUpdaterProviderDependencies dependencies,
+            Set<Publisher> targetPublishers
+    ) {
+        return ContentEquivalenceUpdater.<Item>builder()
+                .withExcludedUris(dependencies.getExcludedUris())
+                .withExcludedIds(dependencies.getExcludedIds())
+                .withGenerators(
+                        ImmutableSet.of(
+                                new ContainerCandidatesItemEquivalenceGenerator(
```
alianos-: That is probably not needed as we don't want to discover episodes.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/WikipediaItemUpdateProvider.java
```diff
@@ -0,0 +1,128 @@
...
+                                        dependencies.getContentResolver(),
+                                        dependencies.getEquivSummaryStore()
+                                ),
+                                new FilmEquivalenceGenerator(
+                                        dependencies.getSearchResolver(),
+                                        targetPublishers,
+                                        DefaultApplication.createWithReads(
+                                                ImmutableList.copyOf(targetPublishers)
+                                        ),
+                                        true),
```
alianos-: no idea what that does.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/WikipediaItemUpdateProvider.java
```diff
@@ -0,0 +1,128 @@
...
+                                        Item.class,
+                                        targetPublishers,
+                                        2
+                                )
+                        )
+                )
+                .withScorers(
+                        ImmutableSet.of(
+                                new TitleMatchingItemScorer(),
+                                new SequenceItemScorer(Score.ONE)
```
alianos-: Again I think we don't need that as we are not doing episodes, and we need the same-year scorer.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/WikipediaItemUpdateProvider.java
```diff
@@ -0,0 +1,128 @@
...
+                                new DummyContainerFilter<>(),
+                                new UnpublishedContentFilter<>(),
+                                ExclusionListFilter.create(
+                                        dependencies.getExcludedUris(),
+                                        dependencies.getExcludedIds()
+                                )
+                        ))
+                )
+                .withExtractor(
+                        PercentThresholdEquivalenceExtractor.moreThanPercent(90)
```
alianos-: and again I have more faith in "everything over thresshold"

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
```diff
@@ -0,0 +1,54 @@
...
+            EquivToTelescopeResults equivToTelescopeResults
+    ) {
+        EquivToTelescopeComponent scorerComponent = EquivToTelescopeComponent.create();
+        scorerComponent.setComponentName("Item Year Scorer");
+        DefaultScoredCandidates.Builder<Item> equivalents = DefaultScoredCandidates.fromSource(YEAR);
+
+        for (Item candidate : candidates) {
+            Score score = score(subject, candidate);
+
+            equivalents.addEquivalent(candidate, score);
```
alianos-: I get the feeling this should be called something like scoredCandidates and not equivalents! I am sure you copied from somewhere, but just saying.

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
```diff
@@ -0,0 +1,54 @@
...
+            );
+        }
+
+        equivToTelescopeResults.addScorerResult(scorerComponent);
+
+        return equivalents.build();
+    }
+
+    private Score score(Item subject, Item candidate) {
+        return subject.getYear().equals(candidate.getYear()) ? matchScore : Score.nullScore();
```
alianos-: I think this needs to take into account that if the year is empty for both, we don't want to match them.

src/main/java/org/atlasapi/equiv/update/updaters/configuration/DefaultConfiguration.java
```diff
@@ -112,7 +113,8 @@
                             ITV_CPS,
                             BBC_NITRO,
                             C4_PMLSD,
-                            UKTV
+                            UKTV,
```
alianos-: while at it, add C5_DATASUBMISSION that was forgotten.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/WikipediaItemUpdateProvider.java
alianos-: I think that extractor finds the average, and only selects things that are over 90% of the average. Are we certain that's what we want?

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
jamiesprax: good point!

src/main/java/org/atlasapi/equiv/update/updaters/configuration/DefaultConfiguration.java
alianos-: Everything not in this list ends up with the standard job I believe. TxLogs equiv to C5 (double check that), and we don't want C5 to be equiving with random stuff.

jamiesprax: Only if you schedule it as a job in EquivTaskModule first. Then it matches the publisher to a config (or default). C5 won't generate equivs from itself to something else but can still be equivved to.

PR #1369, opened 2018-10-11
Update amazon equiv

jamiesprax:
allow amazon content to equiv to itself when the same content has two different asin's but are the same piece of content


src/main/java/org/atlasapi/equiv/generators/ExactTitleGenerator.java
```diff
@@ -0,0 +1,150 @@
...
+        // run if title contains non standard characters (eg cyrillic or chinese characters)
+        if (containsNonStandardChars(content.getTitle())) {
+            return searchForCandidates(content, desc);
+        }
+
+        // discard long titles
+        if (content.getTitle().length() >= 4) {
+            desc.appendText("subject title too long {}", content.getTitle().length());
+            return DefaultScoredCandidates.<T>fromSource(NAME).build();
+        }
```
alianos-: Wouldn't that whole thing be a bit easier to read if it said "run if titles contain non-standard characters, or are less than 4 characters" and otherwise return nothing?

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/AmazonItemUpdaterProvider.java
```diff
@@ -58,6 +61,12 @@ public static AmazonItemUpdaterProvider create() {
                         //candidates which are the item itself (because there is no further filtering
                         //to remove them, whereas the Publisher filter used elsewhere does that).
                         ImmutableSet.of(
+                                new ExactTitleGenerator<>(
+                                        dependencies.getSearchResolver(),
+                                        Item.class,
+                                        true,
+                                        AMAZON_UNBOX
+                                ),
```
alianos-: I get the feeling that the extractor used here is not using the Searcher, it's using the children of the container, so the problem for items might lie with the Scorer not scoring everything. 

The extractor for containers might be using the Searcher.

PR #1371, opened 2018-10-18
Created generator based on aliases (+ optional also namespaces, for...

EmiEnache:
…specific aliases) and an alias scorer for RT. Both added to the RT equivalence updater.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/AliasEquivalenceGenerator.java
+++ b/src/main/java/org/atlasapi/equiv/generators/AliasEquivalenceGenerator.java
@@ -1,0 +1,110 @@
+package org.atlasapi.equiv.generators;
+
+import java.util.Set;
+
+import org.atlasapi.equiv.results.description.ResultDescription;
+import org.atlasapi.equiv.results.scores.DefaultScoredCandidates;
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeComponent;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeResults;
+import org.atlasapi.media.entity.Alias;
+import org.atlasapi.media.entity.Content;
+import org.atlasapi.persistence.content.ContentResolver;
+import org.atlasapi.persistence.content.ResolvedContent;
+import org.atlasapi.persistence.lookup.entry.LookupEntry;
+import org.atlasapi.persistence.lookup.mongo.MongoLookupEntryStore;
+
+import com.metabroadcast.common.stream.MoreCollectors;
+import com.metabroadcast.common.stream.MoreStreams;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableSet;
+
+public class AliasEquivalenceGenerator<T extends Content> implements EquivalenceGenerator<T> {
+
+    private final MongoLookupEntryStore lookupEntryStore;
+    private final ContentResolver resolver;
+    private final Class<T> cls;
+    private String namespaceToMatch = null;
+
+    public AliasEquivalenceGenerator(
+            MongoLookupEntryStore lookupEntryStore,
+            ContentResolver resolver,
+            Class<T> cls
+    ) {
+        this.lookupEntryStore = lookupEntryStore;
+        this.resolver = resolver;
+        this.cls = cls;
+    }
+
+    public AliasEquivalenceGenerator(
+            MongoLookupEntryStore lookupEntryStore,
+            ContentResolver resolver,
+            Class<T> cls,
+            String namespaceToMatch
+    ) {
+        this.lookupEntryStore = lookupEntryStore;
+        this.resolver = resolver;
+        this.cls = cls;
+        this.namespaceToMatch = namespaceToMatch;
+    }
+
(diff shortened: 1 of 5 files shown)


EmiEnache:
This PR is related to equivalence: adds a configurable Alias Generator and an RT specific AliasScorer. The reason behind these additions is because some RT Uri’s can’t be trusted (long story but basically these items should be unpublished, however that might break other things, so instead I’m prioritising Aliases over URIs). Equivalence works decently enough for RT films, but not items (AKA episodes) because of RtFilmGenerator. This workaround should solve equivalence for RT items which use the old URIs.

src/main/java/org/atlasapi/equiv/generators/AliasEquivalenceGenerator.java
```diff
@@ -0,0 +1,117 @@
...
+        desc.startStage("Resolving aliases:");
+
+        EquivToTelescopeComponent generatorComponent = EquivToTelescopeComponent.create();
+        generatorComponent.setComponentName("Alias Equivalence Generator");
+
+        //for logging purposes
+        for (Alias alias : content.getAliases()) {
+            desc.appendText(alias.toString());
+        }
+        desc.finishStage();
```
alianos-: Shouldn't the description be a bit more specific, e.g. which alias matched?

src/main/java/org/atlasapi/equiv/results/filters/RtSpecializationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeResults;
+import org.atlasapi.media.entity.Content;
+import org.atlasapi.media.entity.Specialization;
+
+import com.google.common.base.Objects;
+
+/**
+ * This filter lets the PA specialization tv items equiv to RT specialization film items otherwise,
+ * it should work like SpecializationFilter does.
+ */
```
alianos-: Maybe also explain why this is needed

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/RtItemUpdaterProvider.java
```diff
@@ -73,7 +89,9 @@ public static RtItemUpdaterProvider create() {
                         ConjunctiveFilter.valueOf(ImmutableList.of(
                                 new MinimumScoreFilter<>(0.25),
                                 new MediaTypeFilter<>(),
-                                new SpecializationFilter<>(),
+                                //Some things are TV in PA's data but Films in RT, so need
+                                //specific specialization filter that does not filter those out
+                                new RtSpecializationFilter<>(),
```
alianos-: ha that addresses my previous comment.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/RtItemUpdaterProvider.java
```diff
@@ -85,7 +103,13 @@ public static RtItemUpdaterProvider create() {
                         ))
                 )
                 .withExtractor(
-                        PercentThresholdEquivalenceExtractor.moreThanPercent(90)
+                        // Get all items that scored at least perfect for
+                        // RtAlias. Then let it equiv to other stuff as well.
+                        ContinueUntilOneWorksExtractor.create(ImmutableList.of(
```
alianos-: if the RT aliases matches though, you wont equiv with anything else. So either update the comment or the extractor logic needs tweaking.

src/main/java/org/atlasapi/equiv/scorers/RtAliasScorer.java
```diff
@@ -0,0 +1,96 @@
...
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeComponent;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeResults;
+import org.atlasapi.media.entity.Alias;
+import org.atlasapi.media.entity.Item;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class RtAliasScorer implements EquivalenceScorer<Item> {
```
alianos-: Maybe a comment here why this was done with all your findings and actions we have taken (e.g. manual unpublishing in db). Better yet document in confluence a leave a link to confluence here. Imagine that one day someone will have to move this code from owl to deer and he will probably need to understand what happened and why.

src/main/java/org/atlasapi/equiv/generators/AliasEquivalenceGenerator.java
EmiEnache: The desc.finishStage() is always around this area of the code when used in generators. I don't really know what it does, so I didn't know that it should be somewhere else. What does it do?

EmiEnache: Also regarding mentioning on which alias it matched, it's difficult to get a specific alias without going through all of them. The alias(es) by which a potential candidate is picked up is/are listed already. It's true that it would be useful to see exact which one it matched by, but I wouldn't expect more than 1 alias to be picked up by AliasEquivalenceGenerator, and that will be the one it matched on.

PR #1373, opened 2018-11-20
Groups with deer ids

alianos-:


alianos-:
This is part of the work required for the new BT channel group UI offered via Shaman. 
a) it fixed a few bugs
b) it offers the ability to communicate with the new ID format (use with caution)
c) it allows callers to bust the cache for channelGroups, so you can see the changes.

Note there is work pending, as there are a lot of safeguards missing.

PR #1377, opened 2018-12-17
Debug endpoint which allows setting/unsetting explicit equivs on content

tristanbeardmore-gray:


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -210,6 +211,67 @@ private WriteResponse unpublishOldContent(
...
 
+    /**
+     * This was added as a way to update the explicit equivalence on an existing piece of content without having to
+     * hit the regular content PUT endpoint since we may not care to change the actual content itself.
+     *
+     * In particular this allows an easy way to remove all explicit equivalences on a piece of content since the other
+     * methods do not support updating the explicit equivalences with an empty set.
+     */
+    @Nullable
+    public WriteResponse updateExplicitEquivalence(HttpServletRequest req, HttpServletResponse resp) {
```
alianos-: It feels like much of this code must be a duplicate of something.

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -210,6 +211,67 @@ private WriteResponse unpublishOldContent(
         return null;
     }
 
+    /**
+     * This was added as a way to update the explicit equivalence on an existing piece of content without having to
+     * hit the regular content PUT endpoint since we may not care to change the actual content itself.
+     *
```
alianos-: Does this change the content itself as well, in the sense that the equiv table also exists in the content?

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
tristanbeardmore-gray: It is but it's a little hard to split out since the error checking returns WriteResponse objects

PR #1383, opened 2019-01-25
Keep stale nitro locations

dragospitica:


src/main/java/org/atlasapi/remotesite/bbc/nitro/BbcNitroModule.java
```diff
@@ -157,7 +160,7 @@ public void configure() {
         if (offScheduleIngestEnabled) {
             scheduler.schedule(
                     nitroOffScheduleIngestTask().withName("Nitro off-schedule content updater"),
-                    RepetitionRules.every(Duration.standardHours(3))
+                    RepetitionRules.NEVER
```
alianos-: why did you disable all ingesters?

PR #1386, opened 2019-03-07
Amazon equiv mongo title index

tristanbeardmore-gray:
Update Amazon equiv to use a separate title index instead of Owl search in order to get more reliable results.
On ingest we create an index of all Amazon content which shares the same title.


EmiEnache:
Nice unit tests btw

src/main/java/org/atlasapi/equiv/generators/amazon/AmazonTitleGenerator.java
```diff
@@ -0,0 +1,119 @@
...
+                .filter(content -> publishers.contains(content.getPublisher()))
+                .filter(content -> !onlyIncludeTopLevelContent || topLevelContent(content))
+                .filter(content -> content.getTitle().equals(subject.getTitle())) //shouldn't be needed but included to be safe
+                .collect(MoreCollectors.toImmutableList());
+
+        desc.appendText("%d candidates found for title: %s", filteredContent.size(), subject.getTitle());
+
+        for(T content : filteredContent) {
+            desc.appendText("Candidate: %s", content.getCanonicalUri());
+            scoredCandidates.addEquivalent(content, Score.nullScore());
```
dragospitica: If this is a valid candidate, why do we score it with null? Actually, we don't seem to care about this score at all, right?

src/main/java/org/atlasapi/equiv/generators/FilmEquivalenceGenerator.java
```diff
@@ -56,29 +61,59 @@
...
+     * and its internal FilmTitleMatcher. It also allows you to specify a tolerance for how much the years can differ by
+     * to still be considered a match.
+     * @param acceptNullYears if true will match if the subject on candidate year is null
+     * @param tolerableYearDifference the maximum amount the subject and candidate years can differ by to still match
+     * @param scoreOnImdbMatch the score given on an IMDB match
+     * @param scoreOnYearMismatch the score given if there is no IMDB match and the years do not match
+     * @param scoreOnTitleMatch the score given if there is no IMDB match and the year and title match
+     * @param scoreOnTitleMismatch the score given if there is no IMDB match and the year matches but the title does not
+     */
     public FilmEquivalenceGenerator(
```
alianos-: can we rename this to generatorAndScorer like the newest ones you have created?

src/main/java/org/atlasapi/equiv/generators/FilmEquivalenceGenerator.java
```diff
@@ -40,6 +39,12 @@
     private static final float BROADCAST_WEIGHTING = 0.0f;
     private static final float CATCHUP_WEIGHTING = 0.0f;
 
+    private static int DEFAULT_TOLERABLE_YEAR_DIFFERENCE = 1;
+    private static Score defaultScoreOnImdbMatch = Score.ONE;
+    private static Score defaultScoreOnYearMismatch = Score.negativeOne();
+    private static Score defaultScoreOnTitleMatch = Score.ONE;
+    private static Score defaultScoreOnTitleMismatch = Score.ZERO;
```
alianos-: Should all be final, and should be IN_CAPITALS

src/main/java/org/atlasapi/equiv/generators/amazon/AmazonTitleGenerator.java
alianos-: @dragospitica I think its because the generator returns scored candidates, but in this case we just want the generator to generate.

src/main/java/org/atlasapi/equiv/scorers/FilmYearScorer.java
```diff
@@ -0,0 +1,57 @@
...
+import org.atlasapi.equiv.results.scores.Score;
+import org.atlasapi.equiv.results.scores.ScoredCandidates;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeComponent;
+import org.atlasapi.equiv.update.metadata.EquivToTelescopeResults;
+import org.atlasapi.media.entity.Film;
+import org.atlasapi.media.entity.Item;
+
+import java.util.Set;
+
+public class FilmYearScorer extends ItemYearScorer {
```
alianos-: descriptions plx ^^

src/main/java/org/atlasapi/equiv/scorers/FilmYearScorer.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final String NAME = "Film-Year";
```
alianos-: is that used anywhere?

src/main/java/org/atlasapi/remotesite/amazon/AmazonContentWritingItemProcessor.java
```diff
@@ -203,6 +207,8 @@ public void finish() {
             );
         }
 
+        createTitleIndex();
```
alianos-: remove the existing logic that hits the owlIndexer endpoint directly. Might be in pre-processor, or a step up from this code.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/AmazonItemUpdaterProvider.java
```diff
@@ -61,12 +62,6 @@ public static AmazonItemUpdaterProvider create() {
                         //candidates which are the item itself (because there is no further filtering
                         //to remove them, whereas the Publisher filter used elsewhere does that).
```
alianos-: Add another comment here, stating WITH CAPS, that this not generate candidates that belong to different brands, because the scorer is not checking for that again, and will equiv them together.

src/main/java/org/atlasapi/equiv/generators/FilmTitleMatcher.java
```diff
@@ -1,44 +1,66 @@
...
+            return scoreOnMatch;
         }
         
         int difference = longer.length() - commonPrefix;
 
-        double scaler = - 0.1 / Math.max(shorter.length()-1,1);
+        double scaler = - (0.1 * scoreOnMatch.asDouble()) / Math.max(shorter.length()-1,1);
 
-        return Math.max((difference - 1) * scaler + 0.1, 0);
+        return Score.valueOf(Math.max((difference - 1) * scaler + (0.1 * scoreOnMatch.asDouble()), 0));
```
EmiEnache: TODO: confirm that logic isn't changed

src/main/java/org/atlasapi/equiv/generators/amazon/AmazonTitleGenerator.java
```diff
@@ -0,0 +1,119 @@
...
+    private final Class<? extends T> cls;
+    private final Set<Publisher> publishers;
+    private final boolean onlyIncludeTopLevelContent;
+
+
+    public AmazonTitleGenerator(
+            AmazonTitleIndexStore amazonTitleIndexStore,
+            ContentResolver resolver,
+            Class<? extends T> cls,
+            boolean onlyIncludeTopLevelContent,
```
EmiEnache: Maybe remove this, as we always want it to be true (because if false, we end up generating "Episode 1" as candidate for "Episode 1" regardless of them actually having different brands)

src/main/java/org/atlasapi/equiv/generators/amazon/AmazonTitleGenerator.java
```diff
@@ -0,0 +1,119 @@
...
+        desc.appendText("%d candidates found for title: %s", filteredContent.size(), subject.getTitle());
+
+        for(T content : filteredContent) {
+            desc.appendText("Candidate: %s", content.getCanonicalUri());
+            scoredCandidates.addEquivalent(content, Score.nullScore());
+        }
+        return scoredCandidates.build();
+    }
+
+    private boolean topLevelContent(Content content) {
```
EmiEnache: Check that such a method doesn't already exist somewhere else in the monolith - if so, then use that instead.

src/main/java/org/atlasapi/remotesite/amazon/AmazonContentWritingItemProcessor.java
EmiEnache: TODO: ^^^^^

src/main/resources/config/environment.properties
```diff
@@ -11,6 +11,7 @@ mongo.dbName=atlas
 mongo.maxConnections=100
 mongo.audit.dbname=atlas-audit
 mongo.audit.enabled=false
+mongo.amazon.dbname=amazon
```
EmiEnache: Check if these tend to be added to the config files - if so, then do it (just to maintain consistency)

src/main/java/org/atlasapi/equiv/scorers/FilmYearScorer.java
EmiEnache: Yes: in the scoredCandidates, and in toString()

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/amazon/AmazonToAmazonItemUpdaterProvider.java
```diff
@@ -0,0 +1,106 @@
...
+                .withExcludedUris(dependencies.getExcludedUris())
+                .withExcludedIds(dependencies.getExcludedIds())
+                .withGenerators(
+                        //whatever generators are used here, should prevent the creation of
+                        //candidates which are the item itself (because there is no further filtering
+                        //to remove them, whereas the Publisher filter used elsewhere does that).
+                        //Also, this SHOULD NOT GENERATE candidates that belong to different brands,
+                        //because the scorer DOES NOT CHECK for that and WILL equiv them together
+                        ImmutableSet.of(
+                                new ContainerCandidatesItemEquivalenceGenerator(
```
tristanbeardmore-gray: I think we need to add extra logic to this generator to optionally take in a list of target publishers and only generate candidates from containers which belong to one of the target publishers

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/amazon/AmazonToAmazonContainerUpdaterProvider.java
```diff
@@ -0,0 +1,89 @@
...
+                        )
+                )
+                .withScorers(
+                        ImmutableSet.of(
+                                new TitleMatchingContainerScorer(2)
+                        )
+                )
+                .withCombiner(
+                        new RequiredScoreFilteringCombiner<>(
+                                new AddingEquivalenceCombiner<>(),
```
tristanbeardmore-gray: I think we can just use an AddingEquivalenceCombiner here on its own

tristanbeardmore-gray: Might as well leave it actually just in case it's useful in future

src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesContainerEquivalenceGenerator.java
```diff
@@ -131,12 +148,35 @@ public ContainerCandidatesContainerEquivalenceGenerator(
         return equivSummaryStore.summariesForUris(ImmutableSet.of(parentUri));
     }
 
+    private ImmutableList<String> fetchContainerUris(EquivalenceSummary summary,
+            boolean strongCandidatesOnly, @Nullable Set<Publisher> publishers) {
+        //if we are looking at equivalents only, then we can filter out by publisher before resolving
+        if(strongCandidatesOnly){
```
tristanbeardmore-gray: `if(strongCandidatesOnly){
                return summary.getEquivalents().entries().stream()
                        .filter(input -> publishers == null || publishers.contains(input.getKey()))
                        .map(Map.Entry::getValue)
                        .map(ContentRef::getCanonicalUri)
                        .collect(MoreCollectors.toImmutableList());
        }
        else{
            return summary.getCandidates();
        }`

tristanbeardmore-gray: `
if (strongCandidatesOnly) {
    return summary.getEquivalents().entries().stream()
            .filter(input -> publishers == null || publishers.contains(input.getKey()))
            .map(Map.Entry::getValue)
            .map(ContentRef::getCanonicalUri)
            .collect(MoreCollectors.toImmutableList());
} else {
        return summary.getCandidates();
}
`

src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesItemEquivalenceGenerator.java
```diff
@@ -124,6 +143,29 @@ public ContainerCandidatesItemEquivalenceGenerator(
         return equivSummaryStore.summariesForUris(ImmutableSet.of(parentUri));
     }
 
+    private ImmutableList<String> fetchContainerUris(EquivalenceSummary summary,
+            boolean strongCandidatesOnly, @Nullable Set<Publisher> publishers) {
+        //if we are looking at equivalents only, then we can filter out by publisher before resolving
+        if(strongCandidatesOnly){
+            if(publishers != null){
+                return summary.getEquivalents().entries().stream()
```
tristanbeardmore-gray: Same comment as above

PR #1391, opened 2019-03-22
Exclude BARB broadcast equiv if BCID equiv has occurred

tristanbeardmore-gray:
Since CMS and Txlog broadcasts have been found to be
inaccurate in some cases we want to prevent a txlog
equiving on bcid to one episode of CMS content
but on broadcast to another episode of CMS content
(generally from the same brand) for example.

As a result a AllOverOrEqHighestNonEmptyThresholdExtractor
was created which will extract only scores above
10 if any exist (e.g. bcid equiv) otherwise
it will extract scores above 4 (which will cover
broadcast equiv).

An unfinished inefficient generator for barb was
also removed since it will never be used.


src/main/java/org/atlasapi/equiv/results/extractors/AllOverOrEqHighestNonEmptyThresholdExtractor.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.SortedSet;
+
+/**
+ * This extractor will select all candidates which score over the highest threshold that would extract at least one candidate
+ */
```
alianos-: This extractor will select all candidates which score over or equal to the highest threshold that manages to extract at least one candidate. For example "try to select everything over 6, if nothing scored over 6 then try everything over 3, and so on". Thresholds are attempted from highest to lowest.

PR #1398, opened 2019-04-23
Sky bcid equiv

jamiesprax:
Sky has bad data so if an originating owner broadcaster exists and isnt sky, don't use the parent bcid aliases to generate equiv candidates.

https://metabroadcast.atlassian.net/browse/ENG-267

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorer.java
+++ b/src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorer.java
@@ -65,6 +65,7 @@ public class BarbAliasEquivalenceGeneratorAndScorer<T extends Content> implement
     private static final String STV_OOBG_PREFIX = "gb:barb:originatingOwner:broadcastGroup:111";
     private static final String SKY_BG_PREFIX = "gb:barb:broadcastGroup:5";
     private static final String SKY_OOBG_PREFIX = "gb:barb:originatingOwner:broadcastGroup:5";
+    private static final String SKY_PARENT_BCID_SUFFIX = "5:parentVersionBcid";
     private static final String BARB_CONTENT_ID_NAMESPACE = "gb:barb:contentid";
     
     private static final Set<String> C4_NAMESPACES = ImmutableSet.of(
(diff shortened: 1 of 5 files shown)


src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorer.java
```diff
@@ -282,6 +281,14 @@ private boolean isItself(T subject, LookupEntry entry) {
             
             addPrefixedC4Aliases(expandedAliases, alias.getNamespace(), value);
         }
+
+        // sky has bad data so ignore the parents in favour of the originating broadcaster if exists and isnt sky
+        if (aliases.stream().anyMatch(alias -> alias.getNamespace().startsWith(SKY_BG_PREFIX)) &&
+                expandedAliases.stream()
```
tristanbeardmore-gray: I'd probably use aliases here instead of expandedAliases just so we know the actual content had that oobg instead of us accidentally expanding to one with it

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorer.java
```diff
@@ -282,6 +281,14 @@ private boolean isItself(T subject, LookupEntry entry) {
...
+                        .map(Alias::getNamespace)
+                        .anyMatch(ns -> ns.startsWith(OOBG_PREFIX) && !ns.startsWith(SKY_OOBG_PREFIX))) {
+            expandedAliases.removeIf(alias -> alias.getNamespace().endsWith(PARENT_VERSION_BCID_NAMESPACE_SUFFIX));
```
tristanbeardmore-gray: Don't we only want to remove sky parent version bcid namespaces here? We still want to potentially keep the parentVersionBcid expanded from the oobcid alias

src/test/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorerTest.java
```diff
@@ -692,5 +693,57 @@ public void excludesPublishersWithSameAlias() {
         assertTrue(!scoredCandidates.candidates().containsKey(differentItem));
         assertThat(scoredCandidates.candidates().get(item), is(SCORE_ON_MATCH));
     }
+
```
tristanbeardmore-gray: I'd add a test for a piece of content with these aliases:
bgid:5, bcid: A
oobgid: 1, bcid: B,
parentBgid: 5, bcid: C
Will equiv to a piece of content with these aliases:
bgid: 1, bcid: D,
parentBgid: 1, bcid: B

This covers the case arising from my comment above

src/main/java/org/atlasapi/equiv/generators/BarbAliasEquivalenceGeneratorAndScorer.java
```diff
@@ -66,6 +63,9 @@
     private static final String ITV_OOBG_PREFIX = "gb:barb:originatingOwner:broadcastGroup:2";
     private static final String STV_BG_PREFIX = "gb:barb:broadcastGroup:111";
     private static final String STV_OOBG_PREFIX = "gb:barb:originatingOwner:broadcastGroup:111";
+    private static final String SKY_BG_PREFIX = "gb:barb:broadcastGroup:5";
+    private static final String SKY_OOBG_PREFIX = "gb:barb:originatingOwner:broadcastGroup:5";
+    private static final String SKY_PARENT_BCID_SUFFIX = "5:parentVersionBcid";
```
tristanbeardmore-gray: Make this the exact namespace, it could strip random stuff still with this, e.g. gb:barb:broadcastGroup:115:parentVersionBcid

PR #1401, opened 2019-05-21
trim broadcasts in owl-api on amendments

EmiEnache:
Problem is that broadcast amendments done thru owl-api (eg. stuff like txlogs)
make the broadcast that was already there stick around on that piece of content.
This change makes it so that they get trimmed (ie. marked as unpublished).

https://metabroadcast.atlassian.net/browse/ENG-226

### DIFF
--- a/src/main/java/org/atlasapi/query/QueryExecutorModule.java
+++ b/src/main/java/org/atlasapi/query/QueryExecutorModule.java
@@ -158,7 +158,6 @@ public class QueryExecutorModule {
 
     private BroadcastTrimmer broadcastTrimmer() {
         return new ScheduleResolverBroadcastTrimmer(
-                null,
                 scheduleResolver,
                 contentResolver,
                 contentWriter
--- a/src/main/java/org/atlasapi/query/QueryExecutorModule.java
+++ b/src/main/java/org/atlasapi/query/QueryExecutorModule.java
@@ -1,16 +1,11 @@
 package org.atlasapi.query;
 
-import com.google.common.collect.ImmutableList;
-import com.google.common.util.concurrent.Service;
-import com.google.common.util.concurrent.ServiceManager;
-import com.metabroadcast.common.ids.NumberToShortStringCodec;
-import com.metabroadcast.common.ids.SubstitutionTableNumberCodec;
-import com.metabroadcast.common.persistence.mongo.DatabasedMongo;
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/epg/ScheduleResolverBroadcastTrimmer.java
```diff
@@ -58,7 +58,7 @@ public void trimBroadcasts(
             Channel channel,
             Map<String, String> acceptableIds
     ) {
-        trimBroadcasts(defaultPublisher.get(), scheduleInterval, channel, acceptableIds);
+        trimBroadcasts(null, scheduleInterval, channel, acceptableIds);
```
dragospitica: This will break some ingesters from the monolith to trim broadcasts. Leave the `defaultPublisher` get set in the constructor for source-specific trimmers i.e. BbcNItroModule:244

src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
```diff
@@ -190,10 +184,10 @@ private void updateEbsSchedule(Content content, Item item) {
 
             ImmutableMap<String, String> acceptableIds = ImmutableMap.of(
                     broadcast.getSourceId(),
-                    content.getCanonicalUri()
+                    item.getCanonicalUri()
             );
 
-            trimmer.trimBroadcasts(broadcastInterval, channel, acceptableIds);
+            trimmer.trimBroadcasts(item.getPublisher(), broadcastInterval, channel, acceptableIds);
```
dragospitica: Also, remove EBS source from the trimmer in QueryExecutorModule:161

src/main/java/org/atlasapi/query/QueryExecutorModule.java
```diff
@@ -155,10 +156,9 @@ public KafkaConsumer contentWriteWorker() {
         );
     }
 
-    @Bean
-    BroadcastTrimmer broadcastTrimmer() {
+    private BroadcastTrimmer broadcastTrimmer() {
         return new ScheduleResolverBroadcastTrimmer(
-                Publisher.BT_SPORT_EBS,
+                null,
```
dragospitica: Remove the parameter completely, the other constructor will initiate the publisher as null.

PR #1402, opened 2019-06-06
Avoid cursor timeout

EmiEnache:
This is part of a workaround to prevent MongoCursorNotFoundException which stopped equiv from properly rerunning on all content from a publisher. The change is to use a List of content uris instead of an Iterator of Content, and resolving the content afterwards.

https://metabroadcast.atlassian.net/browse/ENG-316

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -436,7 +436,7 @@
         <dependency>
             <groupId>org.atlasapi</groupId>
             <artifactId>atlas-persistence</artifactId>
-            <version>${atlas.version}</version>
+            <version>5.5-SNAPSHOT</version>
             <exclusions>
                 <exclusion>
                     <artifactId>jetty</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -436,7 +436,7 @@
         <dependency>
             <groupId>org.atlasapi</groupId>
             <artifactId>atlas-persistence</artifactId>
-            <version>5.5-SNAPSHOT</version>
+            <version>5.0-SNAPSHOT</version>
             <exclusions>
                 <exclusion>
                     <artifactId>jetty</artifactId>
--- a/src/main/java/org/atlasapi/equiv/EquivTaskModule.java
+++ b/src/main/java/org/atlasapi/equiv/EquivTaskModule.java
@@ -1,23 +1,13 @@
 package org.atlasapi.equiv;
 
-import com.google.api.client.util.Sets;
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.base.Predicates;
-import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.util.concurrent.MoreExecutors;
-import com.google.common.util.concurrent.Service.Listener;
-import com.google.common.util.concurrent.Service.State;
-import com.metabroadcast.common.persistence.mongo.DatabasedMongo;
-import com.metabroadcast.common.queue.kafka.KafkaConsumer;
-import com.metabroadcast.common.scheduling.RepetitionRule;
-import com.metabroadcast.common.scheduling.RepetitionRules;
-import com.metabroadcast.common.scheduling.ScheduledTask;
-import com.metabroadcast.common.scheduling.SimpleScheduler;
+import java.util.Set;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.PostConstruct;
+
 import org.atlasapi.equiv.results.persistence.RecentEquivalenceResultStore;
 import org.atlasapi.equiv.results.probe.EquivalenceProbeStore;
 import org.atlasapi.equiv.results.probe.EquivalenceResultProbeController;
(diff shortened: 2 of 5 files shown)


pom.xml
```diff
@@ -436,7 +436,7 @@
         <dependency>
             <groupId>org.atlasapi</groupId>
             <artifactId>atlas-persistence</artifactId>
-            <version>${atlas.version}</version>
+            <version>5.5-SNAPSHOT</version>
```
tristanbeardmore-gray: Don't forget to switch this back before merging

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -719,7 +723,7 @@ public MongoScheduleTaskProgressStore progressStore() {
 
     private ContentEquivalenceUpdateTask publisherUpdateTask(final Publisher... publishers) {
         return new ContentEquivalenceUpdateTask(
-                contentLister,
+                (SelectedContentLister) contentLister,
```
tristanbeardmore-gray: This cast feels a bit out of place but I don't really have a better suggestion - maybe where this is autowired we could add a separate bean that has a return type of SelectedContentLister that returns the same object as the normal ContentLister bean?

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +118,91 @@ protected void runTask() {
 
         onStart(progress);
 
-        Iterator<Content> contents = filter(contentLister.listContent(listingCriteria(progress)));
+        List<String> contentUris = contentLister.listContentUris(listingCriteria(progress));
```
tristanbeardmore-gray: Here would be where the Iterator<String> is then converted to List<String>

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +118,91 @@ protected void runTask() {
...
                 //we need to make sure that when progress is written, everything before that item
                 //has completed successfully. The strategy chosen was to batch tasks into
                 //blocks of SAVE_EVERY_BLOCK_SIZE, add them to the executor, wait for the whole
                 //batch to finish, write progress, repeat until done.
                 CountDownLatch latch = new CountDownLatch(SAVE_EVERY_BLOCK_SIZE);
                 int submitted = 0;
-                while (shouldContinue() && contents.hasNext() && submitted < SAVE_EVERY_BLOCK_SIZE) {
-                    current = contents.next();
+                while (shouldContinue() && !uris.isEmpty() && submitted < SAVE_EVERY_BLOCK_SIZE) {
+                    currentUri = uris.get(0);
```
tristanbeardmore-gray: If you're going to do this make uris a Queue<String> (LinkedList is an implementation of the Queue interface)

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +118,91 @@ protected void runTask() {
...
+                    uris.remove(0);
```
tristanbeardmore-gray: Lists don't inherently support efficient remove operations

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +118,91 @@ protected void runTask() {
...
+        String currentUri = null;
+        Content currentContent = null;
         int processed = 0;
         try {
-            while (shouldContinue() && proceed && contents.hasNext()) {
-                current = contents.next();
-                proceed = handle(current);
+            while (shouldContinue() && proceed && !uris.isEmpty()) {
+                currentUri = uris.get(0);
+                uris.remove(0);
```
tristanbeardmore-gray: Same comments as above

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
EmiEnache: Is the LinkedList .remove() more efficient?

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +119,93 @@ protected void runTask() {
...
+            runSynchronously(contentUris);
         } else {
-            runAsyncronously(contents);
+            runAsynchronously(contentUris);
         }
     }
 
-    private void runAsyncronously(Iterator<Content> contents) {
-        Content current = null;
+    private void runAsynchronously(List<String> uris) {
```
tristanbeardmore-gray: Change this type to Queue<String>

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
tristanbeardmore-gray: Yes because the LinkedList remove method (implementing the Queue remove method) is designed to just remove the first node, which is trivial in a LinkedList. The way you had this originally set up meant that you had an underlying ArrayList which would have to resize every time you called remove (and the list could be pretty large since it's all top level content ids).

tristanbeardmore-gray: And if your question is .remove() versus .remove(0) then possibly no (but no guarantees in general), but if you change the method signature to require a Queue interface then you'll only be able to call .remove() anyway.

src/main/java/org/atlasapi/equiv/update/tasks/ContentEquivalenceUpdateTask.java
```diff
@@ -109,68 +119,93 @@ protected void runTask() {
...
             log.error(getName(), e);
-            onFinish(false, null);
+            onFinish(false, null);  //pass null so as to not persist state
+            Throwables.propagate(e);    //in order to set the task as failed
         }
         onFinish(shouldContinue(), null);
     }
 
-    private void runSyncronously(Iterator<Content> contents) {
+    private void runSynchronously(List<String> uris) {
```
tristanbeardmore-gray: See above

EmiEnache:
Same MongoCursorNotFound exception spotted for the DeltaEquivalenceUpdate. More testing/investigation needed, and likely changes to fix that cursor as well.

PR #1404, opened 2019-06-17
Ingest original version pids and actual tx start and end times

tristanbeardmore-gray:
This is for work for BARB relating to equivalence and filling
in episode and original version PIDs in CCIDSTxLogs.
More information can be found in JIRA ENG-339.

* Refactored some existing logic with version pids and
version crids in NitroEpisodeExtractor into
BaseNitroItemExtractor.


src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/BaseNitroItemExtractor.java
```diff
@@ -130,6 +147,20 @@ public boolean apply(@Nullable AvailableVersions.Version input) {
         extractAdditionalItemFields(source, item, now);
     }
 
+    //This was moved from NitroEpisodeExtractor and refactored slightly
+    private void addVersionAliasesToLocations(Item item, Version version, Alias versionPidAlias) {
```
tristanbeardmore-gray: Refactored from NitroEpisodeExtractor

PR #1409, opened 2019-07-17
Api endpoints to update explicit equiv

tristanbeardmore-gray:
ENG-282 contains the context but is to allow
for editors to explicitly equiv brands together
in order to move away from the frankly terrible
automatic brand equiv logic that happens at the
moment between most major broadcasters sources.

* Endpoint to explicitly equiv two pieces of content

* Endpoint to remove explicit equiv between two pieces of content

* Endpoint to update explicit equiv on a single piece of content
This endpoint is not needed but added due to there being a potential
need for it based on previous experience. It was also an initial
implementation for the above before deciding against using it.
Since it does not update the content record explicit equivalence
lists of other content it is explicitly equiving to it may cause
undesirable behaviour when updating explicit equivalence of other
content where the link might break due to a new link being made
to an additional piece of content from the same source - this
is due to the somewhat quirky relationship of the content record
explicit equivalences being a sort-of-subset of the lookup table
explicit equivalences.

### DIFF
--- a/src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
+++ b/src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
@@ -1,13 +1,10 @@
 package org.atlasapi.query.content;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.util.List;
-import java.util.Set;
-
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.metabroadcast.common.base.Maybe;
 import org.atlasapi.input.ModelReader;
 import org.atlasapi.input.ModelTransformer;
 import org.atlasapi.input.ReadException;
@@ -34,16 +31,18 @@ import org.atlasapi.query.content.merge.BroadcastMerger;
 import org.atlasapi.query.content.merge.ContentMerger;
 import org.atlasapi.query.content.merge.VersionMerger;
 import org.atlasapi.remotesite.channel4.pmlsd.epg.BroadcastTrimmer;
-
-import com.metabroadcast.common.base.Maybe;
-
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
 import org.joda.time.Interval;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.List;
+import java.util.Set;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -198,9 +197,14 @@ public class ContentWriteExecutor {
         }
     }
 
+    /**
+     * @param publishers the set of publishers whose explicit equivalences will be modified in the lookup table,
+     *                   with any other explicit equivalences to content from other publishers being unaffected.
+     *                   If null will be later set to all publishers present in the explicitEquivRefs.
+     */
     public void updateExplicitEquivalence(
             Content content,
-            Set<Publisher> publishers,
+            @Nullable Set<Publisher> publishers,
             Set<LookupRef> explicitEquivRefs
     ) {
         content.setEquivalentTo(explicitEquivRefs);
--- a/src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
+++ b/src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
@@ -201,6 +201,12 @@ public class ContentWriteExecutor {
      * @param publishers the set of publishers whose explicit equivalences will be modified in the lookup table,
      *                   with any other explicit equivalences to content from other publishers being unaffected.
      *                   If null will be later set to all publishers present in the explicitEquivRefs.
+     * e.g. Suppose content A (Publisher X) has explicit equivs to B (Publisher Y), C (Publisher Z)
+     *                   and explicitEquivsRefs is just a single element set of D (Publisher Y)
+     *                   If publishers is just a single element set of Publisher Y then the explicit equivs in the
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,6 +229,282 @@ private WriteResponse unpublishOldContent(
...
+            List<String> uris = parseList(req.getParameter(URIS));
+            List<String> ids = parseList(req.getParameter(IDS));
+
+            if (uris.size() + ids.size() != 2) {
+                throw new IllegalArgumentException("Must specify exactly two " + URIS + " or " + IDS);
+            }
+
+            List<Content> contents = resolveContent(uris, ids);
+            if (contents.size() != 2 || contents.get(0).equals(contents.get(1))) {
+                throw new IllegalArgumentException("Must specify exactly two pieces of content");
```
EmiEnache: Change message to "Must specify exactly two distinct pieces of content"?

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -438,6 +739,47 @@ private void removeItemFromEquivSet(Described described, LookupEntry lookupEntry
...
+                            SubstitutionTableNumberCodec.lowerCaseOnly().decode(id).longValue()
+                    )
+            );
+            if (!lookupEntries.iterator().hasNext()) {
+                throw new IllegalArgumentException("No lookup entry found for id " + id);
+            }
+        } else {
+            throw new IllegalArgumentException("No id or uri specified");
+        }
+        if (!lookupEntries.iterator().hasNext()) {
```
EmiEnache: Is this necessary? Won't all cases get caught by either one or the other checks above?

src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
```diff
@@ -198,9 +197,14 @@ private void updateSchedule(Content content, Item item) {
         }
     }
 
+    /**
+     * @param publishers the set of publishers whose explicit equivalences will be modified in the lookup table,
+     *                   with any other explicit equivalences to content from other publishers being unaffected.
+     *                   If null will be later set to all publishers present in the explicitEquivRefs.
+     */
```
alianos-: maybe add an example

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
         return null;
     }
 
+    /**
+     * Adds a bidirectional explicit equivalence link between two pieces of content.
+     * The content record list of explicit equivs is updated much like it is in the regular content update endpoints.
+     * Other existing explicit equivalences are left intact.
+     * The application must have write permissions to both pieces of content, unless the application has the role
+     * "explicit-equivalence-from-read-sources" in which case it must have either write permissions or read permissions
```
alianos-: please document that in confluence roles page

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
...
+        return Optional.of(encodeId(from.getId()));
+    }
+
+    private List<Content> getContentForExplicitEquivFromRequest(HttpServletRequest req) throws InvalidApiKeyException {
+        Application application = getAndValidateApplication(req);
+        List<String> uris = parseList(req.getParameter(URIS));
+        List<String> ids = parseList(req.getParameter(IDS));
+
+        if (uris.size() + ids.size() != 2) {
+            throw new IllegalArgumentException("Must specify exactly two " + URIS + " or " + IDS);
```
alianos-: maybe make it clearer that the sum must be two, this reads like there should be either 2 uris or 2 IDs

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
...
+        }
+
+        List<Content> contents = resolveContent(uris, ids);
+        if (contents.size() != 2 || contents.get(0).equals(contents.get(1))) {
+            throw new IllegalArgumentException("Must specify exactly two distinct pieces of content");
```
alianos-: you can probably throw exactly the same error above

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
...
+        if (!from.getEquivalentTo().contains(LookupRef.from(to))) {
+            return Optional.empty();
+        }
+        Set<LookupRef> newExplicits = new HashSet<>(from.getEquivalentTo());
+        newExplicits.remove(LookupRef.from(to));
+        writeExecutor.updateExplicitEquivalence(from, publishers(from.getEquivalentTo()), newExplicits);
+        return Optional.of(encodeId(from.getId()));
+    }
+
+    private List<Content> getContentForExplicitEquivFromRequest(HttpServletRequest req) throws InvalidApiKeyException {
```
alianos-: add a comment here that this returns exactly a list of size 2, or throws.

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
...
+                throw new IllegalArgumentException("Content is already explicitly equived");
+            }
+            resp.setStatus(HttpServletResponse.SC_OK);
+            return new MultiWriteResponse(updatedIds);
+        } catch (Exception e) {
+            return multiWriteError(req, resp, AtlasErrorSummary.forException(e));
+        }
+    }
+
+    private Optional<String> addExplicitEquivalence(Content from, Content to) {
```
alianos-: javadoc what the response of that method is

src/main/java/org/atlasapi/query/v2/ContentWriteController.java
```diff
@@ -211,17 +229,272 @@ private WriteResponse unpublishOldContent(
...
+        if (from.getEquivalentTo().contains(LookupRef.from(to))) {
+            return Optional.empty();
+        }
+        Set<LookupRef> newExplicits = new HashSet<>(from.getEquivalentTo());
+        newExplicits.add(LookupRef.from(to));
+        writeExecutor.updateExplicitEquivalence(from, publishers(newExplicits), newExplicits);
+        return Optional.of(encodeId(from.getId()));
```
alianos-: and since that response is only for the response, maybe make it something like id1->id2

PR #1415, opened 2019-09-02
Fix txlog equiv

EmiEnache:
BBC-Txlog content is sometimes mis-equiv'ing txlog to too many things because a) barb broadcast generator & scorer is scoring multiple times when there are multiple broadcasts and b) the extractor is too lenient (everything above 4).

We fixed a) to only score max once and for b) we split the equiv into bbc -> txlog equiv and txlog -> bbc equiv. The latter was made to prioritise bcid over broadcasts.
https://metabroadcast.atlassian.net/browse/ENG-447


src/main/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer.java
```diff
@@ -38,6 +38,8 @@
 /**
  * This is mostly a copy of {@link BroadcastMatchingItemEquivalenceGeneratorAndScorer} class but with some logic
  * stripped out. For certain channels it also generates candidates from additional channels.
+ * Candidates are expected to have a maximum score of 3, so we do not add up the score from multiple
```
tristanbeardmore-gray: Phrase this comment to say just that the highest score is taken for each candidate as opposed to summing them like in the regular broadcast generator.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/BbcTxlogsItemUpdaterProvider.java
```diff
@@ -105,10 +110,22 @@ public static BbcTxlogsItemUpdaterProvider create() {
...
                         ))
-                )
-                .withExtractor(
-                        AllOverOrEqThresholdExtractor.create(4)
-                )
-                .build();
+                );
+        //Tighten txlog to bbc equiv because it was mis-equiv'ing to some low scored candidates
+        //generated by BarbBroadcast. Bbc to txlog should stay the same. ENG-447
+        if(isSubjectTxlog) {
```
tristanbeardmore-gray: You don't need to store the builder variable because you could do
`
.withExtractor(
  isSubjectTxlog
   ? new AllOverOrEqHighestNonEmptyThresholdExtractor<>(ImmutableSet.of(10D, 4D))
   : AllOverOrEqThresholdExtractor.create(4)
)
`

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/BbcTxlogsItemUpdaterProvider.java
```diff
@@ -105,10 +110,22 @@ public static BbcTxlogsItemUpdaterProvider create() {
                                 new DummyContainerFilter<>(),
                                 new UnpublishedContentFilter<>()
                         ))
-                )
-                .withExtractor(
-                        AllOverOrEqThresholdExtractor.create(4)
-                )
-                .build();
+                );
+        //Tighten txlog to bbc equiv because it was mis-equiv'ing to some low scored candidates
```
tristanbeardmore-gray: Copy over the comment from the barb txlogs equiv config, and for the explanation for bbc to txlog say: 

> BBC to Txlogs needs to equiv to all candidates since some bbc txlogs are regional variants without bcids and still need to be equived to even if a txlog exists with a bcid

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/BbcTxlogsItemUpdaterProvider.java
EmiEnache: Ah yes, I thought there might be a nicer way of doing it than mine, but I couldn't rmb the syntax. Cheers!

PR #1417, opened 2019-09-09
Improvements for BBC Txlog equiv

tristanbeardmore-gray:
Added broadcast matching threshold logic for barb broadcast equiv

Also bumped BbcTxlogsItemUpdaterProvider score for actual transmission time equiv and extractor logic to factor in new score

Efficiency improvement in Barb Alias Generator to only resolve LookupEntries from target publishers

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer.java
+++ b/src/main/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer.java
@@ -48,12 +48,13 @@ import static org.atlasapi.equiv.generators.barb.utils.BarbGeneratorUtils.getQua
  */
 public class BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer implements EquivalenceGenerator<Item> {
 
+    private static final Duration SHORT_CONTENT_REDUCED_TIME_FLEXIBILITY = Duration.standardMinutes(10);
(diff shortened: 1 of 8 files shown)


src/main/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer.java
```diff
@@ -86,111 +95,170 @@ public BarbBroadcastMatchingItemEquivalenceGeneratorAndScorer(
...
+                                        broadcast,
+                                        getFlexibility(broadcast),
+                                        this::isValidBroadcast
+                                )
+                        );
+                    }
+                }
+            }
+
+            double ratioOfMatchingBroadcasts = ((double) matchingBroadcasts.size()) / numberOfValidBroadcasts;
```
alianos-: check denominator is now zero.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/BbcTxlogsItemUpdaterProvider.java
```diff
@@ -112,13 +109,14 @@ public static BbcTxlogsItemUpdaterProvider create(boolean isSubjectTxlog) {
                         ))
                 )
 
-                // See TxlogsItemUpdaterProvider for reason behind 10-4 extractor on txlog->bbc equiv
+                // See TxlogsItemUpdaterProvider for reason behind 6-4 extractor on txlog->bbc equiv
```
EmiEnache: Where is this "6-4 extractor"?

PR #1419, opened 2019-09-26
Merge empty title

dragospitica:


src/main/java/org/atlasapi/query/content/ContentWriteExecutor.java
```diff
@@ -160,7 +181,11 @@ private void writeContentInternal(
             writer.createOrUpdate(item);
             updateSchedule(content, item);
         } else {
-            writer.createOrUpdate((Container) updatedContent);
+            if (!Iterables.isEmpty(fieldsToRemove)) {
```
EmiEnache: Would swap this if-else around (to be   "if (Iterables.isEmpty(fieldsToRemove))", but it's a slight and personal preference.

PR #1440, opened 2020-01-17
Equiv between BBC regional txlogs (ENG-633)

tristanbeardmore-gray:
Equiv must require exactly same start time, end time, and title.

This is additional work for BBC for network linking, following on from all work done stemming from ENG-339,

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/barb/BarbBbcRegionalTxlogItemEquivalenceGeneratorAndScorer.java
+++ b/src/main/java/org/atlasapi/equiv/generators/barb/BarbBbcRegionalTxlogItemEquivalenceGeneratorAndScorer.java
@@ -1,0 +1,224 @@
+package org.atlasapi.equiv.generators.barb;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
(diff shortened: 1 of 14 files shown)


src/main/java/org/atlasapi/equiv/generators/barb/BarbBbcRegionalTxlogItemEquivalenceGeneratorAndScorer.java
```diff
@@ -0,0 +1,225 @@
...
+    @Nullable
+    private Set<String> expandRegionalChannelUris(String channelUri) {
+        Set<String> expandedUris = null;
+        if (BBC1_TXLOG_CHANNEL_URIS.contains(channelUri)) {
+            expandedUris = BBC1_TXLOG_CHANNEL_URIS;
+        }
+        if (BBC2_TXLOG_CHANNEL_URIS.contains(channelUri)) {
+            expandedUris = BBC2_TXLOG_CHANNEL_URIS;
+        }
+        return expandedUris == null ? null : Sets.difference(expandedUris, ImmutableSet.of(channelUri));
```
alianos-: you could do the difference and return in inside the if, and save 2 checks.

PR #1447, opened 2020-01-24
Eng 648

EmiEnache:
Improve IMDb brand equiv. Fix Amazon year scorer equiv & add logs.

### DIFF
--- a/src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
+++ b/src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
@@ -80,14 +80,19 @@ public class ContainerChildEquivalenceGenerator implements EquivalenceGenerator<
         );
         Multiset<String> parents = HashMultiset.create();
         for (EquivalenceSummary summary : Optional.presentInstances(childSummaries.values())) {
-            Iterables.addAll(parents, Iterables.filter(summary.getEquivalents()
-                    .values()
-                    .stream()
-                    .filter(input -> publishers == null
-                            || publishers.contains(input.getPublisher()))
-                    .map(TO_PARENT::apply)
-                    .collect(Collectors.toList()
-                    ), Predicates.notNull()));
+            Iterables.addAll(
+                    parents,
+                    Iterables.filter(
+                            summary.getEquivalents()
+                                    .values()
+                                    .stream()
+                                    .filter(input -> publishers == null
+                                            || publishers.contains(input.getPublisher()))
+                                    .map(TO_PARENT::apply)
+                                    .collect(Collectors.toList()),
+                            Predicates.notNull()
+                    )
+            );
         }
         return scoreContainers(
                 parents,
--- a/src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
+++ b/src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
@@ -1,5 +1,6 @@
 package org.atlasapi.equiv.generators;
 
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -27,7 +28,6 @@ import com.metabroadcast.common.collect.OptionalMap;
 
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
-import com.google.common.base.Predicates;
 import com.google.common.collect.HashMultiset;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
@@ -82,16 +82,14 @@ public class ContainerChildEquivalenceGenerator implements EquivalenceGenerator<
         for (EquivalenceSummary summary : Optional.presentInstances(childSummaries.values())) {
             Iterables.addAll(
                     parents,
-                    Iterables.filter(
-                            summary.getEquivalents()
-                                    .values()
-                                    .stream()
-                                    .filter(input -> publishers == null
-                                            || publishers.contains(input.getPublisher()))
-                                    .map(TO_PARENT::apply)
-                                    .collect(Collectors.toList()),
-                            Predicates.notNull()
-                    )
+                    summary.getEquivalents()
+                            .values()
+                            .stream()
+                            .filter(input -> publishers == null
+                                    || publishers.contains(input.getPublisher()))
+                            .map(TO_PARENT::apply)
+                            .filter(Objects::nonNull)
+                            .collect(Collectors.toList())
             );
         }
         return scoreContainers(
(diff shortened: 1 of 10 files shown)


src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
```diff
@@ -61,9 +80,20 @@ public ContainerChildEquivalenceGenerator(
...
-            ), Predicates.notNull()));
+            if (publishers == null) {
+                Iterables.addAll(parents, Iterables.filter(Iterables.transform(
+                        summary.getEquivalents().values(), TO_PARENT
+                ), Predicates.notNull()));
+            } else {
+                Iterables.addAll(parents, Iterables.filter(summary.getEquivalents()
+                        .values()
+                        .stream()
+                        .filter(input -> publishers == null
```
tristanbeardmore-gray: This is within the `else` statement where `publishers` is not null. Either remove the if statement or remove this null check

src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
```diff
@@ -61,9 +80,20 @@ public ContainerChildEquivalenceGenerator(
...
+                                || publishers.contains(input.getPublisher()))
+                        .map(TO_PARENT::apply)
```
tristanbeardmore-gray: If you're trying to match the existing logic I believe you need to filter out nulls after mapping to parent

src/main/java/org/atlasapi/equiv/generators/ContentTitleScorer.java
```diff
@@ -17,18 +17,30 @@
     private final Function<String, String> titleTransform;
     private final String name;
     private final Score exactMatchScore;
-    private final Score partialMatchBound;
+    private final Score partialMatchUpperBound;
+    private final boolean disableScalingOnPartialMatch; //basically makes partial scoring more lenient
```
tristanbeardmore-gray: I'd change this to be `scaleOnPartialMatch` and change logic as appropriate

src/main/java/org/atlasapi/equiv/results/scores/ScoredCandidates.java
```diff
@@ -11,6 +11,7 @@
     String source();
 
     Map<T, Score> candidates();
+    //TODO get this map and use it to manipulate the scores
```
tristanbeardmore-gray: You forgot to remove this

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
```diff
@@ -67,9 +80,12 @@ public ItemYearScorer(Score matchScore, Score mismatchScore, Score nullYearScore
     }
 
     protected Score score(Item subject, Item candidate) {
-        if (subject.getYear() == null || candidate.getYear() == null) {
+        if (subject.getYear() == null && candidate.getYear() == null) {
```
tristanbeardmore-gray: I disagree with changing this behaviour, and it also is no longer consistent with the container year scorer

src/main/java/org/atlasapi/equiv/scorers/TitleMatchingContainerScorer.java
```diff
@@ -19,15 +19,20 @@
...
     public TitleMatchingContainerScorer(double exactTitleMatchScore) {
-        this(Score.valueOf(exactTitleMatchScore), Score.ONE);
+        this(Score.valueOf(exactTitleMatchScore), Score.ONE, false);
     }
 
-    public TitleMatchingContainerScorer(Score exactTitleMatchScore, Score partialTitleMatchBound) {
+    public TitleMatchingContainerScorer(
+            Score exactTitleMatchScore,
+            Score partialTitleMatchBound,
+            boolean disableScalingOnPartialMatch
```
tristanbeardmore-gray: Same comment on this boolean

src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
EmiEnache: Are you sure there's a problem there? I didn't do anything except except let the IDE convert the .transform() to use streams instead. Also, there is a filter for nulls that applies 2 lines down, after mapping to parent and collecting it.

(I basically tried to change the minimum necessary that allowed me to add my change)

src/main/java/org/atlasapi/equiv/generators/ContentTitleScorer.java
EmiEnache: I went back forth with myself on this one!

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
```diff
@@ -67,9 +80,12 @@ public ItemYearScorer(Score matchScore, Score mismatchScore, Score nullYearScore
...
             return nullYearScore;
         }
+        if (subject.getYear() == null || candidate.getYear() == null) {
```
tristanbeardmore-gray: Because this isn't really a mismatch just a lack of information

src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGenerator.java
tristanbeardmore-gray: Oh you may be right, think I didn't see the iterable filter because was expecting it to be in a stream.
My honest opinion is that we should completely rework that into stream logic for cleanliness but this might not be the best time

EmiEnache: Similar reasoning here; I wasn't comfortable reworking into it pure stream logic as it's used in too many places

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
EmiEnache: For the purpose of equiv I'm not sure the distinction would be meaningful enough. 
We'd have to replace it with a different, fourth score which is very similar to the nullYearScore (almost always the same, except for Amazon<->Amazon where they have to be different).

EmiEnache: Talked more and agreed it has to be done either as a variable, or as a distinct, bespoke scorer. I did the former as it was faster

src/main/java/org/atlasapi/equiv/scorers/ItemYearScorer.java
EmiEnache: See above

PR #1455, opened 2020-03-06
Eng 737

EmiEnache:


src/main/java/org/atlasapi/input/DescribedModelTransformer.java
```diff
@@ -83,13 +88,36 @@ protected T setFields(T result, F inputContent) {
...
+    ) {
+        return simpleLocalizedDescriptions.stream()
+                .map(input -> {
+                    LocalizedDescription localizedDescription = new LocalizedDescription();
+                    localizedDescription.setDescription(input.getDescription());
+                    localizedDescription.setShortDescription(input.getDescription());
+                    localizedDescription.setMediumDescription(input.getMediumDescription());
+                    localizedDescription.setLongDescription(input.getLongDescription());
+                    localizedDescription.setLocale(
+                            Strings.nullToEmpty(input.getLanguage()),
```
tristanbeardmore-gray: Doesn't this method accept nullable fields anyway?

src/main/java/org/atlasapi/output/simple/DescribedModelSimplifier.java
```diff
@@ -238,7 +238,11 @@ private String applyWatermark(F described, String description) {
     private static final Function<org.atlasapi.media.entity.LocalizedDescription, LocalizedDescription> TO_SIMPLE_LOCALISED_DESCRIPTION = complex -> {
         LocalizedDescription simple = new LocalizedDescription();
 
-        simple.setLanguage(complex.getLanguageTag());
+        //ignore NPE because it's triggered by complex possibly being null, which it likely won't
```
tristanbeardmore-gray: Might want to specify that it's the IDE warning about the NPE you might want to ignore

PR #1457, opened 2020-03-18
(ENG-784) Improvements to BBC Txlog Equivalence

tristanbeardmore-gray:
* Title comparison logic should consider the CMS series
title in addition to brand and episode titles.

* Title comparison logic should deal with the
concatenation of CMS title fields in any
combination or order.

* Actual transmission time equivalence for BBC
incorporates title comparison logic since the
actual transmission time can be incorrect and is
thus an insufficient criteria on its own for
equivalence.

These changes are to improve BBC Txlog equivalence
but I have made this change more generally since
the title comparison logic is intentionally fuzzy and will
potentially improve equivalence for other broadcaster's
content as well.


src/main/java/org/atlasapi/equiv/scorers/barb/BarbTitleMatchingItemScorer.java
```diff
@@ -196,70 +215,164 @@ public static Builder builder() {
...
                 new ContentTitleMatchingFields(subject),
                 new ContentTitleMatchingFields(suggestion),
                 desc
         );
         if (equivScore != scoreOnPerfectMatch) {
-            // txlog titles are often just the title of the brand so score the txlog title
-            // against a brand title as well if applicable
-            Optional<Score> parentScore = getParentScore(subject, suggestion, desc);
+            // txlog titles can often contain brand or series information so score the txlog title
+            // against a these titles as well if applicable
```
EmiEnache: "against these titles"*

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/barb/BbcTxlogsItemUpdaterProvider.java
```diff
@@ -130,8 +125,8 @@ public static BbcTxlogsItemUpdaterProvider create(boolean isSubjectTxlog) {
                 // even if one exists with a bcid. ENG-447
```
EmiEnache: Update this comment to reflect new 10-5 extraction rules.

PR #1468, opened 2020-06-17
Add thematic label to nitro content with the bbc_three masterbrand

obroomhall:


src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroContentExtractor.java
```diff
@@ -113,6 +122,21 @@ public final CONTENT extract(SOURCE source) {
                 log.warn("No master brand mapping found for pid={}, uri={}, masterbrand={}",
                         pid , content.getCanonicalUri(), masterBrand.getMid());
             }
+            if (masterBrand.getMid().equals("bbc_three")) {
+                Optional<Topic> optionalTopic = topicStore.topicFor(LABEL_NAMESPACE, "m0001").toOptional();
+                if (optionalTopic.isPresent()) {
+                    content.addTopicRef(
```
tristanbeardmore-gray, line 128: You don't need to resolve the topic - you can hardcode the topic Id and pass it through in this constructor instead:

`public TopicRef(Long topicId, Float weighting, Boolean supervised, TopicRef.Relationship relationship)`

PR #1469, opened 2020-06-17
Improve implementation of thematic labels

obroomhall:
Move thematic labels to a class with static accessors and caching.

### DIFF
--- a/src/main/java/org/atlasapi/ThematicLabels.java
+++ b/src/main/java/org/atlasapi/ThematicLabels.java
@@ -2,6 +2,8 @@ package org.atlasapi;
 
 import java.util.Optional;
 
+import javax.annotation.PostConstruct;
+
 import org.atlasapi.media.entity.Topic;
 import org.atlasapi.media.entity.TopicRef;
 import org.atlasapi.persistence.topic.TopicStore;
@@ -32,6 +34,11 @@ public class ThematicLabels {
                     return getTopicRefFromThematicValue(value);
                 }
             });
+
+    @PostConstruct
+    private void init() {
+        getInstance();
+    }
 
     public static synchronized ThematicLabels getInstance() {
         if (thematicLabelsInstance == null) {
--- a/src/main/java/org/atlasapi/ThematicLabels.java
+++ b/src/main/java/org/atlasapi/ThematicLabels.java
@@ -12,9 +12,9 @@ import com.google.common.cache.LoadingCache;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
+import org.springframework.context.annotation.Configuration;
 
-@Component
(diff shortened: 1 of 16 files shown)


src/main/java/org/atlasapi/ThematicLabels.java
```diff
@@ -0,0 +1,81 @@
...
+    );
+
+    public static ThematicLabels getInstance() {
+        if (thematicLabelsInstance == null) {
+            thematicLabelsInstance = new ThematicLabels();
+        }
+        return thematicLabelsInstance;
+    }
+
+    public Optional<TopicRef> get(String title) {
```
tristanbeardmore-gray, line 51: Rather than using a string here it would make more sense to have the titles be an enum

src/main/java/org/atlasapi/ThematicLabels.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Optional<Topic> topicOptional = topicStore.topicFor(
+                LABEL_NAMESPACE,
+                value
+        ).toOptional();
+
+        if (!topicOptional.isPresent()) {
+            log.warn(
+                    "Couldn't find thematic label for value {}, the label will "
+                    + "not attached to Nitro content.", value);
```
tristanbeardmore-gray, line 71: Remove reference to Nitro in case it's expanded to other sources

src/main/java/org/atlasapi/ThematicLabels.java
```diff
@@ -0,0 +1,81 @@
...
+        String thematicLabelValue = thematicLabelTitleToValueRelations.get(title);
+        if (thematicLabelValue == null) {
+            log.warn("A thematic label relation could not be found for title {}.", title);
+            return Optional.empty();
+        } else {
+            return thematicLabelCache.getUnchecked(title);
+        }
+    }
+
+    private Optional<TopicRef> getTopicRefFromThematicValue(String value) {
```
tristanbeardmore-gray, line 61: I don't see a use case for this, but if there is you can use the same approach as above (and store the title/value pairs in a BiMap)

src/main/java/org/atlasapi/ThematicLabels.java
```diff
@@ -0,0 +1,81 @@
...
+                public Optional<TopicRef> load(String value) {
+                    return getTopicRefFromThematicValue(value);
+                }
+            });
+
+    private final Map<String, String> thematicLabelTitleToValueRelations = ImmutableMap.of(
+            "bbc_three", "m0001"
+    );
+
+    public static ThematicLabels getInstance() {
```
tristanbeardmore-gray, line 44: This isn't thread-safe so could cause issues if the ingest is multi-threaded

I don't know off the top of my head the best way to do thread-safe singletons but I imagine there are examples online.

tristanbeardmore-gray: One option might just be to make the method synchronised

PR #1471, opened 2020-07-09
Migrate deer-search equiv to Sherlock

obroomhall:
Remove classes and references to elasticsearch and deer. Replace deer search completely with the sherlock client.


src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
```diff
@@ -0,0 +1,172 @@
...
+        this.searcher = checkNotNull(searcher);
+        this.timeout = timeout;
+        this.contentResolver = contentResolver;
+        this.lookupEntryStore = lookupEntryStore;
+        this.idCodec = idCodec;
+    }
+
+    @Override
+    public List<Identified> search(
+            org.atlasapi.search.model.SearchQuery owlQuery,
```
EmiEnache, line 67: Should be "sherlockQuery"

src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
```diff
@@ -0,0 +1,172 @@
...
+            return StreamSupport.stream(input.spliterator(), false)
+                    .map(idCodec::decode)
+                    .map(BigInteger::longValue)
+                    .collect(Collectors.toList());
+        }
+    }
+
+    private List<Identified> resolveIds(List<Long> ids) {
+        Set<LookupEntry> potentialCandidatesSet = ids.stream()
+                .map(this::getLookupEntries) //from atlas
```
EmiEnache, line 149: Minor point, but RE the comment - from owl or deer?

src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
```diff
@@ -0,0 +1,172 @@
...
+            org.atlasapi.search.model.SearchQuery owlQuery,
+            Application application
+    ) {
+        SearchParameter titleSearchParameter = SearchParameter.builder()
+                .withMapping(CONTENT.getTitle())
+                .withValue(owlQuery.getTerm())
+                //.withBoost(owlQuery.getTitleWeighting())
+                .build();
+
+        // TODO Specific weightings are not yet supported in Sherlock
```
tristanbeardmore-gray, line 76: Should we throw exceptions if any of the unsupported fields so that it becomes clear if someone tries to use them that they can't, instead of us silently ignoring them?

src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
obroomhall: No, the owl query gets converted to a sherlock query in this method

src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
obroomhall: We decided not to do this since we are not currently using boostings in equiv, adding these will be part of the migration of search

src/main/java/org/atlasapi/query/content/search/SherlockSearchResolver.java
obroomhall: Addressed.

PR #1472, opened 2020-07-17
Test yv stats

alianos-:


src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -86,4 +91,58 @@ public void statistics(HttpServletRequest request, HttpServletResponse response,
...
+                response.setHeader(HttpHeaders.CONTENT_ENCODING, "gzip");
+                out = new GZIPOutputStream(out);
+            }
+
+            OutputStreamWriter writer = new OutputStreamWriter(out, Charsets.UTF_8);
+
+            try {
+                String name = "YouViewSuccessfulTasks_" + publisherStr + "_" + timespan;
+                writer.write("# TYPE " + name + " gauge\n");
+                writer.write(name + " " + stats.get().successfulTasks());
```
tristanbeardmore-gray, line 134: We also have an alert for too many unsuccessful tasks

src/main/java/org/atlasapi/query/v2/FeedStatsController.java
```diff
@@ -86,4 +91,58 @@ public void statistics(HttpServletRequest request, HttpServletResponse response,
...
+            } finally {
+                Flushables.flushQuietly(writer);
+                if (out instanceof GZIPOutputStream) {
+                    ((GZIPOutputStream) out).finish();
+                }
+            }
+
+
+            modelAndViewFor(request, response, ImmutableSet.of(stats.get()), application);
```
tristanbeardmore-gray, line 143: Will this return a json response? The prometheusMetric endpoint produces plain text

alianos-: oops!

tristanbeardmore-gray: Oh you linked the endpoint and it doesn't seem to be json so this might not be relevant

PR #1474, opened 2020-08-06
RadioTimes Channel Group Automation

obroomhall:
Update the RadioTimes channel group with new channels from PA ingest.

### DIFF
--- a/src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
+++ b/src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
@@ -1,10 +1,12 @@
 package org.atlasapi.remotesite.pa.channels;
 
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -67,7 +69,7 @@ public class PaChannelDataHandler {
         return false;
     };
     private static final Function<Region, String> REGION_TO_URI = Identified::getCanonicalUri;
-    private static final Joiner JOINER = Joiner.on(",");
+    private static final String RADIOTIMES_CHANNEL_GROUP_URI = "http://www.radiotimes.com/platforms/enabled_channels";
 
     private final DateTimeFormatter formatter = ISODateTimeFormat.date();
     private final PaChannelsIngester channelsIngester;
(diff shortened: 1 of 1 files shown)


src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -456,4 +464,45 @@ private void removeExpiredRegionsFromPlatform(Set<String> newRegionUris, Platfor
 
         return combined.build();
     }
+
+    private void updateRadioTimesChannelGroup(List<Channel> channels) {
```
alianos-, line 468: drop a comment here, saying to the best of the authors knowledge, the RT channelGroup is only used so that the the backup script knows for which channels to produce schedule for, which is why the channel numbers are just auto incrementing numbers.

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -456,4 +464,34 @@ private void removeExpiredRegionsFromPlatform(Set<String> newRegionUris, Platfor
 
         return combined.build();
     }
+
+    // To the best of my knowledge, the RadioTimes channel group is only used so that the backup
+    // script knows for which channels to produce schedule for, hence why the channel numbers
```
alianos-, line 469: *backup script = backup-feed script

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -456,4 +464,34 @@ private void removeExpiredRegionsFromPlatform(Set<String> newRegionUris, Platfor
...
+    // assigned are just auto-incrementing numbers, that need not persist through updates.
+    private void updateRadioTimesChannelGroup(List<Channel> channels) {
+
+        Optional<ChannelGroup> maybeChannelGroup = resolveChannelGroup(RADIOTIMES_CHANNEL_GROUP_URI);
+        if (maybeChannelGroup.isPresent()) {
+
+            ChannelGroup channelGroup = maybeChannelGroup.get();
+            channelGroup.setChannelNumberings(ImmutableSet.of());
+
+            int channelNumberToWrite = 1;
```
alianos-, line 479: why not just channelNumber? :p

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
tristanbeardmore-gray: RT seems to be the ones using it, our script doesn't actually use it at all.

src/main/java/org/atlasapi/remotesite/pa/channels/PaChannelDataHandler.java
```diff
@@ -456,4 +464,34 @@ private void removeExpiredRegionsFromPlatform(Set<String> newRegionUris, Platfor
...
+            for (Channel channel : channels) {
```
tristanbeardmore-gray, line 480: You could collect these into a set and then call `setChannelNumberings` instead of having to first set an empty set

obroomhall: I could, but this seemed cleaner

PR #1476, opened 2020-08-12
Ignore broadcasts in the equiv generator for items from tier 2 broadcasters

obroomhall:


src/main/java/org/atlasapi/equiv/generators/barb/ContentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+        ImmutableSet.Builder<ContentProvider> set = ImmutableSet.builder();
+        if (content == null) {
+            return set.build();
+        }
+        ContentProvider cp = fromSource(content.getPublisher().key());
+        if (cp != UNKNOWN) {
+            set.add(cp);
+        }
+        for (String sameAs : content.getAliasUrls()) {
+            cp = fromSource(sameAs);
```
alianos-, line 186: is that from the original class? I don't think that .getAliasUrls will return publisher keys so the .fromSource() will not do anything right.

src/main/java/org/atlasapi/equiv/generators/barb/ContentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+    static {
+        ImmutableMap.Builder<String, ContentProvider> bgidMap = ImmutableMap.builder();
+        ImmutableMap.Builder<String, ContentProvider> sourceBuilder = ImmutableMap.builder();
+        ImmutableMap.Builder<String, ContentProvider> bcidNsMap = ImmutableMap.builder();
+        ImmutableMap.Builder<String, ContentProvider> censusNameMap = ImmutableMap.builder();
+        for (ContentProvider cp : ContentProvider.values()) {
+            if (!cp.getBroadcasterGroupId().isEmpty()) bgidMap.put(cp.getBroadcasterGroupId(), cp);
+            if (!cp.getSource().isEmpty()) sourceBuilder.put(cp.getSource(), cp);
+            for(String bcidNamespace : cp.getBcidNamespaces()) {
+                bcidNsMap.put(bcidNamespace, cp);
```
alianos-, line 136: You are using the BCIDs from the enum, but I don't think this namespaces are the same as the one given to txlogs by the txlog ingester. As I explained yesterday you need to get the BGID and map the broadcaster group to a source.

src/main/java/org/atlasapi/equiv/generators/barb/ContentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+
+            if (value.isEmpty()) return value;
+            return val.startsWith("C4:") ? val : "C4:".concat(val);
+        }
+    },
+    C5(     T1, "4", "datasubmission.channel5.com",  ImmutableSet.of("gb:c5:bcid"),           "C5"),
+    UKTV(   T1, "63", "uktv.co.uk",                   ImmutableSet.of("gb:uktv:bcid"),         "UKTV"),
+    //TODO S4C(    T2, "",                             ImmutableSet.of(),                     "S4C"),
+    SKY(    T2, "5", "",                             ImmutableSet.of("gb:barb:broadcastGroup:5:bcid"),          "Sky"), //will match as an alias
+    UNKNOWN(T2, "", "",                             ImmutableSet.of(),                     "");
```
alianos-, line 58: Is there any point in having non T1 publishers in the enum. Is there a point in having all these values in the enum? Keep in mind someone will have to maintain this, why all this complexity?

src/main/java/org/atlasapi/equiv/generators/barb/ContentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+        content.getAliases().stream()
+                .map(Alias::getNamespace)
+                .map(BGID_FROM_ALIAS::matcher)
+                .filter(Matcher::matches)
+                .map(matcher -> matcher.group(1))
+                .forEach(bgid -> addIfValid(set, bgid));
+
+        content.getAliases().stream()
+                .map(Alias::getNamespace)
+                .map(OOBGID_FROM_ALIAS::matcher)
```
alianos-, line 207: that is plain wrong, since if the OOBGID is from tier1 you would have no expectation whatsoever to match this against a CMS broadcast.

src/main/java/org/atlasapi/equiv/generators/barb/ContentProvider.java
```diff
@@ -0,0 +1,235 @@
...
+                .filter(Matcher::matches)
+                .map(matcher -> matcher.group(1))
+                .forEach(bgid -> addIfValid(set, bgid));
+
+        String bgid = content.getCustomField(TXLOG_BROADCASTER_GROUP);
+        addIfValid(set, bgid);
+        bgid = content.getCustomField(NLE_BROADCASTER_GROUP);
+        addIfValid(set, bgid);
+        bgid = content.getCustomField(CDMF_BROADCASTER_GROUP);
+        addIfValid(set, bgid);
```
alianos-, line 217: NLE and CDMF will never even be considered, as they have no broadcasts.

src/main/java/org/atlasapi/equiv/generators/barb/TieredBroadcaster.java
```diff
@@ -0,0 +1,48 @@
...
+
+        return (publisher != null && tierOneBroadcasterGroups.containsValue(publisher.key()))
+               || (bgid != null && tierOneBroadcasterGroups.containsKey(bgid));
+    }
+
+    public static Optional<String> getSource(String bgid) {
+        return Optional.ofNullable(tierOneBroadcasterGroups.get(bgid));
+    }
+
+    public static Optional<String> getBroadcasterGroupId(String source) {
```
tristanbeardmore-gray, line 42: I don't think you're using this but a better implementation of this would be to use a BiMap instead of a Map internally, which then lets you do lookups on values as well as keys.

src/test/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorerTest.java
```diff
@@ -140,6 +142,7 @@ public void testGenerateEquivalencesForOneMatchingBroadcast() {
                 new Broadcast(BBC_ONE.getUri(), time("2014-03-21T15:00:00Z"), time("2014-03-21T16:00:00Z")),
                 new Broadcast(BBC_ONE_CAMBRIDGE.getUri(), time("2014-03-21T15:00:00Z"), time("2014-03-21T16:00:00Z"))
         );
+        item1.addCustomField(TXLOG_BROADCASTER_GROUP, "1"); // since we now ignore tier 2 broadcasts
```
tristanbeardmore-gray, line 145: Change the publisher from PA to BARB_TRANSMISSIONS to better reflect the actual equiv logic.

src/test/java/org/atlasapi/equiv/generators/barb/BarbBroadcastMatchingItemEquivalenceGeneratorAndScorerTest.java
```diff
@@ -596,12 +603,15 @@ public void testGenerateEquivalencesForBbcTwoEnglandTxlogVariants() {
                 BBC_NITRO,
                 new Broadcast(BBC_TWO_ENGLAND.getUri(), time("2014-03-21T15:00:00Z"), time("2014-03-21T16:00:00Z"))
         );
+        nitroItem.addCustomField(TXLOG_BROADCASTER_GROUP, "1"); // since we now ignore tier 2 broadcasts
```
tristanbeardmore-gray, line 606: Don't add these to the Nitro ones because the Nitro content won't have this custom field.

src/main/java/org/atlasapi/equiv/generators/barb/TieredBroadcaster.java
obroomhall: Interesting, had not seen bimaps until this point

PR #1477, opened 2020-08-19
Transform v4 aliases from simple to complex channel

EmiEnache:


EmiEnache:
This issue was preventing us from adding aliases to simple model channels that we were trying to ingest using atlas-client (e.g. in ingest-youview).

PR #1485, opened 2020-10-21
Improve C4 film equiv

obroomhall:
Create an updater provider based on the ImdbContainerUpdaterProvider to help with the C4 problem (films are ingested as a single brand and episode).


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -1287,7 +1288,7 @@ private static UpdaterConfiguration makeC4PmlsdConfiguration() {
                 Sets.difference(
                         TARGET_SOURCES,
                         //Top level containers from the following sources are explicitly equived to by editors
```
tristanbeardmore-gray, line 1290: `                        //Top level containers from the following sources (except PA) are explicitly equived to by editors`

PR #1495, opened 2021-01-19
Fix channel image merge npe

EmiEnache:


EmiEnache:
We decided against this. The ingest-pa code will instead mimic file-based PA ingest with regards to image – that is, the "Normal" type of image will be saved with the "Light Opaque" theme.

PR #1496, opened 2021-01-22
Eng 982

EmiEnache:
Adds owl api ChannelGroup write support for regions under platform, and platform under regions.

### DIFF
--- a/src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
+++ b/src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
@@ -31,6 +31,7 @@ import org.atlasapi.media.channel.ChannelGroupType;
 import org.atlasapi.media.channel.ChannelNumbering;
 import org.atlasapi.media.channel.ChannelResolver;
 import org.atlasapi.media.channel.Platform;
+import org.atlasapi.media.channel.Region;
 import org.atlasapi.media.entity.Publisher;
 import org.atlasapi.output.Annotation;
 import org.atlasapi.output.AtlasErrorSummary;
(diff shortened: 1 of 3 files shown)


src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -457,6 +448,55 @@ private WriteResponseWithOldIds createOrUpdateChannelGroup(
...
+                .filter(region -> !Strings.isNullOrEmpty(region.getUri()))
+                .forEach(region -> {
+                    Optional<ChannelGroup> innerRegionOptional =
+                            channelGroupResolver.channelGroupFor(region.getUri());
+                    if(innerRegionOptional.isPresent()) {
+                        region.setId(idCodec.encode(BigInteger.valueOf(innerRegionOptional.get().getId())));
+                        resolvedRegions.add(region);
+                    }
+                });
+        simplePlatform.setRegions(resolvedRegions);
```
tristanbeardmore-gray, line 482: If you pass in regions with ids but without uris these are going to be filtered out

src/main/java/org/atlasapi/query/v2/ChannelGroupController.java
```diff
@@ -457,6 +448,55 @@ private WriteResponseWithOldIds createOrUpdateChannelGroup(
...
+     *  initial ChannelGroup.
+     *
+     * @param simpleChannelGroup - potentially modified by this method
+     */
+    private void fillInInnerChannelGroupIds(org.atlasapi.media.entity.simple.ChannelGroup simpleChannelGroup) {
+        if("platform".equals(simpleChannelGroup.getType())) {
+            updateInnerRegions(simpleChannelGroup);
+        }
+        else {
+            updateInnerPlaform(simpleChannelGroup);
```
tristanbeardmore-gray, line 465: To avoid people thinking these might actually update in the database I'd rename these to help make clear they just insert ids.

PR #1500, opened 2021-02-08
Eng 974

IvanKaraslavov:
Equivs A+E data to txlogs


src/main/java/org/atlasapi/equiv/scorers/barb/BarbTitleMatchingItemScorer.java
```diff
@@ -276,14 +291,17 @@ public Score score(Item subject, Item suggestion, ResultDescription desc) {
             nonTxlogItem = subject;
         }
 
+        String txlogTitle = null;
+        if(useTxlogEpisodeTitle) {
+            txlogTitle = txlogItem.getTitle();
+            String episodeTitle = txlogItem.getCustomField(TXLOG_EPISODE_TITLE_CUSTOM_FIELD_NAME);
+            txlogItem.setTitle(episodeTitle);
```
tristanbeardmore-gray, line 298: I'd probably take a copy of the txlogItem and set the title on the copy just in case the original is used later and ends up having its title mutated.

tristanbeardmore-gray: There should be a `.copy()` method on the item.

src/main/java/org/atlasapi/equiv/generators/aenetworks/AeBroadcastMatchingItemEquivalenceGeneratorAndScorer.java
```diff
@@ -0,0 +1,415 @@
...
+            ResultDescription desc,
+            EquivToTelescopeResult equivToTelescopeResult
+    ) {
+
+        DefaultScoredCandidates.Builder<Item> scores = DefaultScoredCandidates.fromSource("A+E Broadcast");
+
+        EquivToTelescopeComponent generatorComponent = EquivToTelescopeComponent.create();
+        generatorComponent.setComponentName("A+E Broadcast Matching Item Equivalence Generator");
+
+        if (!TieredBroadcaster.isAAndE(subject)) {
```
tristanbeardmore-gray, line 81: You probably don't need this any more since we'll always be running this on the a+e content

src/main/java/org/atlasapi/equiv/generators/aenetworks/TieredBroadcaster.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.google.common.collect.BiMap;
+import com.google.common.collect.ImmutableBiMap;
+import org.atlasapi.media.entity.Content;
+import org.atlasapi.media.entity.Publisher;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+public class TieredBroadcaster {
```
tristanbeardmore-gray, line 11: This class won't be needed I believe if the only use is the `isAAndE` check I commented on

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -0,0 +1,750 @@
...
+        }
+
+    }
+
+    private AeTitleMatchingItemScorer(Builder builder) {
+        scoreOnPerfectMatch = checkNotNull(builder.scoreOnPerfectMatch);
+        scoreOnPartialMatch = checkNotNull(builder.scoreOnPartialMatch);
+        scoreOnMismatch = checkNotNull(builder.scoreOnMismatch);
+        contentResolver = builder.contentResolver;
+        topLevelContainerCache = CacheBuilder.newBuilder()
```
tristanbeardmore-gray, line 117: You can probably remove the container/series cache and scoring logic since I believe we don't have containers and series for a+e content right?

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -0,0 +1,750 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.api.client.repackaged.com.google.common.base.Preconditions.checkNotNull;
+import static org.atlasapi.equiv.utils.barb.BarbEquivUtils.TXLOG_PUBLISHERS;
+
+/**
+ * Caveat: Txlog names are scored against a suggestion's brand title and series title as well,
+ * along with their permutations so do not rely to heavily on this score for equiv.
```
EmiEnache, line 37: too* not "to"

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -0,0 +1,750 @@
...
+        String suggTitle = normalize(suggestionTitle);
+
+        if (appearsToBeWithApostrophe(subjectTitle)) {
+            String regexp = normalizeRegularExpression(subjectTitle);
+            matches = Pattern.matches(regexp, suggTitle);
+        } else if (appearsToBeWithApostrophe(suggestionTitle)) {
+            String regexp = normalizeRegularExpression(suggestionTitle);
+            matches = Pattern.matches(regexp, subjTitle);
+        } else {
+            matches = matchWithoutDashes(subjTitle, suggTitle) || subjTitle.equals(suggTitle) || (subjTitle.startsWith(suggTitle) || suggTitle.startsWith(subjTitle));
```
EmiEnache, line 464: This returns true if either `subjTitle.startsWith(suggTitle)` or `suggTitle.startsWith(subjTitle)`

So if you have let's say:
subj A title: "Fast and Furious"
and
sugg B title: "Fast and Furious 2"

this results in matches = true, meaning that in line 470 below this would score it as if it was a perfect match.

Assuming the above is a correct interpretation, then is this truly what we want?

PR #1503, opened 2021-02-12
Eng 988

EmiEnache:
Allows ingest thru owl client/API of zero duration broadcasts.
This relates to new PA API data, which is just like old PA
except ingested thru client, and the BroadcastAssertions
prevented broadcasts where startTime == endTime (for
no particular reason; the inner overlap checks are safe
for this kind of cases)

### DIFF
--- a/src/main/java/org/atlasapi/query/content/merge/BroadcastMerger.java
+++ b/src/main/java/org/atlasapi/query/content/merge/BroadcastMerger.java
@@ -179,7 +179,7 @@ public class BroadcastMerger {
 
             checkNotNull(from);
             checkNotNull(to);
-            checkArgument(from.isBefore(to));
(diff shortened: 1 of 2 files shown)


src/main/java/org/atlasapi/query/content/merge/BroadcastMerger.java
```diff
@@ -179,7 +179,7 @@ private BroadcastAssertion(String channelUri, DateTime from, DateTime to) {
 
             checkNotNull(from);
             checkNotNull(to);
-            checkArgument(from.isBefore(to));
+            checkArgument(from.isBefore(to) || from.isEqual(to));
```
tristanbeardmore-gray, line 182: `checkArgument(!from.isAfter(to));`

PR #1510, opened 2021-03-17
ENG-1003

IvanKaraslavov:


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4AtomEntryVersionExtractor.java
```diff
@@ -133,6 +131,13 @@ private Location extractLocation(String uri, C4VersionData data, Long playerId,
             location.addAliasUrl(data.getId().replace("tag:pmlsc", "tag:www"));
         }
         location.setTransportType(TransportType.LINK);
+
+        // Adds provider information only to locations with uris in the format of
+        // http://www.channel4.com/programmes/a-new-life-in-the-sun/on-demand/70296-021
+        // because of efficiency
+        if(uri.contains("channel4")) {
```
tristanbeardmore-gray, line 138: It's probably better if this is "channel4.com" in case they add links in future that are like:
"http://www.newProvider.com/channel4/someShow"

PR #1511, opened 2021-03-18
break equivalence for sources that don't need it

alianos-:


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -172,14 +167,41 @@ private static UpdaterConfiguration makeDefaultConfiguration(Publisher publisher
...
-                                STANDARD_ITEM, MoreSets.add(TARGET_SOURCES, LOVEFILM)
+                                NO_ITEM, ImmutableSet.of()
                         ),
                         STANDARD_ITEM_HANDLER,
                         STANDARD_ITEM_MESSENGER
                 )
                 .withTopLevelContainerEquivalenceUpdater(
                         ImmutableMap.of(
-                                STANDARD_TOP_LEVEL_CONTAINER, MoreSets.add(TARGET_SOURCES, LOVEFILM)
+                                NOP_CONTAINER, ImmutableSet.of()
```
tristanbeardmore-gray, line 177: You renamed NOP_ITEM to NO_ITEM but haven't changed NOP_CONTAINER

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -102,21 +96,22 @@ public static UpdaterConfigurationRegistry create() {
                 makeNitroConfiguration(),
                 makeC4PmlsdConfiguration(),
                 makeUktvConfiguration(),
-                makeWikipediaConfiguration(),
+//                makeWikipediaConfiguration(),
```
tristanbeardmore-gray, line 99: Why have you removed some but commented out others? And although you've removed some here why have you left the actual method intact in all cases?

src/main/java/org/atlasapi/equiv/update/updaters/configuration/DefaultConfiguration.java
```diff
@@ -163,22 +38,39 @@
     // default equiv on. However over time we have found that default equiv generally doesn't meet our purposes and
     // helps contribute to large equiv sets.
     public static final ImmutableSet<Publisher> DEFAULT_EQUIV_SOURCES = ImmutableSet.of(
```
EmiEnache, line 40: this method and/or this comment should clarify that these are sources which will have no outgoing equiv

alianos-: Arguably these are still the sources that we run default equiv on, which changed from something to nothing! i.e. the comment should be on the method that runs default equiv. Regardless though it is an academic point cause once we break them all we'll just remove the list alltogether!

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -539,30 +395,6 @@ private void addEquivalenceJobs(Set<ScheduledTask> jobsAtStartup) {
                 EBS_SPORTS_EQUIVALENCE_REPETITION,
                 jobsAtStartup
         );
-        scheduleEquivalenceJob(
```
EmiEnache, line 542: Don't think this should be removed

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -700,42 +511,42 @@ private void addYouViewScheduleEquivalenceJobs(Set<ScheduledTask> jobsAtStartup)
...
-                YOUVIEW_STAGE_SCHEDULE_EQUIVALENCE_REPETITION
-        );
-        // This job is scheduled for late in the day so run it for +8 days to ensure we get +7 day
-        // coverage from the next day onwards
-        scheduleEquivalenceJob(
-                taskBuilder(0, 8)
-                        .withPublishers(YOUVIEW_SCOTLAND_RADIO)
-                        .withChannelsSupplier(youviewChannelsSupplier())
-                        .build()
-                        .withName("YouView Scotland Radio Schedule Equivalence (8 day) Updater"),
```
EmiEnache, line 727: This is being used, so should not be removed

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -700,42 +511,42 @@ private void addYouViewScheduleEquivalenceJobs(Set<ScheduledTask> jobsAtStartup)
...
-                YOUVIEW_SCHEDULE_EQUIVALENCE_REPETITION
-        );
-        scheduleEquivalenceJob(
-                taskBuilder(0, 7)
-                        .withPublishers(YOUVIEW_SCOTLAND_RADIO_STAGE)
-                        .withChannelsSupplier(youviewChannelsSupplier())
-                        .build()
-                        .withName(
-                                "YouView Stage Scotland Radio Schedule Equivalence (8 day) Updater"),
```
EmiEnache, line 736: same as above

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
EmiEnache: The data is definitely used, but there is a possibility equiv is not relevant anymore? Would need looking into https://metabroadcast.atlassian.net/browse/MTS-1484

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
```diff
@@ -302,102 +239,21 @@ private void addEquivalenceJobs(Set<ScheduledTask> jobsAtStartup) {
...
-                RepetitionRules.NEVER,
-                jobsAtStartup
-        );
         scheduleEquivalenceJob(
                 publisherUpdateTask(YOUVIEW).withName("YouView Equivalence Updater"),
                 RepetitionRules.NEVER,
                 jobsAtStartup
         );
-        scheduleEquivalenceJob(
-                publisherUpdateTask(YOUVIEW_STAGE).withName("YouView Stage Equivalence Updater"),
```
EmiEnache, line 346: While this does not have equiv now, it might in the future, as the source is being used. So if you're leaving YOUVIEW source equiv here, I'd leave this as well.

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -113,30 +136,9 @@ public static UpdaterConfigurationRegistry create() {
...
-                makeBtVodConfiguration(),
-                makeBtTveVodConfiguration()
-        );
-
         configurations.add(
                 makeYouviewConfiguration(),
-                makeYouviewStageConfiguration(),
-                makeYouviewBtConfiguration(),
-                makeYouviewBtStageConfiguration(),
-                makeYouviewScotlandRadioConfiguration(),
```
EmiEnache, line 126: leave this config in, as source is being used

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -113,30 +136,9 @@ public static UpdaterConfigurationRegistry create() {
...
-                makeYouviewScotlandRadioStageConfiguration()
```
EmiEnache, line 127: as above

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -255,37 +223,6 @@ private static UpdaterConfiguration makeBarbXMasterConfiguration() {
                 .build();
     }
 
-    private static UpdaterConfiguration makeImdbConfiguration() {
```
EmiEnache, line 258: Please don't remove this, as it will be added back soon enough (albeit reworked somewhat)

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -462,295 +336,26 @@ private static UpdaterConfiguration makeYouviewConfiguration() {
...
-                        ImmutableMap.of(
-                                NOP_CONTAINER, ImmutableSet.of()
-                        ),
-                        NOP_CONTAINER_HANDLER,
-                        NOP_CONTAINER_MESSENGER
-                )
-                .build();
-    }
-
-    private static UpdaterConfiguration makeYouviewScotlandRadioConfiguration() {
```
EmiEnache, line 572: this could probably stay

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -462,295 +336,26 @@ private static UpdaterConfiguration makeYouviewConfiguration() {
...
-                        ImmutableMap.of(
-                                NOP_CONTAINER, ImmutableSet.of()
-                        ),
-                        NOP_CONTAINER_HANDLER,
-                        NOP_CONTAINER_MESSENGER
-                )
-                .build();
-    }
-
-    private static UpdaterConfiguration makeYouviewScotlandRadioStageConfiguration() {
```
EmiEnache, line 608: this could probably stay too

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
alianos-: it should. no keys have access to the source.

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
alianos-: sorry my bad. I left the method call commented out but blindly removed the method!

src/main/java/org/atlasapi/equiv/EquivTaskModule.java
alianos-: For future generations. It looks like it should be used but wasn't, so we added that to YV's key and put the code back in.

PR #1517, opened 2021-04-14
Fix episode equiv

EmiEnache:


src/main/java/org/atlasapi/equiv/generators/SherlockTitleSearchGenerator.java
```diff
@@ -0,0 +1,279 @@
...
+            return;
+        }
+        if (content instanceof Episode) {
+            titleQuery.addFilter(TermsParameter.of(CONTENT.getType(), ImmutableSet.of(subjectType, "item")));
+            return;
+        }
+        if (content instanceof Film) {
+            titleQuery.addFilter(TermsParameter.of(CONTENT.getType(), ImmutableSet.of(subjectType, "item")));
+        }
+        if (content instanceof Item) {
```
tristanbeardmore-gray, line 259: This will also happen for Films

PR #1525, opened 2021-04-28
Add exact description match scorer, use it for PA API to fix dupes

EmiEnache:


src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesItemEquivalenceGenerator.java
```diff
@@ -149,8 +166,14 @@ private void addResult(
             ResultDescription desc, Builder<Item> result,
             EquivToTelescopeComponent generatorComponent,
             Container resolvedContainer,
-            Item child) {
+            Item child,
+            Item subject
+    ) {
 
+        if (includeSiblings && child.equals(subject)) {
```
tristanbeardmore-gray, line 173: This check seems to be covered by lines 139-141

src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesItemEquivalenceGenerator.java
```diff
@@ -103,8 +116,12 @@ public ContainerCandidatesItemEquivalenceGenerator(
             }
         }
 
-
-        for (String containerUri : getContainerUris(subject.getContainer())) {
+        ImmutableList.Builder<String> containerUrisBuilder = ImmutableList.builder();
```
tristanbeardmore-gray, line 119: I'd use an ImmutableSet here instead as a precaution against duplicates since otherwise we'd be resolving containers more than necessary.

src/main/java/org/atlasapi/equiv/results/extractors/TypeBasedExtractor.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public Set<ScoredCandidate<T>> extract(
+            List<ScoredCandidate<T>> candidates,
+            T target,
+            ResultDescription desc,
+            EquivToTelescopeResult equivToTelescopeResult
+    ) {
+        EquivToTelescopeComponent extractorComponent = EquivToTelescopeComponent.create();
+        extractorComponent.setComponentName("Size Matching Extractor");
```
tristanbeardmore-gray, line 54: Name seems wrong

src/main/java/org/atlasapi/equiv/results/extractors/TypeBasedExtractor.java
```diff
@@ -0,0 +1,81 @@
...
+
+        if (candidates.isEmpty()) {
+            return ImmutableSet.of();
+        }
+
+        desc.startStage("Extracting from delegate");
+
+        // fetch delegate based on subject type - with care for episodes/films being items
+        Set<ScoredCandidate<T>> extractedCandidates = delegates.get(target.getClass()).extract(
```
tristanbeardmore-gray, line 63: This could NPE if the class doesn't exist in the map

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -1,56 +1,55 @@
...
+ * If shortest not long enough, quits early. Needs exact match to score.
  */
-@Deprecated
 public class DescriptionMatchingScorer<T extends Described> implements EquivalenceScorer<T> {
 
-    public static final String NAME = "Description Matching";
-    private static final Set<String> commonWords = ImmutableSet.of(
-            "the", "in", "a", "and", "&", "of", "to", "show"
-    );
+    @VisibleForTesting
```
tristanbeardmore-gray, line 26: Surely you could just choose a value to be used in the unit test instead of making this value visible?

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -83,57 +93,89 @@ private DescriptionMatchingScorer() {
...
+        }
+        else if (!Strings.isNullOrEmpty(subject.getMediumDescription())
+                && subject.getMediumDescription().length() > MINIMUM_DESCRIPTION_LENGTH) {
+            return true;
+        }
+        else if (!Strings.isNullOrEmpty(subject.getShortDescription())
+                && subject.getShortDescription().length() > MINIMUM_DESCRIPTION_LENGTH){
+            return true;
+        }
+        return false;
```
tristanbeardmore-gray, line 109: Why don't we check the `description` field? There are 4 fields in total

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -83,57 +93,89 @@ private DescriptionMatchingScorer() {
...
-            return tokenize(description).stream()
-            .filter(s -> !commonWords.contains(s))
-            .collect(Collectors.toSet());
-        } else {
-            return Sets.newHashSet();
+        if (!score.equals(perfectMatch)) {
+            descriptionToCompare = subject.getShortDescription();
+            score = compare(descriptionToCompare, candidate.getShortDescription());
+        }
+
```
tristanbeardmore-gray, line 144: Should we not check the `description` field too?

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/StandardItemSequenceUpdaterProvider.java
```diff
@@ -48,13 +52,17 @@ public static StandardItemSequenceUpdaterProvider create() {
...
                                 )
                         )
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new SequenceItemScorer(Score.ONE)
+                                new SequenceItemScorer(Score.ONE),
+                                // the following scorers were added to prevent equiv'ing to bad data
+                                // from subject-source (e.g. multiple S1E1's with different title/description)
+                                new SelfPublisherOnlyItemScorer()
```
tristanbeardmore-gray, line 65: If this scores at least 1 then we can equiv things even if the series/episode numbers do not match. Is that intended or an oversight?

tristanbeardmore-gray, line 65: I think the internal scorers used within this class should be passed in specifically here. It would help make it clear that this either scores 0 or negative values.
Otherwise it's not clear that this is effectively a negative scorer (since otherwise it looks like matching on series/episode number would be enough to equiv).

src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesItemEquivalenceGenerator.java
EmiEnache: Less "covered" and more "prevents this from ever being true", which is not what we want.
I will remove the following id equals check occuring on lines 139-141:

!Objects.equals(
                        child.getId(),
                        subject.getId()

tristanbeardmore-gray: I don't follow, why do we want this here? It's just preventing adding itself as a candidate which is already enforced in the other code.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
EmiEnache: But if I did that, then if one changes the default scores, then the unit tests will fail...
... unless of course the test class sets the scores it should give. D'oh. I'll do that.

EmiEnache: But if I did that, then if one changes the default scores, then the unit tests will fail...
... unless of course the test class sets the scores it should give. D'oh. I'll do that.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
EmiEnache: We didn't use it because we different ingests fill in that field differently... e.g. PA description is long, but PA API description is short (or some such thing).

But you have a point, doesn't hurt to add it as well as a kind of fallback, so I will

src/main/java/org/atlasapi/equiv/generators/ContainerCandidatesItemEquivalenceGenerator.java
EmiEnache: I misread! You're right, this check is entirely pointless. Will remove.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/StandardItemSequenceUpdaterProvider.java
EmiEnache: I will pass the scores in, but not the scorers since using generic scorers is a pain.

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -83,57 +107,105 @@ private DescriptionMatchingScorer() {
...
+                );
+            }
+        }
+        return equivalents.build();
     }
 
-    private boolean descriptionMatch(T subject, T candidate) {
-        Set<String> candidateList = descriptionToProcessedList(candidate.getDescription());
-        Set<String> subjectList = descriptionToProcessedList(subject.getDescription());
+    // visible for delegate
```
alianos-, line 145: hm?

src/main/java/org/atlasapi/equiv/scorers/DescriptionMatchingScorer.java
```diff
@@ -83,57 +107,105 @@ private DescriptionMatchingScorer() {
...
+        }
 
-    private Set<String> descriptionToProcessedList(String description) {
-        if (!Strings.isNullOrEmpty(description)) {
-            return tokenize(description).stream()
-            .filter(s -> !commonWords.contains(s))
-            .collect(Collectors.toSet());
-        } else {
-            return Sets.newHashSet();
+        if (!score.equals(perfectMatch)) {
```
alianos-, line 163: why do we keep going until there is a perfect match? Maybe you found a missmatch already and you should return no?

EmiEnache: There might be a match further down, amongst the shorter descriptions. We've seen them not be consistent WRT whether or not both duplicates have a long description for example, but they end up having the same medium description. 
(having said that, we haven't seen a case of different long descriptions but same medium description, but still made me want to implement it this way)

EmiEnache: As discussed, we will change this to equiv more defensively e.g. score mismatch if any found, instead of digging deeper for a perfect match

PR #1526, opened 2021-05-04
Add a quality field in the simple model

IvanKaraslavov:


src/main/java/org/atlasapi/input/EncodingModelTransformer.java
```diff
@@ -56,7 +53,16 @@ private Encoding encodingFrom(org.atlasapi.media.entity.simple.Location inputLoc
         encoding.setVideoHorizontalSize(inputLocation.getVideoHorizontalSize());
         encoding.setVideoProgressiveScan(inputLocation.getVideoProgressiveScan());
         encoding.setVideoVerticalSize(inputLocation.getVideoVerticalSize());
+        if(inputLocation.getHighDefinition() != null && inputLocation.getQuality() != null) {
```
EmiEnache, line 56: use upperCase not lowerCase

PR #1528, opened 2021-05-07
Fill in series number on nitro episodes from parent series

EmiEnache:


src/main/java/org/atlasapi/remotesite/bbc/nitro/extract/NitroEpisodeExtractor.java
```diff
@@ -168,7 +169,14 @@ protected void extractAdditionalItemFields(NitroItemSource<Episode> source, Item
             if (position != null) {
                 episodeContent.setEpisodeNumber(position.intValue());
             }
-            episodeContent.setSeriesRef(getSeriesRef(episode));
```
tristanbeardmore-gray, line 171: Technically we never set this value as 'null' any more if the if statement is not entered; this might be fine if it is not set elsewhere, otherwise we might not be unsetting a value that is meant to be unset.

PR #1530, opened 2021-05-10
New sequence matching

EmiEnache:


src/main/java/org/atlasapi/equiv/scorers/SequenceItemScorer.java
```diff
@@ -48,11 +70,13 @@ public SequenceItemScorer(Score matchingScore) {
                         .withMatchingScore(matchingScore)
                         .withMissingScore(Score.nullScore())
                         .withMismatchScore(Score.nullScore())
+                        .withAllowNullSeries(false)
+                        .withTitlesMustMatch(false)
+                        .withOnlyMatchCanHaveTitle(false)
+                        .withMissingEpisodeNumberIsMismatch(false)
                         .withRequireBothOrNeitherChildOfTopLevelSeries(true)
-                        .withMissingEpisodeNumberIsMismatch(true)
```
tristanbeardmore-gray, line 52: Why have we changed default behaviour?

EmiEnache: This was an accident

PR #1532, opened 2021-05-12
Make SoleCandidate scorer account for equivs

EmiEnache:


src/main/java/org/atlasapi/equiv/scorers/SoleSubjectAndCandidateTitleMatchingScorer.java
```diff
@@ -184,6 +205,24 @@ private void addScore(
                 .collect(MoreCollectors.toImmutableList());
     }
 
+    // checks that all of contentToCheck are equiv'd to contenntToCheckAgainst
+    private boolean allAreEquivdTo(Iterable<T> contentToCheck, T contentToCheckAgainst) {
+        LookupRef lookupRef = LookupRef.from(contentToCheckAgainst);
+        for (T pubCandidate : contentToCheck) {
```
tristanbeardmore-gray, line 211: Just resolve the lookup entry for 'contentToCheckAgainst' and see if any of `contentToCheck` aren't in the direct/explicits. That way we only resolve one entry instead of several.

src/main/java/org/atlasapi/equiv/scorers/SoleSubjectAndCandidateTitleMatchingScorer.java
```diff
@@ -184,6 +205,24 @@ private void addScore(
                 .collect(MoreCollectors.toImmutableList());
     }
 
+    // checks that all of contentToCheck are equiv'd to contenntToCheckAgainst
+    private boolean allAreEquivdTo(Iterable<T> contentToCheck, T contentToCheckAgainst) {
```
tristanbeardmore-gray, line 209:
```suggestion
    private boolean allAreEquived(T contentToCheck, Iterable<T> possibleEquivs) {
```

src/main/java/org/atlasapi/equiv/scorers/SoleSubjectAndCandidateTitleMatchingScorer.java
```diff
@@ -122,7 +135,15 @@ public SoleCandidateTitleMatchingScorer(
                 addScore(publisherCandidates.iterator().next(), exactTitleMatchScore, equivalents, scorerComponent);
             } else {
                 for (T publisherCandidate : publisherCandidates) {
-                    addScore(publisherCandidate, mismatchScore, equivalents, scorerComponent);
+                    if (!allAreEquivdTo(publisherCandidates, publisherCandidate)) {
```
tristanbeardmore-gray, line 138: Don't do this check multiple times; do the check once with the first candidate as one parameter and all other candidates as the second. If the check succeeds score all of them as match, otherwise as mismatch.

PR #1533, opened 2021-05-13
Add version merging logic for containers

IvanKaraslavov:
The series and brands didn't have merging logic for their versions. The complex model was not updated with the locations from the simple model. The versions were also added in the container simplifier when the corresponding LOCATIONS or AVAILABLE_LOCATIONS annotations are added to the request.


src/main/java/org/atlasapi/output/simple/ContainerModelSimplifier.java
```diff
@@ -5,15 +5,11 @@
...
-import org.atlasapi.media.entity.ChildRef;
-import org.atlasapi.media.entity.Container;
-import org.atlasapi.media.entity.EntityType;
-import org.atlasapi.media.entity.Item;
-import org.atlasapi.media.entity.ParentRef;
-import org.atlasapi.media.entity.Series;
-import org.atlasapi.media.entity.SeriesRef;
+import com.metabroadcast.common.intl.Countries;
+import com.metabroadcast.common.time.DateTimeZones;
+import org.atlasapi.media.entity.*;
```
EmiEnache, line 12: your IDE seems to be handling these differently from ours

PR #1535, opened 2021-05-14
Fix item search equiv, also cleans up some equivs/classes

EmiEnache:
- switch imdb equiv to use standard for PA + PA API 
(they were the same)
- delete old, obsolete IMDb equiv UpdaterProviders
- remove MinimumScoreFilter where redundant
- StandardItemSearch equiv now has a SequenceScorer 
which doesn’t score matches, but does negatively score
mismatches (to prevent episodes with exact same name
from different brands equiv’ing if they have different 
series/episode numbers)


src/main/java/org/atlasapi/equiv/update/updaters/providers/item/StandardItemSequenceUpdaterProvider.java
```diff
@@ -85,7 +83,6 @@ public static StandardItemSequenceUpdaterProvider create() {
                 )
                 .withFilter(
                         ConjunctiveFilter.valueOf(ImmutableList.of(
-                                new MinimumScoreFilter<>(0.99),
```
tristanbeardmore-gray, line 88: This isn't redundant since the extractor doesn't require a minimum

EmiEnache: Pardon me – good catch!

PR #1542, opened 2021-05-25
Update existing equiv to account for blacklisting+explicit

EmiEnache:


src/main/java/org/atlasapi/equiv/generators/ContainerEquivalentsItemEquivalenceGenerator.java
```diff
@@ -184,23 +173,22 @@ private void addResult(
...
-                    .map(ContentRef::getCanonicalUri)
-                    .distinct()
-                    .collect(MoreCollectors.toImmutableList());
-        } else {
-            return summary.getCandidates();
-        }
+    private ImmutableList<String> fetchContainerUris(
+            LookupEntry lookupEntry,
+            @Nullable Set<Publisher> publishers,
+            String parentUri
```
tristanbeardmore-gray, line 183: This shouldn't be needed since it should be accessible from lookupEntry

PR #1554, opened 2021-06-24
Always take series into consideration in standard item sequence equiv

EmiEnache:


src/main/java/org/atlasapi/equiv/generators/ContainerEquivalentsItemEquivalenceGenerator.java
```diff
@@ -82,6 +87,7 @@ public static Builder builder(){
             if (subject instanceof Episode) {
                 epSubject = (Episode) subject;
                 if (epSubject.getSeriesNumber() == null || epSubject.getEpisodeNumber() == null) {
+                    // TODO shouldn't return here, since filters shouldn't be X-OR
```
tristanbeardmore-gray, line 90: Is this a TODO that needs doing?

EmiEnache: Yes, but I do not have atm time to examine the implications and be confident it won't cause issues, so I will leave the TODO for another time (and will add it to jira - https://metabroadcast.atlassian.net/browse/ENG-1086)

PR #1556, opened 2021-07-13
ENG-1081

IvanKaraslavov:
Adds a key field in BT Channels


src/main/java/org/atlasapi/remotesite/bt/channels/BtChannelDataUpdater.java
```diff
@@ -256,14 +257,22 @@ private Channel findOrCreateSourceChannel(Channel channel) {
                 codec.encode(BigInteger.valueOf(channel.getId()))
         );
 
-        return channelResolver.fromUri(baseChannelUri).valueOrDefault(
+        Maybe<Channel> maybeChannel = channelResolver.fromUri(baseChannelUri);
+        if(maybeChannel.hasValue()) {
+            Channel existingChannel = maybeChannel.requireValue();
+            existingChannel.setKey(existingChannel.getUri());
```
tristanbeardmore-gray, line 263: Since we decided to update the database I don't think this step is necessary.

PR #1557, opened 2021-07-14
Merge sameAs on channels to not nuke equivalence on writes

EmiEnache:


src/main/java/org/atlasapi/query/v2/ChannelWriteExecutor.java
```diff
@@ -377,8 +377,12 @@ private WriteResponse deserializeAndUpdateChannel(HttpServletRequest req, HttpSe
             );
         }
 
+        Optional<Channel> existingChannel = store.fromUri(channel.getCanonicalUri()).toOptional();
+        //Preserves the equiv list ENG-1082
+        if(existingChannel.isPresent()) {
```
alianos-, line 382: I would hope "merge" would do that already, so to avoid doing it twice this should be an "else if" to the merge.

tristanbeardmore-gray: It doesn't already since the channel API doesn't support any kind of equiv writing. This step should realistically happen regardless of whether merge is true.

PR #1559, opened 2021-08-03
ENG-1095

IvanKaraslavov:
Changes the nitro merging strategy for aliases to retain the old version pids


src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -308,6 +313,18 @@ public Content mergeTopics(Content current, Content extracted) {
         }
     }
 
+    private static class BbcNitroKeepVersionAliases implements AliasMergeStrategy {
+
+        @Override
+        public Item mergeAliases(Item current, Item extracted) {
+            Set<Alias> currentVersionAliases = current.getAliases().stream()
+                    .filter(alias -> alias.getNamespace().equals(VERSION_PID_NAMESPACE) || alias.getNamespace().equals(ORIGINAL_VERSION_PID_NAMESPACE))
+                    .collect(Collectors.toSet());
```
alianos-, line 322: If this code is duplicated, try to prevent duplication. If that requires significant refactoring, leave a comment in both places where the code is duplicated.

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -308,6 +313,18 @@ public Content mergeTopics(Content current, Content extracted) {
         }
     }
 
+    private static class BbcNitroKeepVersionAliases implements AliasMergeStrategy {
```
alianos-, line 316: leave a comment with the ticket number here.

src/main/java/org/atlasapi/remotesite/ContentMerger.java
```diff
@@ -308,6 +313,18 @@ public Content mergeTopics(Content current, Content extracted) {
...
+
+        @Override
+        public Item mergeAliases(Item current, Item extracted) {
+            Set<Alias> currentVersionAliases = current.getAliases().stream()
+                    .filter(alias -> alias.getNamespace().equals(VERSION_PID_NAMESPACE) || alias.getNamespace().equals(ORIGINAL_VERSION_PID_NAMESPACE))
+                    .collect(Collectors.toSet());
+            current.setAliases(Sets.union(currentVersionAliases, extracted.getAliases()));
```
alianos-, line 323: do we know for sure that the `extracted` has had its aliases calculated already, or it relies in random order of things?

PR #1560, opened 2021-08-03
Add siblingsOnly option; switch to using Sherlock search query instead

EmiEnache:
The option to look at subject siblingsOnly makes sense
for episodes, since you wouldn't want equiv to stop
early if another brand has an episode with same name
e.g. ("The Beginning", or "Episode 10").

Using Sherlock SearchQuery instead of Owl allows
increased flexibility in the query (since it was ending
up being converted into that anyway).


src/main/java/org/atlasapi/equiv/scorers/SoleSubjectAndCandidateTitleMatchingScorer.java
```diff
@@ -53,17 +55,20 @@
 public class SoleSubjectAndCandidateTitleMatchingScorer<T extends Content> implements EquivalenceScorer<T>{
 
     public static final String NAME = "Sole Candidate Title Match";
+    private static final ContentMapping CONTENT = IndexMapping.getContentMapping();
 
     private final ContentTitleScorer<T> titleScorer;
-    private final SearchResolver searchResolver;
+    private final SherlockSearchResolver searchResolver;
     private final MongoLookupEntryStore lookupEntryStore;
+    private final boolean compareTitleToSiblingsOnly;
```
tristanbeardmore-gray, line 63:
```suggestion
    private final boolean compareSubjectToSiblingsOnly;
```

PR #1566, opened 2021-10-01
ENG-1112

IvanKaraslavov:
This PR updates the RT equiv by:
1) adding equiv rules from RT to PA and PA API using the RT id, ingested as alias, as the basis of this equiv
2) exposing the target publishers for RT_ITEM. There was a hidden logic in RT_ITEM which was equiving RT to PA without exposing the PA publisher. This PR consolidate the item to be more like others and clearly equiving to PA
3) removing RT-> JustWatch equiv
4) adding RT-> IMDB equiv rules based on title+year


src/main/java/org/atlasapi/equiv/generators/RadioTimesFilmEquivalenceGenerator.java
```diff
@@ -41,9 +45,14 @@ public RadioTimesFilmEquivalenceGenerator(ContentResolver resolver) {
 
         EquivToTelescopeComponent generatorComponent = EquivToTelescopeComponent.create();
         generatorComponent.setComponentName("Radio Times Film Equivalence Generator");
-        
+
         Builder<Item> results = DefaultScoredCandidates.fromSource("RTtoPA");
-        
+
+        if(StreamSupport.stream(publishers.spliterator(), false).count() != 1 || !publishers.iterator().next().equals(Publisher.PA)) {
```
tristanbeardmore-gray, line 51: This check should be in the constructor and just needs to check that PA is there, not that there's only one target publisher.

src/main/java/org/atlasapi/equiv/generators/RadioTimesFilmEquivalenceGenerator.java
```diff
@@ -26,9 +28,11 @@
     private final String paFilmUriPrefix = "http://pressassociation.com/films/";
     
     private final ContentResolver resolver;
+    private final Iterable<Publisher> publishers;
 
-    public RadioTimesFilmEquivalenceGenerator(ContentResolver resolver) {
+    public RadioTimesFilmEquivalenceGenerator(ContentResolver resolver, Iterable<Publisher> publishers) {
```
tristanbeardmore-gray, line 33: It'll make your life easier using a Set instead of an iterable here, especially since the class that instantiates this will be passing in a Set anyway.

src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -319,8 +321,11 @@ private static UpdaterConfiguration makeRadioTimesConfiguration() {
                 .withSource(RADIO_TIMES)
                 .withItemEquivalenceUpdater(
                         ImmutableMap.of(
-                                RT_ITEM, ImmutableSet.of(AMAZON_UNBOX),
-                                IMDB_ITEM, ImmutableSet.of(IMDB, JUSTWATCH)    //equiv via IMDb id
+                                RT_ITEM, ImmutableSet.of(PA), // the logic in this item can only equiv to PA, the old RT -> PA will be kept as there are a lot of PA content doesn't have v4 aliases
```
EmiEnache, line 324: the logic in this item can only equiv to PA, the old RT -> PA will be kept as there are a lot of PA content doesn't have v4 aliases
should be
the logic in this can only equiv to PA; this old RT -> PA will be kept as there is a lot of PA content that doesn't have v4 aliases, so we cannot rely on those.

src/main/java/org/atlasapi/equiv/update/updaters/providers/item/RtItemUpdaterProvider.java
```diff
@@ -46,15 +46,8 @@ public static RtItemUpdaterProvider create() {
                 .withGenerators(
                         ImmutableSet.of(
                                 new RadioTimesFilmEquivalenceGenerator(
-                                        dependencies.getContentResolver()
```
EmiEnache, line 49: Also consider adding some Javadoc to RtItemUpdaterProvider to explain why it's important to keep as is  ie. why it cannot just be superseded by the alias-based equiv

PR #1571, opened 2021-12-07
ENG-1149

IvanKaraslavov:
Adds release dates on content level so all types of content can have them.


src/main/java/org/atlasapi/input/ContentModelTransformer.java
```diff
@@ -270,4 +274,20 @@ private CrewMember transformPerson(Person person, Publisher publisher) {
...
+            ReleaseDate complexReleaseDate = new ReleaseDate(convertToLocalDate(releaseDate.getDate()),
+                    Countries.fromCode(releaseDate.getCountry()),
+                    ReleaseDate.ReleaseType.valueOf(releaseDate.getType().toUpperCase()));
+            builder.add(complexReleaseDate);
+        }
+        return builder.build();
+    }
+
+    private LocalDate convertToLocalDate(Date date) {
+        return new LocalDate(date);
```
tristanbeardmore-gray, line 291: Why did you remove the null check?

src/main/java/org/atlasapi/query/content/merge/ContentMerger.java
```diff
@@ -257,4 +262,13 @@ private Content merge(
     private <T> List<T> merge(List<T> existing, List<T> posted) {
         return ImmutableSet.copyOf(Iterables.concat(posted, existing)).asList();
     }
+
+    public Item mergeReleaseDates(Item existing, Item update, boolean merge) {
```
tristanbeardmore-gray, line 266: It doesn't look like this is used any more.

src/main/java/org/atlasapi/input/ContentModelTransformer.java
tristanbeardmore-gray: The complex `ReleaseDate` does enforce the date to be non-null but you're better off letting that handle raising the exception instead of inadvertently NPEing here.

pawel-mtbcst: @tristanbeardmore-gray if we're sure it won't be null, why write the code that is not needed?
We could add @nonNull annotation on the param. Also, this method can be static, or even can be moved to some Utility class.

src/main/java/org/atlasapi/query/content/merge/ContentMerger.java
```diff
@@ -257,4 +262,13 @@ private Content merge(
     private <T> List<T> merge(List<T> existing, List<T> posted) {
         return ImmutableSet.copyOf(Iterables.concat(posted, existing)).asList();
```
pawel-mtbcst, line 263: Alternative using JDK methods, that does not create set just to remove duplicates:
`concat(existing.stream(), posted.stream()).distinct().collect(toList())`

src/main/java/org/atlasapi/input/ContentModelTransformer.java
tristanbeardmore-gray: This class essentially converts between two very similar but different data models (from the 'simple' model to the  'complex' model). It looks like the date field is nullable for the simple model but not the complex. I don't consider it this class's responsibility to care for this difference, and the validation can be left within the simple or complex model class itself.

That said this is a rather minor point, especially because the exception that would be thrown otherwise from the complex model class doesn't really provide any additional information about the exception (unless the IllegalArgumentException that gets thrown is caught and converted into a meaningful HTTP response whilst the NPE possibly doesn't - but I don't believe this is the case).

PR #1579, opened 2022-01-20
C4 ingest films as films

EmiEnache:


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandUpdateController.java
```diff
@@ -59,4 +76,102 @@ public void updateBrand(HttpServletResponse resp,
...
+            HttpServletResponse resp,
+            @RequestParam("uri") String uri
+    ) throws IOException {
+
+        C4BrandUpdater updater = dflt;
+
+        if (!updater.canFetch(uri)) {
+            resp.sendError(400, "Can't update URI " + uri);
+            return;
+        }
```
tristanbeardmore-gray, line 93: You removed this block for the other endpoint; is it intentional to have kept it here?

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandUpdateController.java
```diff
@@ -59,4 +76,102 @@ public void updateBrand(HttpServletResponse resp,
...
+                .toOptional()
+                .filter(Item.class::isInstance)
+                .map(Item.class::cast);
+
+        if (!optionalEpisode.isPresent()) {
+            resp.sendError(404, String.format("Cannot find episode %s", uri));
+            return;
+        }
+
+        Item episode = optionalEpisode.get();
```
tristanbeardmore-gray, line 132: Might be worth adding a check that this is an Episode

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandUpdateController.java
```diff
@@ -59,4 +76,102 @@ public void updateBrand(HttpServletResponse resp,
...
+        if (!episode.getGenres().contains(AtlasGenre.FILM.getUri())) {
+            resp.sendError(406, String.format("Episode is not actually a film: %s", uri));
+            return;
+        }
+
+        Film film = createAsFilm(episode);
```
tristanbeardmore-gray, line 138: It looks like this might not retain the container parent field which would orphan the content

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandUpdateController.java
EmiEnache: Yes it's intentional, since it helps check that it's a C4 brand. 
I removed it from the other block since there's no existing "is this a c4 episode?" check. I will add an episode check in the other block.

src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandUpdateController.java
EmiEnache: Does it not? The copy method within it does
`        to.setParentRef(from.getContainer());`

EmiEnache: Where "to" is the film object that gets returned by "createAsFilm", and "from" is the episode passed into it.

tristanbeardmore-gray: Sorry, I'm blind; was scanning for `setContainer`

PR #1584, opened 2022-03-01
Unpublish task module

IvanKaraslavov:


src/main/java/org/atlasapi/equiv/update/tasks/ContentUnpublishingTask.java
```diff
@@ -0,0 +1,323 @@
...
+        if(!unpublishAllContentFromSource) {
+            publisher = Publisher.IMDB;
+            try {
+                Set<String> imdbIdsToUnpublish = Sets.newHashSet(FileUtils.readLines(new File(IMDB_IDS_TO_UNPUBLISH)));
+
+                log.info("Unpublishing content from {}", publisher);
+                Queue<String> contentBatch = getUriBatchImdb(uriIterator, imdbIdsToUnpublish);
+                while (!contentBatch.isEmpty()) {
+                    if(!shouldContinue()) {
+                        unpublishContent(publisher, contentBatch, true);
```
tristanbeardmore-gray, line 153: I don't see the benefit of this, it looks like it tries to finish one additional batch when it doesn't really need to?

PR #1587, opened 2022-05-13
ENG-1191

IvanKaraslavov:
Adds a new unpublishing mode which unpublishes not only the piece of content but its hierarchy. What the new mode is doing is:
1. Unpublish all children of the piece of content 
* if it is a brand then unpublish all its series and episodes
* if it is a series then unpublish all its episodes
3. Unpublish the parent if all its children are unpublished 
* if it is an episode and the series (or brand) doesn't have any other actively published children then unpublish the parent
* if it is a series and the brand doesn't have any other actively published children then unpublish the brand


src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+
+        // Check if the api key has write permissions for this publisher
+        validateApiKey(described.getPublisher(), possibleApplication);
+
+        Optional<String> optionalUnpublishHierarchy =
+                req.getParameter(UNPUBLISH_HIERARCHY) != null
+                        ? Optional.of(req.getParameter(UNPUBLISH_HIERARCHY))
+                        : Optional.empty();
+        boolean unpublishHierarchy =
+                optionalUnpublishHierarchy.filter(Boolean::parseBoolean).isPresent();
```
tristanbeardmore-gray, line 116:
```suggestion
        boolean unpublishHierarchy = Boolean.parseBoolean(req.getParameter(UNPUBLISH_HIERARCHY));
```

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+
```
tristanbeardmore-gray, line 117: Add a check which throws an error if `unpublishHierarchy` is set to true when the mode is set to publish content.

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+        if (unpublishHierarchy && !status) {
+            unpublishHierarchy(lookupEntry, described);
+        } else {
+            updateContent(status, lookupEntry, described);
+        }
+    }
+
+    private void unpublishHierarchy(LookupEntry lookupEntry, Described described) {
```
tristanbeardmore-gray, line 125: This method should call individual methods that handle each case to prevent this method becoming too large.

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+            Episode episode = (Episode) described;
+            // check if this was the last actively published episode from the series
+            if(episode.getSeriesRef() != null && !Strings.isNullOrEmpty(episode.getSeriesRef().getUri())) {
+                List<Identified> episodeSiblings = mongoContentResolver.findSiblingsForEpisodeOnSeriesLevel(described.getCanonicalUri()).getAllResolvedResults();
+                if(episodeSiblings.isEmpty()) {
+                    Maybe<Identified> maybeSeries = contentResolver.findByCanonicalUris(Collections.singletonList(episode.getSeriesRef().getUri())).getFirstValue();
+                    if(maybeSeries.hasValue()) {
+                        updateContent(false, lookupEntry, (Described) maybeSeries.requireValue());
+                        //check if this was the last actively published episode from the brand
+                        unpublishBrandIfNecessary(lookupEntry, episode);
```
tristanbeardmore-gray, line 174: What about if there are series siblings that aren't unpublished? We seem to account for this in the other case but not here.

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+            for (Identified identifiedEpisode : episodesFromSeries) {
+                updateContent(false, lookupEntry, (Described) identifiedEpisode);
+            }
+
+            // get actively published series siblings
+            List<Identified> seriesSiblings = mongoContentResolver.findSiblingsForSeries(described.getCanonicalUri()).getAllResolvedResults();
+            if(seriesSiblings.isEmpty()) {
+                Series series = (Series) described;
+                // get all episodes of that brand (there might be episodes without series)
+                List<Identified> episodesFromBrand = mongoContentResolver.findChildrenForParent(series.getParent().getUri(), true).getAllResolvedResults();
```
tristanbeardmore-gray, line 155: What's different here compared to `unpublishBrandIfNecessary`? Do we need an `unpublishSeriesIfNecessary` too?

src/main/java/org/atlasapi/system/UnpublishContentController.java
```diff
@@ -1,58 +1,42 @@
...
-        this.idCodec = checkNotNull(idCodec);
-        this.contentResolver = checkNotNull(contentResolver);
-        this.lookupEntryStore = checkNotNull(lookupEntryStore);
-        this.contentWriter = checkNotNull(contentWriter);
-        this.noEquivalenceContentWriter = checkNotNull(noEquivalenceContentWriter);
-        this.equivalenceBreaker = checkNotNull(equivalenceBreaker);
+            EquivalenceBreaker equivalenceBreaker,
+            MongoContentResolver mongoContentResolver) {
+        this.publishedStatusUpdater =
+                new PublishedStatusUpdater(
```
tristanbeardmore-gray, line 32: Make the constructor accept a `PublishedStatusUpdater` instead of instantiating it here to reduce the number of constructor parameters needed.

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    private void unpublishHierarchy(LookupEntry lookupEntry, Described described) {
+        if(lookupEntry.lookupRef().category().equals(ContentCategory.CONTAINER)) {
+            // get all episodes of that brand
+            List<Identified> episodesFromBrand = mongoContentResolver.findChildrenForParent(described.getCanonicalUri(), true).getAllResolvedResults();
+            // get all series of that brand
+            List<Identified> series = mongoContentResolver.findSeriesForBrand(described.getCanonicalUri()).getAllResolvedResults();
+
+            // unpublish episodes
```
tristanbeardmore-gray, line 132: This is going to end up being horrendously slow if a large brand needs to be unpublished. This might not need changing now but this should be designed with a Kafka queue in mind that does the unpublishing in the background.

tristanbeardmore-gray (changes requested):

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
tristanbeardmore-gray: Are you sure this is actually resolved?

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -124,62 +151,83 @@ public void setPublishStatusOfItem(
...
-            }
+            unpublishBrandHierarchy(lookupEntry, described);
         } else if(lookupEntry.lookupRef().category().equals(ContentCategory.PROGRAMME_GROUP)) {
-            // get all episodes of that series
-            List<Identified> episodesFromSeries = mongoContentResolver.findChildrenForParent(described.getCanonicalUri(), false).getAllResolvedResults();
+            unpublishSeriesHierarchy(lookupEntry, described);
+        } else if(lookupEntry.lookupRef().category().equals(ContentCategory.CHILD_ITEM)) {
+            unpublishEpisodeHierarchy(lookupEntry, (Episode) described);
+        }
+        updateContent(false, lookupEntry, described);
```
tristanbeardmore-gray, line 160: I think this needs to be moved within the specific hierarchy unpublishing methods since it will interfere with the checks some of those do to check all children have been unpublished. It should happen after unpublishing any children of the subject content and before any attempts to unpublish the subject's parents.

tristanbeardmore-gray (changes requested):

src/main/java/org/atlasapi/system/PublishedStatusUpdater.java
```diff
@@ -0,0 +1,424 @@
...
+                Maybe<Identified> maybeBrand = contentResolver.findByCanonicalUris(Collections.singletonList(series.getParent().getUri())).getFirstValue();
+                if(maybeBrand.hasValue()) {
+                    updateContent(false, lookupEntry, (Described) maybeBrand.requireValue());
+                }
+            }
+        }
+    }
+
+    private void unpublishBrandHierarchy(LookupEntry lookupEntry, Described described) {
+        //TODO: Add Kafka queue that does the unpublishing in the background
```
EmiEnache, line 201: Is this intentionally left behind, to be done another time?

IvanKaraslavov: Yeah, it is intentional, for future improvement regarding unpublishing big brands with a lot of children

PR #1588, opened 2022-05-18
ENG-1175

IvanKaraslavov:
This PR makes brand title-based equiv stricter as there are some false positives. The equiv rules for brands are:

* when comparing the title brands the stopwords (e.g. ‘the', 'and’ etc) should not be striped
* doesn't remove spaces
* some symbols need to be converted to words and not striped from the title (e.g. ‘&', convert to "and")
* the titles need to be compared with and without those symbols
* comparing titles with and without hyphens


src/main/java/org/atlasapi/equiv/update/updaters/providers/container/BroadcastItemContainerUpdaterProvider.java
```diff
@@ -60,7 +60,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                                 )
                         ))
                 .withScorers(
-                        ImmutableSet.of(new TitleMatchingContainerScorer(2)))
+                        ImmutableSet.of(TitleMatchingContainerScorer.forTopLevelContainer(2)))
```
tristanbeardmore-gray, line 63: This is changing behaviour of YouView->BBC equiv which is out of scope of the original planned work. This might be an improvement but it would need separate testing directly on the YouView and BBC data.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/BtVodContainerUpdaterProvider.java
```diff
@@ -57,7 +57,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new TitleMatchingContainerScorer(2),
+                                TitleMatchingContainerScorer.forNonTopLevelContainer(2),
```
tristanbeardmore-gray, line 60: This preserves existing behaviour but use a new creator method mentioned in a separate comment to avoid any confusion over its purpose since it's not clear if this class is only meant for non-top-level containers. Consider just removing the class as well since it's not really worth maintaining old unused equiv classes at this point.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/RtUpcomingContainerUpdaterProvider.java
```diff
@@ -66,7 +66,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.<EquivalenceScorer<Container>>of(
-                                new TitleMatchingContainerScorer(2)
+                                TitleMatchingContainerScorer.forNonTopLevelContainer(2)
```
tristanbeardmore-gray, line 69: This preserves existing behaviour this method name ends up being misleading since this class is used for both top-level and non-top level containers. Create a new creator method that avoids this confusion.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/RteContainerUpdaterProvider.java
```diff
@@ -57,7 +57,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new TitleMatchingContainerScorer(2),
+                                TitleMatchingContainerScorer.forNonTopLevelContainer(2),
```
tristanbeardmore-gray, line 60: This preserves existing behaviour but use a new creator method mentioned in a separate comment to avoid any confusion over its purpose since it's not clear if this class is only meant for non-top-level containers. Consider just removing the class as well since it's not really worth maintaining old unused equiv classes at this point.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/WikipediaContainerUpdateProvider.java
```diff
@@ -54,7 +54,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new TitleMatchingContainerScorer(2.0),
+                                TitleMatchingContainerScorer.forNonTopLevelContainer(2.0),
```
tristanbeardmore-gray, line 57: This preserves existing behaviour but use a new creator method mentioned in a separate comment to avoid any confusion over its purpose. Consider just removing the class as well since it's not really worth maintaining old unused equiv classes at this point.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/amazon/AmazonToAmazonContainerUpdaterProvider.java
```diff
@@ -56,7 +55,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new TitleMatchingContainerScorer(2)
+                                TitleMatchingContainerScorer.forTopLevelContainer(2)
```
tristanbeardmore-gray, line 58: Don't change Amazon<->Amazon equiv. Any changes would need approval and wouldn't be worth the time to change and test.

src/main/java/org/atlasapi/equiv/update/updaters/providers/container/amazon/AmazonToAmazonSeriesUpdaterProvider.java
```diff
@@ -55,7 +54,7 @@ public EquivalenceResultUpdater<Container> getUpdater(
                 )
                 .withScorers(
                         ImmutableSet.of(
-                                new TitleMatchingContainerScorer(2),
+                                TitleMatchingContainerScorer.forNonTopLevelContainer(2),
```
tristanbeardmore-gray, line 57: As long as this preserves existing behaviour then this is fine.

src/main/java/org/atlasapi/equiv/scorers/SoleSubjectAndCandidateTitleMatchingScorer.java
```diff
@@ -104,7 +102,7 @@ private SoleSubjectAndCandidateTitleMatchingScorer(Builder<T> builder) {
         mismatchScore = checkNotNull(builder.mismatchScore);
         classFilter = checkNotNull(builder.classFilter);
 
-        this.titleScorer = new ContentTitleScorer<>(
+        this.titleScorer = new StandardContentTitleScorer<>(
```
tristanbeardmore-gray, line 105: This class most likely needs logic to use the new top level container scorer when trying to score top level containers, in addition to the standard one for the other cases.

src/main/java/org/atlasapi/equiv/utils/TitleScorerUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Nonnull
+    public static String convertSomeSymbolsToWords(@Nonnull String title) {
+        return title.replace(" & ", " and ");
+    }
+
+    @Nonnull
+    public static String stripAllSymbolsFromTitle(@Nonnull String title) {
+        return title.replaceAll("[^\\d\\w\\s]", "");
```
tristanbeardmore-gray, line 21: I think this is too aggressive and will completely remove accented characters or other characters from non-latin alphabets.

tristanbeardmore-gray (changes requested):

src/main/java/org/atlasapi/equiv/utils/TitleScorerUtils.java
```diff
@@ -17,8 +17,8 @@ public static String convertSomeSymbolsToWords(@Nonnull String title) {
     }
 
     @Nonnull
-    public static String stripAllSymbolsFromTitle(@Nonnull String title) {
-        return title.replaceAll("[^\\d\\w\\s]", "");
+    public static String stripPunctuationAndSymbolsFromTitle(@Nonnull String title) {
+        return title.replaceAll("[\\p{Punct}\\p{S}\\s]", "");
```
tristanbeardmore-gray, line 21: I'm not sure replacing all whitespace is sensible - is there a specific case we've observed that this fixes? If the issue is just different whitespace characters (e.g. tabs versus spaces) or different amounts of whitespace (e.g. 2 spaces between words instead of one) then we could simplify replace all consecutive whitespace characters with a single space.

src/main/java/org/atlasapi/equiv/scorers/TopLevelContainerTitleScorer.java
```diff
@@ -49,90 +49,163 @@ public Score score(Container subject, Container candidate, ResultDescription des
...
+        if (subject.getTitle() == null || candidate.getTitle() == null) {
+            return Score.nullScore();
+        }
+
+        BestSubjectCandidateTitles bestSubjectCandidateTitles = getBestSubjectCandidateTitles(subject, candidate);
+
+        return bestSubjectCandidateTitles.getScore();
+    }
+
+    private BestSubjectCandidateTitles getBestSubjectCandidateTitles(
```
tristanbeardmore-gray, line 69: You can make this a lot cleaner:
1. Rename `BestSubjectCandidateTitles` to something like `ScoredTitles` - we'll use this to keep track of which pair of titles received which score.
2. Introduce a `private ScoredTitles scoreSpecificTitles(String subjectTitle, String candidateTitle)` which does the specific title comparison.
3. Introduce another method for each combination we want to do e.g. `private ScoredTitles scoreTitlesWithSymbolsConvertedToWords(String subjectTitle, String candidateTitle)` which does the title transformation and passes them to the method from 2.
4. This method now just calls each of the methods from 3, adds them to a list of `ScoredTitles` and iterates through it to find the one with maximum score (it can use a variable of type `ScoredTitles` initially null to keep track for example).

tristanbeardmore-gray (changes requested):

PR #1591, opened 2022-05-31
ENG-1176

IvanKaraslavov:
Removes the ContainerChildEquivalence scoring part from some of the providers (StandardTopLevelContainerUpdaterProvider, ImdbContainerUpdaterProvider, OldStandardTopLevelContainerUpdaterProvider) in order to remove the deadlock problem (once equived some brands can no longer automatically break the equivalence because their children are already equived).


src/main/java/org/atlasapi/equiv/generators/ContainerChildEquivalenceGeneratorAndScorer.java
```diff
@@ -33,6 +33,8 @@
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Multiset;
 
+import static org.atlasapi.equiv.results.scores.Score.nullScore;
+
 /**
  * Finds & scores candidates by looking at subject's children, then finding the content these children are equiv'd to,
```
EmiEnache, line 39: Add note to this comment that scoring is optional / can be disabled

PR #1593, opened 2022-08-11
ENG-1357

IvanKaraslavov:
Adds broadcast v4 aliases to the simplifier and increases the max request header size


src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
```diff
@@ -367,6 +368,14 @@ private org.atlasapi.media.entity.simple.Broadcast simplify(
         simpleModel.setNewSeries(broadcast.getNewSeries());
         simpleModel.setNewEpisode(broadcast.getNewEpisode());
         simpleModel.setAliases(broadcast.getAliasUrls());
+        simpleModel.setV4Aliases(
+                ImmutableSet.copyOf(
```
pawel-mtbcst, line 372: If you're copying this collection just so that customers cannot modify the underlying set, you don't have to - `Collectors.toSet()` will create a new instance of the set anyway. 
If that's just a defensive programming technique, you can simplify this to:
`.collect(collectingAndThen(toSet(), ImmutableSet::copyOf)));`

this can be even simpler in newer JDKs:
https://www.baeldung.com/java-stream-immutable-collection

src/main/java/org/atlasapi/AtlasMain.java
```diff
@@ -157,7 +157,7 @@ private void createServerConnector(
 
     private HttpConnectionFactory createHttpConnectionFactory() {
         HttpConfiguration config = new HttpConfiguration();
-        config.setRequestHeaderSize(8192);
+        config.setRequestHeaderSize(32000);
```
pawel-mtbcst, line 160: Personally, I would set it to `32768` just so that it's round.

src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
tristanbeardmore-gray: You can also collect to ImmutableSets directly (since we're using an old version of Guava we have a utility method that does it - `MoreCollectors.toImmutableSet`).

src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
```diff
@@ -367,6 +368,14 @@ private org.atlasapi.media.entity.simple.Broadcast simplify(
...
+                        broadcast.getAliases().stream()
```
tristanbeardmore-gray, line 373: There's likely some code duplication here - see if there's any code from `IdentifiedModelSimplifier` that can be reused for converting the Aliases.

tristanbeardmore-gray (approved):

PR #1595, opened 2022-09-07
ENG-1416

IvanKaraslavov:
Improves the equiv rules between A+E Networks and txlogs by stripping the stopword 'the' and converting '&' to 'and'


src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -565,6 +565,10 @@ private String processTitle(String title, Map<String, String> customFields, bool
             title = title.trim().replaceAll(episodeInTitleMatcher.group(1), "").trim();
         }
 
+        //Removes the stopword 'the'
+        title = removeStopWordsPrefixes(title);
+
+        title = title.replace(" & ", " and ");
```
tristanbeardmore-gray, line 571: Also make a method for "replacements" like this to help improve readability and keep method sizes small, especially if we end up adding more replacements in future.

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -565,6 +565,10 @@ private String processTitle(String title, Map<String, String> customFields, bool
             title = title.trim().replaceAll(episodeInTitleMatcher.group(1), "").trim();
         }
 
+        //Removes the stopword 'the'
```
tristanbeardmore-gray, line 568: This comment isn't really necessary and just risks being misleading if the method is ever changed without the comment being updated. You're also using a comment when javadoc would have been more appropriate.

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
```diff
@@ -593,6 +597,10 @@ private String processTitle(String title, Map<String, String> customFields, bool
         return title.trim();
     }
 
+    private String removeStopWordsPrefixes(String title) {
+        return (title.startsWith("the ") ? title.substring(4) : title);
```
tristanbeardmore-gray, line 601: A slightly nicer way to achieve this would be by creating a list (or ordered Set) of the stop words and iterating through it; this is so that for some `String stopWord` you would be able to get its length just with `stopWord.length()` and avoid accidentally making a mistake with the manual substring call. It would help avoid errors in future if more stop words were needed.

src/test/java/org/atlasapi/equiv/scorers/AeTitleMatchingItemScorerTest.java
```diff
@@ -67,6 +67,7 @@ public void testAeNetworksTxlogCustomRuleExamples() {
         assertTxlogAeScore("New: World's Deadliest Weather", "World's Deadliest Weather S1 - - Episode 7", "3", scoreOnPartialMatch);
         assertTxlogAeScore("New: World's Deadliest Weather", "", "3", scoreOnMismatch);
         assertTxlogAeScore("New: The Drone Wars", "Storage Wars 2 - 2 - Drone Wars 24, The", "24", scoreOnMatch);
+        assertTxlogAeScore("New: Palm Trees & Face Palms", "Storage Wars 2 - 2 - Palm Trees And Face Palms 24, The", "24", scoreOnMatch);
```
tristanbeardmore-gray, line 70: It's generally best to have two separate test cases for the two different logic changes you've made, instead of having one catch both. It's generally also better to have each case be in a separate unit test method with descriptive name to make it clearer what the test is meant to be doing.

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
tristanbeardmore-gray: Seeing as there's actually already prefix stripping happening elsewhere, and this is specially to strip the title prefix after potentially moving the word "The" to the start, perhaps we should rename the method to indiciate that it's only to remove the word "The".

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
tristanbeardmore-gray: I say that, but also just noticed this method is doing all kinds of replacements with different kinds of logic.

src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java
pawel-mtbcst: @IvanKaraslavov @tristanbeardmore-gray 
You can also use
`StringUtils.removeStart()` or `StringUtils.removeStartIgnoreCase()` from apache commons library. 

```
Removes a substring only if it is at the beginning of a source string, otherwise returns the source string.
A null source string will return null. An empty ("") source string will return the empty string. A null search string will return the source string.
 StringUtils.removeStart(null, *)      = null
 StringUtils.removeStart("", *)        = ""
 StringUtils.removeStart(*, null)      = *
 StringUtils.removeStart("www.domain.com", "www.")   = "domain.com"
 StringUtils.removeStart("domain.com", "www.")       = "domain.com"
 StringUtils.removeStart("www.domain.com", "domain") = "www.domain.com"
 StringUtils.removeStart("abc", "")    = "abc"
 ```

File: src/main/java/org/atlasapi/equiv/scorers/aenetworks/AeTitleMatchingItemScorer.java [hunk 5]
```diff
@@ -430,7 +430,7 @@ private String removePrefixes(String title, Set<String> prefixes) {
         }
         while (removedAtLeastOne);
 
-        return remainingTitle;
+        return removeStopWordsPrefixes(remainingTitle);
     }
 
     private String regularExpressionReplaceSpecialChars(String title) {
```
pawel-mtbcst, line 436: I know it's not strictly related to your changes, but why does the name of this method mention that it internally uses regular expressions?
It's just an implementation detail, it's not important for a client of this method whether it achieves its goal using regular expressions, ASCII dictionary, or any other way. 
The client is only interested in the end result of this operation. This should be renamed to `removeSpecial()`

PR #1596, opened 2022-11-10
ENG-1344 SVOD Enrichment Atlas Model Changes

pawel-mtbcst:


src/main/java/org/atlasapi/input/Content2ModelTransformer.java
```diff
@@ -0,0 +1,76 @@
...
+        super(lookupStore, topicStore, idCodec, clipsModelTransformer, clock, imageModelTransformer);
+        this.encodingTransformer = EncodingModelTransformer.create();
+        this.restrictionTransformer = RestrictionModelTransformer.create();
+    }
+
+    @Override
+    protected T setFields(T result, F simple) {
+        super.setFields(result, simple);
+        result.setCountriesOfOrigin(simple.getCountriesOfOrigin());
+        result.setProductions(simple.getProductions().stream().map(Production::copy).collect(toSet()));
```
tristanbeardmore-gray, line 38: You might need to collect to ImmutableSets here for your new fields to preserve order. We need sets to preserve order in order to prevent content being marked as changed when it hasn't when it is written to the database (admittedly there's likely a better solution for this, but I'm mentioning it here as something to be mindful of).

src/main/java/org/atlasapi/output/simple/ItemModelSimplifier.java
```diff
@@ -213,6 +205,14 @@ private void copyProperties(Item fullItem, org.atlasapi.media.entity.simple.Item
             simpleItem.setScheduleOnly(fullItem.isScheduleOnly());
         }
 
+        if(annotations.contains(Annotation.DISTRIBUTION)) {
```
tristanbeardmore-gray, line 208: Can these be in ContentModelSimplifier?

src/main/java/org/atlasapi/output/simple/ContainerModelSimplifier.java
```diff
@@ -149,6 +143,14 @@ public Playlist simplify(Container fullPlayList, Set<Annotation> annotations, Ap
                 simplePlaylist.setBrandSummary(summaryFromResolved(series.getParent(), annotations));
             }
         }
+
+        if(annotations.contains(Annotation.DISTRIBUTION)) {
```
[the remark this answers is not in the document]
pawel-mtbcst: `ContentModelSimplifier` operates on `Description` from a simple model containing neither distribution nor production fields.
I could change it so that `ContentModelSimplifier` operates on `MediaContent` instead of `Description`, which will be fine from the `atlas` repository perspective (as there are only two subclasses: for `complex Container -> simple Playlist` and `Item -> Item`, both inheriting from `MediaContent`), however, this will introduce a risk that in some other repository we have another subclass of `ContentModelSimplifier` that will break.
I thought of that before and considered extracting an additional level of Simplifier, which will keep backward compatibility and allow us to extract common logic, I just figured that it won't be as beneficial as with the `Transformer` case, so I just repeated those 4 lines.

src/main/java/org/atlasapi/query/content/merge/ContentMerger.java
```diff
@@ -223,10 +217,16 @@ private Content merge(
 
         if(merge) {
             existing.addCustomFields(update.getCustomFields());
+            existing.addDistributions(update.getDistributions());
```
tristanbeardmore-gray, line 220: I'd suggest not doing any kind of merging at all for production/distribution and always replace with whatever `update` has. Which fields have this merging or not is a bit arbitrary and based on the comment on line 211 it looks like we've been moving away from doing merge-on-write.

PR #1601, opened 2022-11-30
Improve merge on write

EmiEnache:


src/main/java/org/atlasapi/query/content/merge/VersionMerger.java
```diff
@@ -66,10 +81,56 @@ public void mergeVersions(
...
         }
     }
+
+    // TODO handle cases where there are multiple locations per encoding
+    // existing encodings are discarded if a new one replaces it
+    private Set<Encoding> mergeEncodings(Set<Encoding> existing, Set<Encoding> update) {
+        Set<Encoding> finalEncodings = new HashSet<>(existing);
+        for (Encoding newEncoding : update) {
+            for (Encoding existingEncoding : existing) {
+                if (sameLocation(getFirstLocation(existingEncoding), getFirstLocation(newEncoding))) {
```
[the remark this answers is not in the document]
EmiEnache: Do you agree that this addresses the issue?
https://github.com/atlasapi/atlas/pull/1601/commits/41b2251466510db851afaeac35f3ba30c8b2d588

src/main/java/org/atlasapi/query/content/merge/FieldMergeStrategy.java
```diff
@@ -44,16 +58,19 @@ public static FieldMergeStrategy fromString(String fieldsToMergeString) {
...
+       return fieldsToMerge.contains(field);
     }
 
-    public void addFieldToMerge(MergeableFields field){
-        fieldsToMerge.add(field);
+    public boolean isDefault() {
+        return this.fieldsToMerge.equals(defaultMergeStrategy().getFieldsToMerge());
     }
 
+    public Set<MergeableFields> getFieldsToMerge() {
```
tristanbeardmore-gray, line 73: Can (and ideally should) be private in my opinion, doesn't appear to be used externally.

src/main/java/org/atlasapi/query/content/merge/FieldMergeStrategy.java
```diff
@@ -44,16 +58,19 @@ public static FieldMergeStrategy fromString(String fieldsToMergeString) {
 
     }
 
-    public boolean shouldMerge(MergeableFields field){
-        return mergeEverything || fieldsToMerge.contains(field);
+    public void addFieldToMerge(MergeableFields field){
```
tristanbeardmore-gray, line 61: Realise this is pre-existing code but can we make this private whilst we're here? It's not used externally and would be a very quick and easy change to make this class effectively immutable.

src/main/java/org/atlasapi/query/content/merge/FieldMergeStrategy.java
```diff
@@ -4,31 +4,45 @@
...
+        Set<MergeableFields> fieldsToMerge = new HashSet<>();
+        for (MergeableFields field : MergeableFields.values()) {
+            if (!DO_NOT_MERGE_BY_DEFAULT.contains(field)) {
+                fieldsToMerge.add(field);
+            }
+        }
+        return new FieldMergeStrategy(fieldsToMerge);
     }
 
     public static FieldMergeStrategy create() {
```
tristanbeardmore-gray, line 40:
```suggestion
    public static FieldMergeStrategy none() {
```
If combined with the comment below about making `addFieldToMerge` we can then give this method a more meaningful name, which will improve the readability of the code for there places where we're intentionally doing no merging.

src/main/java/org/atlasapi/query/content/merge/FieldMergeStrategy.java
```diff
@@ -4,31 +4,45 @@
...
+
 import org.atlasapi.media.entity.simple.MergeableFields;
 
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
 
 public class FieldMergeStrategy {
     private static final Splitter COMMA_SPLITTER = Splitter.on(',');
+    private static final Set<MergeableFields> DO_NOT_MERGE_BY_DEFAULT = ImmutableSet.of(
```
tristanbeardmore-gray, line 17: We should have a set of default fields rather than a set of non-default fields, the reason being that as we add new mergeable fields we'll constantly hit the same issue of not wanting to break existing behaviour. With a set of non-default fields this is a potential point of failure if someone forgets to update this set.

src/main/java/org/atlasapi/query/content/merge/VersionMerger.java
```diff
@@ -72,4 +88,55 @@ public void mergeVersions(
...
+                        }
+                    }
+                }
+            }
+        }
+        finalEncodings.addAll(update);
+        return finalEncodings;
+    }
+
+    // this could maybe just be the .equals method of Location, but we don't have time to implement
```
tristanbeardmore-gray, line 109: Comment no longer applies

tristanbeardmore-gray (changes requested):

src/main/java/org/atlasapi/query/content/merge/VersionMerger.java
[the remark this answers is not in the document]
tristanbeardmore-gray: I think so, especially for a first implementation.

src/main/java/org/atlasapi/query/content/merge/VersionMerger.java
```diff
@@ -81,56 +82,59 @@ public void mergeVersions(
 
         existing.setSegmentEvents(update.getSegmentEvents());
         existing.setRestriction(update.getRestriction());
-        if(mergeStrategy.shouldMergeAll() && update.getProvider() != null) {
+        if(update.getProvider() != null) {
             existing.setProvider(update.getProvider());
+        } else {
+            existing.setProvider(null);
```
tristanbeardmore-gray, line 88: This is technically different logic to before I believe. The old logic wouldn't change `existing`'s provider if `update`'s is null, whereas here we now change `existing` to be null.

EmiEnache: I think this is the same as before. Old logic would actually change existing's provider if update's is null. I've reverted this to be exactly what it was before my attempted change.

tristanbeardmore-gray: Yeah ignore me, the diff I was looking at was based off a different commit and not master I think

tristanbeardmore-gray (approved):

PR #1603, opened 2023-02-03
Store c4 images in the  field

EmiEnache:


src/main/java/org/atlasapi/remotesite/channel4/pmlsd/C4BrandExtractor.java
```diff
@@ -88,13 +88,17 @@ private SetMultimap<ModelWithPayload<Series>, ModelWithPayload<Episode>> setBran
...
             Episode episode = episodeModelWithPayload.getModel();
+            // TODO remove unused code
 //            if (equivalentTitles(brand, episode)) {
 //                setHierarchicalTitle(episode);
 //            }
             if (episode.getImage() == null) {
                 episode.setImage(brand.getImage());
                 episode.setThumbnail(brand.getThumbnail());
             }
+            if (episode.getImages() == null) {
```
tristanbeardmore-gray, line 99: Might need to check if it's empty here as well

PR #1604, opened 2023-02-16
Fix year filter tolerance; set it to 1 for amazon->pa (for rt)

EmiEnache:
Also fixes config of other sources' equiv which seemed to be set to
yearFilterTolerance of 1, but in reality were not scoring them with any
tolerance. We are leaving them as-is with a tolerance of 0 for now.


src/main/java/org/atlasapi/equiv/update/updaters/configuration/UpdaterConfigurationRegistry.java
```diff
@@ -393,16 +393,16 @@ private static UpdaterConfiguration makeV3AmazonConfiguration() {
                         ImmutableMap.<ItemEquivalenceUpdaterType, ImmutableSet<Publisher>>builder()
                                 .put(IMDB_TOP_LEVEL_ITEM, ImmutableSet.of(IMDB))
                                 .put(STANDARD_ITEM_SEQUENCE, ImmutableSet.of(IMDB, PA, PA_API))
```
tristanbeardmore-gray, line 395: Can remove old PA from here as well whilst you're here - will help keep transitive equiv set sizes down

src/main/java/org/atlasapi/equiv/update/updaters/types/ItemEquivalenceUpdaterType.java
```diff
@@ -147,16 +147,16 @@ public enum ItemEquivalenceUpdaterType {
             )
     ),
     STANDARD_ITEM_SEARCH(
-            StandardItemSearchUpdaterProvider.create(1, 1)
+            StandardItemSearchUpdaterProvider.create(1, 0)
     ),
     STANDARD_MULTI_SAME_SOURCE_CANDIDATE_ITEM_SEARCH( //to sources that can contain duplicates
-            StandardItemSearchUpdaterProvider.create(null, 1)
+            StandardItemSearchUpdaterProvider.create(null, 0)
```
tristanbeardmore-gray, line 153: Technically we're changing logic here - we find fewer candidates from the `SherlockTitleSearchGenerator` which might've still been able to score high enough to equiv, even without the tolerance in `NonEpisodeItemYearScorer`; is this desired?

EmiEnache: Good question...
Before, the yearFilterConfig = 1 would let us find candidates (in cases where the subject had a year) that had a year off by one compared to the subject. However, these extra candidates would be scored -1 by the NonEpisodeItemYearScorer. So they would not have been able to score high enough to equiv, even if they had exact title match + unique title (sole subject and candidate title scorer). So there would be no way for them to ever equiv, because they would not be able to score 3 or higher in any circumstances.

After, the yearFilterConfig = 0 means we don't generate these candidates with year off by one. 

So no, I don't think this makes a difference in how much we equiv. (Other than as a side effect of the fact that we only take top 20 or 30 matches from the sherlock title search – with this change, that subset will be different. But unlikely to make a real difference.)

tristanbeardmore-gray: Ah, yes agreed that these candidates wouldn't actually score high enough to equiv. Also not worried about the Sherlock results capping leading to potentially more relevant candidates.

